PR #78, opened 2019-02-14
KMLTrackWriter (Mapping Project)

mamoonrashid:
Contains KMLTrackWriter class and ancillary objects that provide output of Tracker objects in a KML file.


sdhiscocks:
Looking at the KMKL generated from the test, this doesn't seem to be valid. I think it's due to the coordinates not being quite right (contains `[` and `]`).

sdhiscocks:
Closing for now, as other ways to produce mapping (see current demos). If someone is interested in KML, please comment and we'll look to reopen and fix to work on latest code.

PR #100, opened 2019-04-20
Add Ornstein-Uhlenbeck transition model

sglvladi:
Added an implementation of the Ornstein-Uhlenbeck transition model.

Similar to `ConstantVelocity`, `OrnsteinUhlenbeck` assumes that a target moves with constant velocity, but also includes a damping term that causes the velocity to exponentially decay to zero if there is no input for a prolonged period of time.


sglvladi:
@sdhiscocks Done 👍

PR #103, opened 2019-04-28
correct meas_volume

Huang-Chuan:
The volume calculated in clutter_gen() seems to be wrong.

### DIFF
--- a/stonesoup/simulator/simple.py
+++ b/stonesoup/simulator/simple.py
@@ -136,7 +136,7 @@ class SimpleDetectionSimulator(DetectionSimulator):
         clutter detections per unit volume per timestep"""
         meas_volume = float(1)
         for dim in range(self.meas_range.ndim):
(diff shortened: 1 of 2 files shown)


sdhiscocks:
Thanks @Huang-Chuan.

I think there are other issues with this method in how this is calculated meaning it only works for 2 dimensions also (use of `self.meas_range.ndim`).

I think the following could be simpler fix:
```python
    @property
    def clutter_spatial_density(self):
        """returns the clutter spatial density of the measurement space - num
        clutter detections per unit volume per timestep"""
        return self.clutter_rate/np.prod(np.diff(self.meas_range))
```

Also need to update the tests' value [here](https://github.com/Huang-Chuan/Stone-Soup/blob/480ac9a5310f78b5a49a3b6741b178cf62e6d3b2/stonesoup/simulator/tests/test_detections.py#L45).

PR #105, opened 2019-05-07
Matlab wrapper

sglvladi:
Added a `MatlabWrapper` class that can be used to create/connect to MATLAB sessions, run MATLAB code and convert the input/output.

An example of how the class can be utilised to create a `MatlabKalmanPredictor`, that computes the Kalman prediction equations in MATLAB, has also been included.


sdhiscocks:
Could you look at using this pytest import skip feature for the the Matlab test script.

sglvladi:
@sdhiscocks Thanks for pointing me to the specific resource! I think it should be fixed now. Let me know if it looks ok.

sdhiscocks:
We'll look to make this into a separate plugin.

PR #108, opened 2019-05-17
Detection Feeders

sglvladi:
This pull request includes 2 new `DetectionFeeder` objects, which may be of interest:

1. The `MetadataValueFilter` class aims to provide a simple, yet relatively flexible scheme for filtering/removing unwanted detections based on the value of some metadata field (e.g. MMSI, SNR, size, etc.). Being a subclass of `MetadataReducer`, this feeder inherits the `metadata_field` property, which it uses to store the field name of the meta value of interest. For any given detection, assuming that the meta field exists, the feeder passes the extracted meta value `x` through a user-defined unary operator/function of the form `b=f(x)`, whose handle is stored in the `operator` class property, and must be configured to return `True` if the detection should be allowed through the filter (i.e. yielded) and `False` if the detection should be dropped/filtered out. An optional `keep_unmatched` class property is defined that dictates whether detections lacking the meta field should be dropped (default) or yielded.

2. The `BoundingBoxDetectionReducer`  allows users to reduce/limit the amount of detections to a subset that falls within the limits of a given n-dimensional bounding box. The bounding box is defined as a sequence of min/max coordinate pairs, one for each box dimension, based on which the feeder then applies a filter to the incoming detections, allowing to pass through only those that fall within the specified limits, and dropping the ones that do not.   

More information about the classes can be found in the code.

### DIFF
--- a/stonesoup/feeder/filter.py
+++ b/stonesoup/feeder/filter.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from operator import attrgetter
+import numpy as np
 
 from ..base import Property
 from .base import Feeder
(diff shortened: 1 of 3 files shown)


stonesoup/feeder/filter.py
```diff
@@ -41,3 +42,147 @@ def detections_gen(self):
...
+        object,
+        doc="A unary operator/function of the form :code:`b = f(val)`, "
+            "where :code:`val` is the value of the selected `metadata_field`. "
+            "The function MUST return a :py:class:`~bool` type that evaluates "
+            "to `True` when a particul ar object satisfies the condition(s) "
+            "set by the operator, and thus should be allowed through the "
+            "filter. Any detections causing the operator to return `False` "
+            "will be filtered out. "
+            "`See here <https://docs.python.org/3/library/operator.html#module-operator>`_ "  # noqa(E504)
+            "for a list of built-in Python operators that can be readily used."
```
sdhiscocks: I think it's worth providing an example in the main documentation section, like how to use _greater than_ (like you've used in your test).

stonesoup/feeder/tests/test_filter.py
```diff
@@ -28,3 +34,83 @@ def test_metadata_reducer(detector):
         assert all(time == detection.timestamp for detection in detections)
 
     assert multi_none
+
+
+def test_metadata_value_filter(detector):
+    feeder = MetadataValueFilter(detector,
+                                 metadata_field="score",
+                                 operator=lambda x: operator.ge(x, 0.1))
```
sdhiscocks: Does the following work?
```python
lambda x: x < 0.1
```

May be easier than trying to explain use of `operator` module in your documentation string.

stonesoup/feeder/filter.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from operator import attrgetter
+import numpy as np
```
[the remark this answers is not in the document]
sglvladi: Fixed 👍

stonesoup/feeder/filter.py
```diff
@@ -41,3 +42,147 @@ def detections_gen(self):
...
+    def detections_gen(self):
+        for time, detections in self.detector.detections_gen():
+            filtered_detections = set()
+            for detection in detections:
+                value = detection.metadata.get(self.metadata_field)
+                if value is None:
+                    if self.keep_unmatched:
+                        filtered_detections.add(detection)
+                    else:
+                        continue
```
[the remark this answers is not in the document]
sglvladi: Good point! I have also made some changes to combine of the _if_ statements.

stonesoup/feeder/filter.py
```diff
@@ -41,3 +42,147 @@ def detections_gen(self):
...
+        return self._detections.copy()
+
+    def detections_gen(self):
+        num_dims = len(self.limits)
+        for time, detections in self.detector.detections_gen():
+            outlier_detections = set()
+            for detection in detections:
+                state_vector = detection.state_vector
+                for i in range(num_dims):
+                    print(self.limits)
```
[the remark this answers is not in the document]
sglvladi: Well spotted! Thanks

stonesoup/feeder/filter.py
sglvladi: Done (along with other changes relating to `operator` module)

stonesoup/feeder/tests/test_filter.py
sglvladi: I have completely removed any mention and use of the `operator` module, relating to the newly introduced class. I agree that basing the explanation around it can be confusing to new users.

The `operator` module was used more heavily in the initial draft of the class, which I ended up modifying to add more flexibility.

stonesoup/feeder/filter.py
```diff
@@ -16,7 +19,7 @@ class MetadataReducer(Feeder):
 
     metadata_field = Property(
         str,
-        doc="Field used to reduce unique set of detections")
+        doc="Field used to reduce set of detections")
```
sglvladi: Removed the word `unique` to make the docstring more generic.

stonesoup/feeder/tests/test_filter.py
```diff
@@ -28,3 +33,83 @@ def test_metadata_reducer(detector):
...
+        assert all(time == detection.timestamp for detection in detections)
+    assert not nones
+
+    # Keep unmatched
+    feeder.keep_unmatched = True
+    nones = False
+    for time, detections in feeder.detections_gen():
+        all_scores = [detection.metadata.get('score')
+                      for detection in detections]
+        print(all_scores)
```
sdhiscocks: This should be removed.

sdhiscocks:
> Looks fine as far as I can make out. Of course what I'd really like, and can't work out how to do is to be able to muck about with the tests myself. Any way to do that?

https://help.github.com/en/articles/checking-out-pull-requests-locally

PR #110, opened 2019-05-29
Simpler Kalman squash

jmbarr:
Supersedes the now outdated Simpler Kalman pull request.


jmbarr:
> Looks good, minus the failed CircleCI tests.

These now pass...

sdhiscocks:
@jmbarr: Could you please update the documentation strings so they are in NumPy format.

stonesoup/models/base.py
```diff
@@ -54,6 +54,8 @@ def function(self, state_vector, noise=None, **kwargs):
         """
 
         if noise is None:
+            # noise = sp.zeros(state_vector.shape)
+            # TODO resolve the noise=None generating noise issue
             noise = self.rvs(**kwargs)
```
sdhiscocks: This now contradicts the documentation. Suggest this is addressed in separate pull request.

stonesoup/models/measurement/linear.py
```diff
@@ -73,7 +74,8 @@ def function(self, state_vector, noise=None, **kwargs):
         """
 
         if noise is None:
-            noise = self.rvs()
+            # noise = sp.zeros(state_vector.shape)
+            noise = self.rvs()  # TODO noise=None generates noise!
```
sdhiscocks: As per other comment, this can be done in separate pull request.

stonesoup/models/transition/nonlinear.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
```
sdhiscocks: Please remove until ready, and raise separate pull request.

stonesoup/predictor/tests/test_kalman.py
```diff
@@ -3,11 +3,11 @@
 import pytest
 import numpy as np
 
-from ...models.transition.linear import ConstantVelocity
-from ...predictor.kalman import (
+from stonesoup.models.transition.linear import ConstantVelocity
```
sdhiscocks: This change shouldn't be required.

stonesoup/updater/kalman.py
```diff
@@ -3,831 +3,299 @@
...
+        # Complete the calculation of the posterior
+        kalman_gain = m_cross_cov @ np.linalg.inv(innov_cov)  # This isn't
+        # optimised
+        posterior_mean = predicted_state.state_vector + \
+            kalman_gain @ (hypothesis.measurement.state_vector - pred_meas)
+        posterior_covariance = \
+            predicted_state.covar - kalman_gain @ innov_cov @ kalman_gain.T
+
+        # TODO: check and remove this line
+        # posterior_state_covariance = (P_post + P_post.T) / 2 # !!! kludge
```
sdhiscocks: Can this be removed?

stonesoup/models/base.py
jmbarr: Done.

stonesoup/models/measurement/linear.py
jmbarr: Done.

stonesoup/models/transition/nonlinear.py
jmbarr: ok. Whole file removed.

stonesoup/predictor/tests/test_kalman.py
jmbarr: indeed it isn't. Wonder why I felt the need. Reverted.

stonesoup/updater/kalman.py
jmbarr: I wanted a second opinion (still do). 

It's a practical kludge to cover over those instances where numerical instabilities leave you with a non-positive semidefinite covariance matrix. Which isn't allowed and causes onward problems. Our options are:

1. Leave it in - if the matrix is symmetric (as it should be) it has zero effect
2. Take it out - either on principle, or because it acts to cover up instances where the matrix is misbehaving. And in general we'd want to know about that.
3. Compromise - put in a flag which will implement this symmetrisation if required (or by default)

jmbarr: Went for 3. Update now has a keyword flag, <force_symmetric_covariance=False>. Set to True if you want to force the symmetry of the covariance matrix.

jmbarr:
> 
> 
> @jmbarr: Could you please update the documentation strings so they are in NumPy format.

Done; now compliant, as far as I can see. Just chasing the last Flake8 errors...

PR #112, opened 2019-06-21
Nth Derivative model

snaylor20:
This model keeps the Nth derivative with respect to time constant.
Also packaged `ConstantVelocity` and `ConstantAcceleration` as variants with N=1 and N=2.

### DIFF
--- a/stonesoup/base.py
+++ b/stonesoup/base.py
@@ -159,8 +159,12 @@ class BaseMeta(ABCMeta):
         cls._properties.update(
             (key, value) for key, value in namespace.items()
             if isinstance(value, Property))
-        # Optional arguments must follow mandatory
         for name in list(cls._properties):
+            # Remove items which are no longer properties
+            if name in namespace and not isinstance(namespace[name], Property):
+                del cls._properties[name]
+                continue
+            # Optional arguments must follow mandatory
             if cls._properties[name].default is not Property.empty:
                 cls._properties.move_to_end(name)
 
--- a/stonesoup/models/transition/linear.py
+++ b/stonesoup/models/transition/linear.py
@@ -309,7 +309,7 @@ class OrnsteinUhlenbeck(LinearGaussianTransitionModel, TimeVariantModel):
 
 
 class ConstantNthDerivative(LinearGaussianTransitionModel, TimeVariantModel):
-    r"""Model based on the Nth derivative of position being constant,
+    r"""Model based on the Nth derivative with respect to time being constant,
     to set derivative use keyword argument :attr:`constant_derivative`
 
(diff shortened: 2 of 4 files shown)


stonesoup/models/transition/linear.py
```diff
@@ -307,12 +308,92 @@ def covar(self, time_interval, **kwargs):
         return CovarianceMatrix(covar)
 
 
-class ConstantVelocity(LinearGaussianTransitionModel, TimeVariantModel):
+class ConstantNthDerivative(LinearGaussianTransitionModel, TimeVariantModel):
+    r"""Model based on the Nth derivative of position being constant,
```
sglvladi: An alternative to the word "position" may be required. I assume that the same model can also be applied to variables like bearing, orientation, etc.

stonesoup/models/transition/linear.py
```diff
@@ -307,12 +308,92 @@ def covar(self, time_interval, **kwargs):
...
+
+    It is hard to represent the matrix form of these due to the fact that they
+    vary with N, examples for N=1 and N=2 can be found in the
+    :class:`~.ConstantVelocity` and :class:`~.ConstantAcceleration` models
+    respectively. To aid visualisation of :math:`F_t` the elements are
+    calculated as the terms of the taylor expansion of each state variable.
+    """
+
+    constant_derivative = Property(
+        int, doc="The position derivative to be kept constant, "
```
sglvladi: See above comment about "position"

stonesoup/base.py
```diff
@@ -159,8 +159,12 @@ def __init__(self, *args, **kwargs):
         cls._properties.update(
             (key, value) for key, value in namespace.items()
             if isinstance(value, Property))
-        # Optional arguments must follow mandatory
         for name in list(cls._properties):
+            # Remove items which are no longer properties
```
sglvladi: Looks like these changes may be messing up the documentation. Under the documentation for `constant_derivative` property for the `RandomWalk` and `ConstantVelocity` classes I am getting the documentation for `Property`.

stonesoup/models/transition/linear.py
sglvladi: May be worth using different terminology to  "position". I suspect that this can also be applied to variables like bearing, orientation, etc.

stonesoup/models/transition/linear.py
sglvladi: See above comment above using word "position".

stonesoup/base.py
snaylor20: Cheers, I don't seem to get this issue, could you possibly send a screenshot.

PR #114, opened 2019-06-25
DistanceHypothesiser gate out hypotheses greater than missed distance

sdhiscocks:


stonesoup/hypothesiser/tests/test_distance.py
```diff
@@ -35,3 +41,26 @@ def test_mahalanobis(predictor, updater):
...
+    hypothesiser = DistanceHypothesiser(
+        predictor, updater, measure=measure, missed_distance=1,
+        include_all=True)
+
+    hypotheses = hypothesiser.hypothesise(track, detections, timestamp)
+
+    # There are 4 hypotheses - Detections and Missed Detection
+    assert len(hypotheses) == 4
+
+    # And not detection3
```
dlast-dstl: Looks like this comment got copied directly from another test and does not accurately reflect what is happening here.

PR #120, opened 2019-07-10
Create OpenSky Network REST API Reader

sdhiscocks:


svidal-dstl (approved):

sdhiscocks:
I've just rebased this onto `main` and added some tests which can be run locally (skipped by default due to requirement for remote data).

oharrald-Dstl (approved):

Issue #118, opened 2019-06-28
Reading Track metadata becomes costly for Tracks with long state histories


sglvladi:
Following experimentation with some large scale AIS datasets and it has been observed that continuously reading the Track metadata becomes a costly operation, especially as the state history of a Track grows.  This due to the fact that the getter method has to iterate over all the states to yield the result as it can be seen [here](https://github.com/dstl/Stone-Soup/blob/ac8299435f8f53b8e27e1939fc95c9ed1efb0178/stonesoup/types/track.py#L47). 

A potential fix can be applied by replacing the above lines as follows:

```python
@property
def metadata(self):
...
    for state in reversed(self.states):
        if isinstance(state, Update)\
           and state.hypothesis.measurement.metadata is not None:
            metadata.update(state.hypothesis.measurement.metadata)
            break
...
```
Here is an example profiler output running the same code with the old and new versions:

Old
---
```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
4789021  184.180    0.000 1897.829    0.000 ...\stonesoup\stonesoup\types\track.py:31(metadata)
```

New
---
```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
4789021  103.760    0.000  399.821    0.000 ...\stonesoup\stonesoup\types\track.py:31(metadata)
```

However, by doing so, only the metadata fields of the latest detection are returned, which may not be desireable (e.g. in cases where the metadata fields vary between detections). 

Seeing as metadata is becoming quite heavily used in StoneSoup, it may be worth having a different approach to storing/updating the Track metadata.

sglvladi:
This has been fixed in #121

PR #121, opened 2019-07-10
Track metadata handling update

sglvladi:
Following discussions with @sdhiscocks during Fusion in relation to #118, this pull request aims to solve the performance issues relating to reading metadata as follows:

- Track objects will now maintain a copy of what should be the most up-to-date metadata at each time. Thus, reading metadata should be more efficient, at the expense of the additional memory required to store the metadata and a small computational overhead when new states are added to a track.
- Metadata is updated automatically every time a new state is added to a track. This assumes (as we have done so far) that the last state in the list of states is also the latest one, i.e. the proposed approach will "fail" if states are added in an out-of-sequence manner.
- Although not implemented here, this approach can be extended to allow users to edit and/or add to the track metadata (e.g. by implementing a `Track.metadata()` setter method).

As part of the above, a small correction has also been applied to `StateMutableSequence` to ensure that the `states` property is always a list after initialisation, even if a single `State` object is passed. Previously, a statement of the form ```track = Track(states=State(np.array([[1]]), timestamp=datetime.datetime.now())``` would cause `track.states` to be a `State` object, rather than a list of length 1.

### DIFF
--- a/stonesoup/types/state.py
+++ b/stonesoup/types/state.py
@@ -61,6 +61,9 @@ class StateMutableSequence(Type, MutableSequence):
     def __init__(self, states=None, *args, **kwargs):
         if states is None:
             states = []
+        elif not isinstance(states, list):
+            # Ensure states is a list
+            states = [states]
         super().__init__(states, *args, **kwargs)
 
     def __len__(self):
--- a/stonesoup/types/track.py
+++ b/stonesoup/types/track.py
@@ -2,6 +2,7 @@
 import uuid
 
 from ..base import Property
+from .multihypothesis import MultipleHypothesis
 from .state import State, StateMutableSequence
 from .update import Update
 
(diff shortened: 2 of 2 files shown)


stonesoup/types/track.py
```diff
@@ -42,11 +57,33 @@ def metadata(self):
...
 
-        return metadata
+        if isinstance(state, Update):
+            if isinstance(state.hypothesis, MultipleHypothesis):
+                # Sort and iterate through multiple hypotheses such that most
+                # likely hypothesis comes last. This ensures that metadata
+                # from all hypotheses are retained, but more likely
+                # hypotheses will over-write the metadata set by less likely
+                # ones.
+                for hypothesis in sorted(state.hypothesis, reverse=True):
```
sglvladi: Not sure if this is correct. My (possibly wrong) intuition says that calling `sorted` on a MultipleHypothesis object should return a list of hypotheses, with the most likely being first. In the case of `SingleProbabilityHypothesis` this means that the hypothesis with highest probability would come first, while for `SingleDistanceHypothesis` the hypothesis with lowest distance would be first.. Thus, to have them sorted in reverse, I have included the argument `reverse=True`.

stonesoup/types/state.py
```diff
@@ -61,6 +61,9 @@ class StateMutableSequence(Type, MutableSequence):
     def __init__(self, states=None, *args, **kwargs):
         if states is None:
             states = []
+        elif not isinstance(states, list):
```
sdhiscocks: May be better to check if `states` is a `collections.abc.MutableSequence` instead.

PR #125, opened 2019-07-26
Add inverse bearing function &reverse method names

lflaherty-dstl:
Changed the method names as per discussion with Steve about their logical meaning.
Also added an inverse function to convert range/bearing co-ordinates to Cartesian, using Jacobian matrices.

### DIFF
--- a/stonesoup/models/base.py
+++ b/stonesoup/models/base.py
@@ -114,7 +114,7 @@ class ReversibleModel(NonLinearModel):
 
     @abstractmethod
     def inverse_function(self, state_vector, **kwargs):
-        """Takes in the result of the functionr and
+        """Takes in the result of the function and
         computes the inverse function, returning the initial
         input of the function.
 
--- a/stonesoup/models/base.py
+++ b/stonesoup/models/base.py
@@ -56,7 +56,7 @@ class LinearModel(Model):
         if noise is None:
             noise = self.rvs(**kwargs)
 
-        return self.matrix(**kwargs)@state_vector + noise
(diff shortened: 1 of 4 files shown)


stonesoup/models/base.py
```diff
@@ -104,6 +104,32 @@ def function(self, state_vector, noise=None, **kwargs):
...
 
+class ReversibleModel(NonLinearModel):
+    """Non-linear model containing sufficient co-ordinate
+    information such that the linear co-ordinate conversions
+    can be calculated from the non-linear counterparts.
+
+    Contains an inverse function which computes the reverse
+    of the relevant linear-to-non-linear function"""
+
+    def inversefunction(self, state_vector, **kwargs):
```
sdhiscocks: This should have an `@abstractmethod` decorator.

Also, for consistency with other methods, this should be `inverse_function`.

stonesoup/models/base.py
```diff
@@ -104,6 +104,33 @@ def function(self, state_vector, noise=None, **kwargs):
...
+    """Non-linear model containing sufficient co-ordinate
+    information such that the linear co-ordinate conversions
+    can be calculated from the non-linear counterparts.
+
+    Contains an inverse function which computes the reverse
+    of the relevant linear-to-non-linear function"""
+
+    @abstractmethod
+    def inverse_function(self, state_vector, **kwargs):
+        """Takes in the result of the functionr and
```
DaveKirkland:
```suggestion
        """Takes in the result of the function and
```

PR #126, opened 2019-07-26
Add 2d assignment alg + new gnn

lflaherty-dstl:
I added the 2d assignment algorithm from David Crouse into a section of stone soup for the purpose of creating an improved Global Nearest Neighbour algorithm - with success, it is faster!
I am aware it will probably need to be moved into a better place.


stonesoup/dataassociator/neighbour.py
```diff
@@ -106,3 +111,92 @@ def associate(self, tracks, detections, time):
...
+                tracks[i], detections, time)
+            for hypothesis in track_hypotheses:
+                if isinstance(hypothesis.measurement, MissedDetection):
+                    row[len(detections) + i] = hypothesis
+                else:
+                    row[int(detections.index(
+                        hypothesis.measurement))] = hypothesis
+            hypothesis_matrix[i] = row
+
+        probability_flag = isinstance(self.hypothesiser, PDAHypothesiser)
```
sdhiscocks: This should check the types of _hypothesis_ as this will work more generically with other _hypothesisers_. Maybe something like (not tested):
```suggestion
        if all(isinstance(hypothesis, ProbabilityHypothesis) for hypothesis in hypothesis_matrix):
            probability_flag = True
        elif all(isinstance(hypothesis, DistanceHypothesis) for hypothesis in hypothesis_matrix):
            probability_flag = False
        else:
            raise RuntimeError("2d assignment doesn't support mixed types of Hypothesis")
```

stonesoup/dataassociator/neighbour.py
```diff
@@ -106,3 +111,92 @@ def associate(self, tracks, detections, time):
...
+        """
+
+        # Convert sets to indexable lists
+        tracks = list(tracks)
+        detections = list(detections)
+
+        # Generate 2d array "matrix" of hypotheses mapping track to detection
+        hypothesis_matrix = np.empty(
+            [len(tracks), len(detections) + len(tracks)],
+            SingleDistanceHypothesis)
```
sdhiscocks: Whilst this doesn't have an impact as _object_ array is created, this implies _distance_. Could just have `SingleHypothesis` or `object` to keep generic.

stonesoup/dataassociator/neighbour.py
```diff
@@ -106,3 +111,92 @@ def associate(self, tracks, detections, time):
...
+        for i in range(len(tracks)):
+            row = np.empty(
+                [len(detections) + len(tracks)], SingleDistanceHypothesis)
```
sdhiscocks: See comment above ref. using `SingleDistanceHypothesis`

stonesoup/assignment/assignAlgs.py
```diff
@@ -0,0 +1,313 @@
+import numpy
+
+
+def assign2D(C, maximize=False):
```
sdhiscocks: I think this should be move to `stonesoup.dataassociator._assignment` as not intend to be public component.

Also would be good to get in are usual documentation format, but not essential.

stonesoup/dataassociator/_assignment.py
```diff
@@ -0,0 +1,319 @@
+import numpy
+
+
+class AssignAlgs:
```
sdhiscocks: No need for this class; just having the function will be fine.

stonesoup/dataassociator/neighbour.py
```diff
@@ -106,3 +110,108 @@ def associate(self, tracks, detections, time):
...
+        detections = list(detections)
+
+        # Generate 2d array "matrix" of hypotheses mapping track to detection
+        hypothesis_matrix = np.empty(
+            [len(tracks), len(detections) + len(tracks)],
+            SingleHypothesis)
+        for i in range(len(tracks)):
+            row = np.empty(
+                [len(detections) + len(tracks)], SingleHypothesis)
+            track_hypotheses = self.hypothesiser.hypothesise(
```
sglvladi: It might be worth using the standard way of generating the hypotheses prior to the for loop, i.e.
```python
# Generate a set of hypotheses for each track on each detection
hypotheses = {
    track: self.hypothesiser.hypothesise(track, detections, time)
    for track in tracks}
```

and then this line can be replaced by:
```python
track_hypotheses = hypotheses[track]
```

This will make the hypothesis generation clearer (as it follows the same format as other data associators), while also provisioning for easier integration of changes introduced by #145.

stonesoup/dataassociator/neighbour.py
```diff
@@ -106,3 +110,108 @@ def associate(self, tracks, detections, time):
...
+        """
+
+        associations = {}
+
+        # Convert sets to indexable lists
+        tracks = list(tracks)
+        detections = list(detections)
+
+        # Generate 2d array "matrix" of hypotheses mapping track to detection
+        hypothesis_matrix = np.empty(
```
sglvladi: Further speed improvements can be achieved by only performing data association for tracks that are detected (i.e. not missed), as the 2D assignment complexity scales according the number of targets (evident from the shape requirement of `hypothesis_matrix` and `distance_matrix`).

This could be achieved by replacing `tracks = list(tracks)` with:
```python
# Extract detected tracks
# NOTE: It is assumed that if a track is undetected/missed, then it will only have a single hypothesis, which
#       is also a miss-detection hypothesis (i.e. evaluates to False)
detected_tracks = [track for track in tracks if not(len(hypotheses[track]) == 1 and not hypotheses[track][0])]

# Store associations for undetected/missed tracks
for track in tracks-set(detected_tracks):
    associations[track] = hypotheses[track][0]

# No need to perform data association if all tracks are missed
if not len(detected_tracks):
    return associations
```
and then replacing all occurrences of `tracks` with `detected tracks` further down.

The above assumes that hypothesis generation has been performed as suggested in my other comment.

stonesoup/dataassociator/_assignment.py
```diff
@@ -0,0 +1,319 @@
...
+    # [3]   Murty, K. G. "An algorithm for ranking all the assignments in
+    #       order of increasing cost," Operations Research, vol. 16, no. 3,
+    #       pp. 682-687, May-Jun. 1968.
+    #
+    # May 2018 David F. Crouse, Naval Research Laboratory, Washington D.C.
+    # (UNCLASSIFIED) DISTRIBUTION STATEMENT A. Approved for public release.
+    # This work was supported by the Office of Naval Research through the
+    # Naval Research Laboratory 6.1 Base Program
+
+    C = numpy.matrix(C)  # In case a multidimensional array was passed.
```
sglvladi: This seems to throw a numpy PendingDeprecationWarning:

```
PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
    C = numpy.matrix(C)  # In case a multidimensional array was passed.
```

Might be worth fixing

stonesoup/dataassociator/neighbour.py
```diff
@@ -106,3 +110,108 @@ def associate(self, tracks, detections, time):
...
+            [len(tracks), len(detections) + len(tracks)],
+            SingleHypothesis)
+        for i in range(len(tracks)):
+            row = np.empty(
+                [len(detections) + len(tracks)], SingleHypothesis)
+            track_hypotheses = self.hypothesiser.hypothesise(
+                tracks[i], detections, time)
+            for hypothesis in track_hypotheses:
+                if isinstance(hypothesis.measurement, MissedDetection):
```
sglvladi: I think this can be re-written more compactly as: 
``` python
if not hypothesis:
```

sdhiscocks:
@sglvladi: Recent commits hopefully address your comments. Would be good if you could test to ensure still functional. Also, would be good to double check there isn't impact on performance (I'm expecting your suggested approach to be faster).

PR #128, opened 2019-08-08
add multi meas initiator & non linear w tests

lflaherty-dstl:
Altered the linear initiator to incorporate non-linear co-ordinates and changed name to simple. Added multi measurement initiator. Altered test file to test new parts.

### DIFF
--- a/stonesoup/initiator/simple.py
+++ b/stonesoup/initiator/simple.py
@@ -60,8 +60,8 @@ class SimpleMeasurementInitiator(GaussianInitiator):
     This initiator utilises the :class:`~.MeasurementModel` matrix to convert
     :class:`~.Detection` state vector and model covariance into state space.
 
-    Utilises the CartesianToBearingRange inverse function to convert
-    non-linear bearing/range co-ordinates into Cartesian x/y co-ordinates
+    Utilises the ReversibleModel inverse function to convert
+    non-linear spherical co-ordinates into Cartesian x/y co-ordinates
     for use in predictions and mapping.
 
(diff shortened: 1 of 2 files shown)


lflaherty-dstl:
NOTE: this request will not pass until inverse function request is approved as there are dependencies :)

stonesoup/initiator/simple.py
```diff
@@ -48,11 +52,18 @@ def initiate(self, unassociated_detections, **kwargs):
...
-class LinearMeasurementInitiator(GaussianInitiator):
+class SimpleMeasurementInitiator(GaussianInitiator):
     """Initiator that maps measurement space to state space
 
+    Works for both linear and non-linear co-ordinate input
+
     This initiator utilises the :class:`~.MeasurementModel` matrix to convert
     :class:`~.Detection` state vector and model covariance into state space.
+
+    Utilises the CartesianToBearingRange inverse function to convert
```
sdhiscocks: This comment should be updated, as in works for more than just the _CartesianToBearingRange_ model.

lflaherty-dstl: Thanks Steve! Will do this when the branch is ready for re-testing i.e. when the inverse function is merged :)

sdhiscocks:
> Does this just support a set of M consecutive detections or does it allow for M of N?

It can indirectly driven by the deleter, which can differ from the one in the tracker itself (recently experimented with this using 3 of 5 for initialisation, but more forgiving deleter in the main tracker).

PR #133, opened 2019-08-21
Nth Derivative Decay

snaylor20:
Adds the Nth derivative decay model, and packages the OU and singer in to that.

### DIFF
--- a/stonesoup/models/transition/linear.py
+++ b/stonesoup/models/transition/linear.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 import math
+from functools import lru_cache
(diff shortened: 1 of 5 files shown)


DaveKirkland:
For the CV, CA and Nth derivative models, I think we should document whether they are discrete or continuous versions of the models. I didn't see it in there - but I may have missed it. I don't know if there is the same distinction for the decaying model types.

DaveKirkland:
For completeness I'd like to see a test of the RandomWalk class. I know it's pretty minimal. I'm not sure if this was a new addition or if it was in beforehand.

PR #134, opened 2019-08-21
Adds a variety of switching simulators.

snaylor20:
The ground truth simulators take a list of transition models, they
then switch between these based on a Markov transition matrix that the
user inputs.

The detection simulator changes the probability of detection
based on what transition model the ground truth user is using at the
given time.

### DIFF
--- a/stonesoup/simulator/simple.py
+++ b/stonesoup/simulator/simple.py
@@ -2,7 +2,6 @@
 import datetime
 
 import numpy as np
-
 from ..base import Property
 from ..models.measurement import MeasurementModel
 from ..models.transition import TransitionModel
@@ -31,6 +30,7 @@ class SingleTargetGroundTruthSimulator(GroundTruthSimulator):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._groundtruth_paths = set()
+        self.index = 0
 
     @property
     def groundtruth_paths(self):
@@ -41,7 +41,8 @@ class SingleTargetGroundTruthSimulator(GroundTruthSimulator):
         time = self.initial_state.timestamp or datetime.datetime.now()
 
         gttrack = GroundTruthPath([
-            GroundTruthState(self.initial_state.state_vector, timestamp=time)])
+            GroundTruthState(self.initial_state.state_vector, timestamp=time,
+                             metadata={"index": self.index})])
         self._groundtruth_paths.add(gttrack)
         yield time, self.groundtruth_paths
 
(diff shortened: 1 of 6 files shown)


stonesoup/simulator/simple.py
```diff
@@ -52,16 +53,43 @@ def groundtruth_paths_gen(self):
...
 
 
+class SwitchOneTargetGroundTruthSimulator(SingleTargetGroundTruthSimulator):
+    """Target simulator that produces a single target. This target switches
+    between multiple transition models based on a markov matrix
+    (:attr:`model_probs`)"""
+    transition_models = Property(
+        [TransitionModel], doc="List of transition models to be used, ensure\
+        that they all have the same dimensions.")
+    model_probs = Property([float], doc="A matrix of probabilities.\
```
sdhiscocks: The type should probably be `np.ndarry` rather than `[float]`.

stonesoup/simulator/tests/test_groundtruth.py
```diff
@@ -67,3 +68,80 @@ def test_multitarget_ground_truth_simulator(transition_model, timestep):
...
+    initial_state = State(np.array([[1]]), timestamp=datetime.datetime.now())
+    model_probs = [[0.5, 0.5], [0.5, 0.5]]
+    simulator = SwitchOneTargetGroundTruthSimulator(
+        transition_models=[transition_model1, transition_model2],
+        model_probs=model_probs,
+        initial_state=initial_state,
+        timestep=timestep)
+
+    for step, (time, groundtruth_paths) in \
+            enumerate(simulator.groundtruth_paths_gen()):
```
rjgreen-dstl: Is this line consistent with those changes in #137

stonesoup/simulator/simple.py
```diff
@@ -52,16 +53,43 @@ def groundtruth_paths_gen(self):
...
                 time_interval=self.timestep)
             gttrack.append(GroundTruthState(
-                trans_state_vector, timestamp=time))
+                trans_state_vector, timestamp=time,
+                metadata={"index": self.index}))
 
             yield time, self.groundtruth_paths
 
 
+class SwitchOneTargetGroundTruthSimulator(SingleTargetGroundTruthSimulator):
```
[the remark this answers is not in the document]
snaylor20: If memory serves, I used that name to save characters. As before it was over the character limit. There might be a work around to this I'm not aware of.

stonesoup/simulator/tests/test_groundtruth.py
sdhiscocks: Shouldn't be an issue, but I'll resolve on merging.

PR #136, opened 2019-08-21
Constant turn sandwich

snaylor20:
Adds a transition model which allows for a constant turn in 2 state variables which are not adjacent in the state vector, eg x and z in the (x, y, z) state vector

### DIFF
--- a/stonesoup/models/transition/linear.py
+++ b/stonesoup/models/transition/linear.py
@@ -683,7 +683,85 @@ class SingerApproximate(Singer):
         return CovarianceMatrix(covar)
 
 
-class ConstantTurn(LinearGaussianTransitionModel, TimeVariantModel):
+class ConstantTurnSandwich(LinearGaussianTransitionModel, TimeVariantModel):
+    r"""This is a class implementation of a time-variant 2D Constant Turn
+    Model. This model is used, as opposed to the normal :class:`~.ConstantTurn`
+    model, when the turn occurs in 2 dimensions that are not adjacent in the
+    state vector, eg if the turn occurs in the x-z plane but the state vector
+    is of the form :math:`(x,y,z)`. The list of transition models are to be
+    applied to any state variables that lie in between, eg if for the above
+    example you wanted the y component to move with constant velocity, you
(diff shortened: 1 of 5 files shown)


stonesoup/models/transition/linear.py
```diff
@@ -744,74 +822,14 @@ class ConstantTurn(LinearGaussianTransitionModel, TimeVariantModel):
                       \end{bmatrix}
     """
 
-    noise_diff_coeffs = Property(
+    turn_noise_diff_coeffs = Property(
```
sdhiscocks: These properties are inherited, so no longer required here.

stonesoup/models/transition/tests/test_ct.py
```diff
@@ -43,24 +44,24 @@ def base(model, state_vec, noise_diff_coeffs, turn_rate):
              [0, sp.sin(turn_ratedt),
               0, sp.cos(turn_ratedt)]])
 
-    qx = noise_diff_coeffs[0]
-    qy = noise_diff_coeffs[1]
-    Q = sp.array([[sp.power(qx, 2) * sp.power(timediff, 3) / 3,
```
etfrogers-dstl: I assume `sp.array` will not merge well here, as `master` is now using `np.array`

This applies throughout

stonesoup/models/transition/linear.py
```diff
@@ -501,7 +380,147 @@ def constant_derivative(self):
...
+        float, doc="The noise diffusion coefficient :math:`q`")
+    damping_coeff = Property(
+        float, doc="The Nth derivative damping coefficient :math:`K`")
+
+    @property
+    def ndim_state(self):
+        return self.decay_derivative + 1
+
+    @staticmethod
+    @lru_cache()
```
etfrogers-dstl: Like the use of `lrucache`. Have you considered whether an explicit cache size is useful? Is the default of 128 sensible given the likely number of combinations of `N` and `K`?

stonesoup/models/transition/tests/test_ct.py
sdhiscocks: I can take care of this when merging.

PR #139, opened 2019-09-11
SIAP metric output table with red and green visual

lflaherty-dstl:
Adds a new kind of metric visualisation, so far implemented for SIAP metrics only. This allows metrics to be displayed along with a short description of what the metric means as well as a target value for each metric (as these can vary). The metric value is also shown with a coloured background from red-green to visually represent the accuracy of each metric.

This will be of particular interest to the OSTEWG challenge participants, as it provides an easy way of comparing the accuracy of different trackers.

### DIFF
--- a/stonesoup/metricgenerator/base.py
+++ b/stonesoup/metricgenerator/base.py
@@ -36,14 +36,14 @@ class MetricManager(Base):
     """
 
 
-class MetricTableGenerator(Base):
+class MetricTableGenerator(MetricGenerator):
     """Metric Table base class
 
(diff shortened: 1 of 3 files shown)


stonesoup/metricgenerator/base.py
```diff
@@ -36,6 +36,27 @@ class MetricManager(Base):
     """
 
 
+class MetricTableGenerator(Base):
```
sdhiscocks: This should inherit from MetricGenerator, and then _generate_table_ could renamed _compute_metric_ method.

PR #142, opened 2019-10-03
Handling model orders in mixture models

DaveKirkland:
I've based this on Lyudmils IMM branch. This code handles the different order models. I've included some examples in the example directory. The user has a fair bit to manage to get it right. The list of models provided to the IMMupdater and IMMpredictor need to coincide. 

I've also coded to routines to do the state conversions cv2ca and ca2cv. These are provided to the Updater and Predictor - these handle the conversion of the state predictions, measurement predictions and the state updates. They need to work with a number of different types GaussianState, GaussianMeasurementPrediction, WeightedGaussianState, and a few others.

I'm not looking to merge this into the master just yet, but I'd like to get some feedback on the design.

The IMMPredictor/IMMUpdater is calling the conversion routines explicitly right now - so that part would need to be duplicated in the other mixture model code. I don't really like it this way - it is very easy to get wrong. A better way is to used a decorator on the predict() and update() functions. Trying to add a decorator around the predict() and update() abstractmethods in the base class doesn't work - they don't get called. I can put them explicitly around the methods in the KalmanPredictor and KalmanUpdater classes and that does work.


DaveKirkland:
I've updated the implementation. We now create a new class for the predictors/updaters to  be used in the IMM predictor/updater. This keeps the code inside the IMM very clean.

Some TBD:
 - If called multiple times with the same class type, this will create a new class every time.
 - I put the new class in the predictor directory for now. This should probably be put somewhere else.

examples/imm_example2.py
```diff
@@ -0,0 +1,287 @@
...
+
+from matplotlib.patches import Ellipse
+
+#######################################################################
+# This uses a CV and CA IMM. Inside the IMM uses the CA model
+# The taarget simulator uses a CV model.
+#######################################################################
+
+
+def cv2ca(state):
```
rjgreen-dstl, line 39: I think that both cv2ca and ca2cv could be generalized to 2d and 3d and added to stonesoup.functions - they hare likely to be the mainstay for a lot of approaches like IMM.

examples/imm_example2.py
```diff
@@ -0,0 +1,287 @@
...
+
+KalmanPredictorMM = add_mixture_capability(KalmanPredictor)
+predictor_1 = KalmanPredictor(transition_model_1)
+predictor_2 = KalmanPredictor(transition_model_2)
+predictor_cv = KalmanPredictorMM(trans_1, convert2common_state=cv2ca,
+                                 convert2local_state=ca2cv)
+predictor_ca = KalmanPredictorMM(trans_2, convert2common_state=null_convert,
+                                 convert2local_state=null_convert)
+imm_predictor = IMMPredictor([predictor_cv, predictor_ca],
+                             model_transition_matrix)
```
rjgreen-dstl, line 186: Would it be possible to convert the IMMPredictor to take a filter class, motion models and transition matrix as inputs (with optional input for convert2common_state and convert2local_state). Internally the IMMPredictor could then construct the set of filters from its inputs. 

At the moment their is a significant burden on the user to build up these banks of filters and it requires significant knowledge of the code base. Something like the below would be easier for user pickup

``imm_predictor = IMMPredictor(KalmanPredictor, [transition_model_1, transition_model_2], model_transition_matrix)

internally the IMMPredictor could then check the transition models to build the appropriate filter class MM extensions

examples/imm_example2.py
```diff
@@ -0,0 +1,287 @@
...
+
+KalmanUpdaterMM = add_mixture_capability(KalmanUpdater)
+updater_cv = KalmanUpdaterMM(measurement_model,
+                             convert2common_state=cv2ca,
+                             convert2local_state=ca2cv)
+updater_ca = KalmanUpdaterMM(measurement_model2)
+imm_updater = IMMUpdater([updater_cv, updater_ca], model_transition_matrix)
```
rjgreen-dstl, line 193: as above

PR #145, opened 2019-10-22
Add tree data structures for gating

sdhiscocks:
This constructs a kd-tree for detections, wrapping an existing hypothesiser to provide a form of gating. Results in notable performance improvement, and configurable.

TODO:
- [ ] Test cases
- [ ] Max distance factor of track prediction covariance?

### DIFF
--- a/stonesoup/dataassociator/tree.py
+++ b/stonesoup/dataassociator/tree.py
@@ -2,6 +2,7 @@
 import datetime
 from collections import defaultdict
 from operator import attrgetter
+from typing import Sequence
 
 import numpy as np
 import scipy as sp
@@ -35,18 +36,16 @@ class DetectionKDTreeMixIn(DataAssociator):
     This is only suitable where measurements are in same space as each other
     and at the same timestamp.
     """
-    predictor = Property(
-        Predictor,
+    predictor: Predictor = Property(
         doc="Predict tracks to detection times")
-    updater = Property(
-        Updater,
+    updater: Updater = Property(
         doc="Updater used to get measurement prediction")
-    number_of_neighbours = Property(
-        int, default=None,
+    number_of_neighbours: int = Property(
+        default=None,
         doc="Number of neighbours to find. Default `None`, which means all "
             "points within the :attr:`max_distance` are returned.")
-    max_distance = Property(
-        float, default=np.inf,
+    max_distance: float = Property(
+        default=np.inf,
         doc="Max distance to return points. Default `inf`")
 
     def generate_hypotheses(self, tracks, detections, timestamp, **kwargs):
@@ -92,14 +91,11 @@ class TPRTreeMixIn(DataAssociator):
 
     Construct a TPR-tree.
     """
-    measurement_model = Property(
-        MeasurementModel,
+    measurement_model: MeasurementModel = Property(
         doc="Measurement model used within the TPR tree")
-    horizon_time = Property(
-        datetime.timedelta,
+    horizon_time: datetime.timedelta = Property(
         doc="How far the TPR tree should look into the future")
-    vel_mapping = Property(
-        np.ndarray,
+    vel_mapping: Sequence[int] = Property(
         default=None,
         doc="Used to generate coordinates")
 
--- a/stonesoup/dataassociator/tree.py
+++ b/stonesoup/dataassociator/tree.py
@@ -1,0 +1,176 @@
+# -*- coding: utf-8 -*-
+import datetime
+from collections import defaultdict
+from operator import attrgetter
+
+import numpy as np
+import scipy as sp
+import rtree
+from scipy.spatial import KDTree
+
+
+from .base import DataAssociator
+from ..base import Property
+from ..models.base import LinearModel
+from ..models.measurement import MeasurementModel
+from ..predictor import Predictor
+from ..types.update import Update
+from ..updater import Updater
+
+
+class DetectionKDTreeMixIn(DataAssociator):
+    """Detection kd-tree based mixin
+
+    Construct a kd-tree from detections and then use a :class:`~.Predictor` and
+    :class:`~.Updater` to get prediction of track in measurement space. This is
+    then queried against the kd-tree, and only matching detections are passed
+    to the :attr:`hypothesiser`.
+
+    Notes
+    -----
+    This is only suitable where measurements are in same space as each other
+    and at the same timestamp.
+    """
+    predictor = Property(
+        Predictor,
+        doc="Predict tracks to detection times")
+    updater = Property(
+        Updater,
+        doc="Updater used to get measurement prediction")
+    number_of_neighbours = Property(
+        int, default=None,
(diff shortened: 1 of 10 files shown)


sglvladi:
Looks good!

Seeing how the introduced classes only modify the newly introduced `generate_hypotheses()` method, it might be more appropriate to define these under a `Gater` class that operates on the full set of hypotheses generated for all tracks. 

The above relates to discussions during the Liverpool Sprint about modifying the `Hypothesisers` to operate on the complete set of tracks, rather than on a track-by-track basis (by moving `generate_hypotheses()` inside the `Hypothesiser`). If this was done, then `Gaters` could be used to wrap `Hypothesisers` in a similar manner to how `Feeders` are used to wrap `Detectors`.

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # Create tree
+        tree_property = rtree.index.Property(
+            type=rtree.index.RT_TPRTree,
+            tpr_horizon=self.horizon_time.total_seconds(),
+            dimension=len(self.pos_mapping))
+        self._tree = rtree.index.RtreeContainer(properties=tree_property)
+        self._coords = dict()
+
+    def _track_tree_coordinates(self, track):
```
sglvladi: In cases where the measurement noise is much greater than the process/dynamic noise, and the time window is relatively small, the tracks seem to miss detections very often. 

One way of alleviating the above would be to consider the measurement noise covariance when computing `state_delta`, e.g.:
```python
state_delta = 3 * np.sqrt(
            np.diag(track.covar)[self.pos_mapping].reshape(-1, 1))
meas_delta = 3 * np.sqrt(np.diag(self.measurement_model.covar()).reshape(-1, 1))
state_delta += meas_delta
```

However, this may not necessarily be correct when it comes to individual detections, that carry their own measurement model.

sdhiscocks: Currently the intersection query only uses the measurement as is, passing it's state vector in as min and max. This could be changed to use the measurement model error similar to track and this would then allow measurements with different models/errors.

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,176 @@
...
+        for track in sorted(tracks.union(self._tree),
+                            key=attrgetter('timestamp')):
+            if c_time is None:
+                c_time = track.timestamp
+            if track not in self._tree:
+                self._coords[track] = self._track_tree_coordinates(track)
+                self._tree.insert(track, self._coords[track])
+            elif track not in tracks:
+                coords = self._coords[track][:-1] \
+                    + ((self._coords[track][-1], c_time.timestamp()),)
```
sdhiscocks, line 139: Note `timestamp()` will be in local time, and cause issues with changing time zones. Should be converted to UTC first.

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,176 @@
...
+        doc="Number of neighbours to find. Default `None`, which means all "
+            "points within the :attr:`max_distance` are returned.")
+    max_distance = Property(
+        float, default=np.inf,
+        doc="Max distance to return points. Default `inf`")
+
+    def generate_hypotheses(self, tracks, detections, time, **kwargs):
+        # No need for tree here.
+        if not tracks:
+            return set()
```
sdhiscocks, line 51: This should return an empty dictionary.

jswright-dstl (approved):

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,196 @@
...
+    updater = Property(
+        Updater,
+        doc="Updater used to get measurement prediction")
+    number_of_neighbours = Property(
+        int, default=None,
+        doc="Number of neighbours to find. Default `None`, which means all "
+            "points within the :attr:`max_distance` are returned.")
+    max_distance = Property(
+        float, default=np.inf,
+        doc="Max distance to return points. Default `inf`")
```
oharrald-Dstl, line 50: Will need to change to new Property format.

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,196 @@
...
+    measurement_model = Property(
+        MeasurementModel,
+        doc="Measurement model used within the TPR tree")
+    horizon_time = Property(
+        datetime.timedelta,
+        doc="How far the TPR tree should look into the future")
+    vel_mapping = Property(
+        np.ndarray,
+        default=None,
+        doc="Used to generate coordinates")
```
oharrald-Dstl, line 104: As above, change format.

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,196 @@
...
+                    meas_pred.state_vector.ravel(),
+                    r=self.max_distance)
+            else:
+                _, indexes = tree.query(
+                    meas_pred.state_vector.ravel(),
+                    k=self.number_of_neighbours,
+                    distance_upper_bound=self.max_distance)
+
+            for index in np.atleast_1d(indexes):
+                if index != len(detections_list):
```
oharrald-Dstl, line 82: Document why this check is needed

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,196 @@
...
+        doc="Used to generate coordinates")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # if no vel_mapping take position mapping and plus 1 to each dimension
+        # e.g. 0,2 would become 1,3
+        self.pos_mapping = self.measurement_model.mapping
+        if self.vel_mapping is None:
+            self.vel_mapping = self.measurement_model.mapping[1::3]
```
oharrald-Dstl, line 113: Is that really what is being done here?

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,196 @@
...
+        else:
+            self.pos_mapping = self.measurement_model.mapping
```
oharrald-Dstl, line 115: Not needed

stonesoup/dataassociator/tree.py
```diff
@@ -0,0 +1,196 @@
...
+        min_pos = (state_vector - state_delta).ravel()
+        max_pos = (state_vector + state_delta).ravel()
+        min_vel = (vel_vector - vel_delta).ravel()
+        max_vel = (vel_vector + vel_delta).ravel()
+
+        return ((*min_pos, *max_pos), (*min_vel, *max_vel),
+                track.timestamp.astimezone(datetime.timezone.utc).timestamp())
+
+    def generate_hypotheses(self, tracks, detections, timestamp, **kwargs):
+        # No need for tree here.
```
oharrald-Dstl, line 142: Wouldn't mind some documentation here to talk through process/reasoning.

oharrald-Dstl (approved):

PR #147, opened 2019-11-12
Passive sensor

rjgreen-dstl:
Adds a bearings only sensor class which will be required in order to undertake bearings only tracking.

### DIFF
--- a/stonesoup/sensor/base.py
+++ b/stonesoup/sensor/base.py
@@ -17,7 +17,7 @@ class Sensor(Base):
         MeasurementModel, default=None, doc="Measurement model")
 
     @abstractmethod
-    def gen_measurement(**kwargs):
+    def measure(**kwargs):
         """Generate a measurement"""
         raise NotImplementedError
 
@@ -42,7 +42,7 @@ class Sensor3DCartesian(Sensor):
             origin")
 
     @abstractmethod
-    def gen_measurement(**kwargs):
+    def measure(**kwargs):
         """Generate a measurement"""
         raise NotImplementedError
 
--- a/stonesoup/sensor/base.py
+++ b/stonesoup/sensor/base.py
@@ -2,7 +2,6 @@
 from abc import abstractmethod
 
 from ..base import Base, Property
-from ..models.measurement import MeasurementModel
 from ..types.state import StateVector
 
 
@@ -12,9 +11,6 @@ class Sensor(Base):
     A sensor object that operates according to a given
     :class:`~.MeasurementModel`.
     """
-
-    measurement_model = Property(
-        MeasurementModel, default=None, doc="Measurement model")
 
     @abstractmethod
     def measure(**kwargs):
@@ -45,11 +41,3 @@ class Sensor3DCartesian(Sensor):
     def measure(**kwargs):
         """Generate a measurement"""
         raise NotImplementedError
-
-    def set_position(self, position):
-        self.position = position
-        self.measurement_model.translation_offset = position
-
-    def set_orientation(self, orientation):
-        self.orientation = orientation
-        self.measurement_model.rotation_offset = orientation
--- a/stonesoup/sensor/base.py
+++ b/stonesoup/sensor/base.py
@@ -3,6 +3,7 @@ from abc import abstractmethod
 
 from ..base import Base, Property
 from ..models.measurement import MeasurementModel
+from ..types.state import StateVector
 
 
 class Sensor(Base):
@@ -19,3 +20,29 @@ class Sensor(Base):
     def gen_measurement(**kwargs):
         """Generate a measurement"""
         raise NotImplementedError
+
+class Sensor3DCartesian(Sensor):
+    """Sensor base class extended to include 3D cartesian motion
+
+
+    """
+    position = Property(StateVector,
+                        doc="The sensor position on a 3D Cartesian plane,\
+                                expressed as a 3x1 array of Cartesian coordinates\
+                                in the order :math:`x,y,z`")
+    orientation = Property(
+        StateVector,
+        doc="A 3x1 array of angles (rad), specifying the sensor orientation in \
+               terms of the counter-clockwise rotation around each Cartesian \
+               axis in the order :math:`x,y,z`. The rotation angles are positive \
+               if the rotation is in the counter-clockwise direction when viewed \
+               by an observer looking along the respective rotation axis, \
+               towards the origin")
+
+    def set_position(self, position):
+        self.position = position
+        self.measurement_model.translation_offset = position
+
+    def set_orientation(self, orientation):
+        self.orientation = orientation
+        self.measurement_model.rotation_offset = orientation
--- a/stonesoup/sensor/base.py
+++ b/stonesoup/sensor/base.py
@@ -21,6 +21,7 @@ class Sensor(Base):
         """Generate a measurement"""
         raise NotImplementedError
 
+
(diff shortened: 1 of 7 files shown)


stonesoup/sensor/passive.py
```diff
@@ -0,0 +1,92 @@
...
+        -------
+        :class:`~.Detection`
+            A measurement generated from the given state. The timestamp of the\
+            measurement is set equal to that of the provided state.
+        """
+
+        measurement_vector = self.measurement_model.function(
+            ground_truth.state_vector, noise=noise, **kwargs)
+
+        model_copy = copy.copy(self.measurement_model)
```
jmbarr: Why copy?

stonesoup/sensor/passive.py
```diff
@@ -0,0 +1,92 @@
...
+            ndim_state=ndim_state,
+            mapping=mapping,
+            noise_covar=noise_covar,
+            translation_offset=position,
+            rotation_offset=orientation)
+
+        super().__init__(position, orientation, ndim_state, mapping,
+                         noise_covar, *args, measurement_model, **kwargs)
+
+    def set_position(self, position):
```
jmbarr: In general don't like setters where the parameters aren't protected. You could just use obj.position = position. But some like it. Is there a policy?

stonesoup/sensor/passive.py
```diff
@@ -0,0 +1,92 @@
...
+        self.position = position
+        self.measurement_model.translation_offset = position
+
+    def set_orientation(self, orientation):
+        self.orientation = orientation
+        self.measurement_model.rotation_offset = orientation
+
+    def gen_measurement(self, ground_truth, noise=None, **kwargs):
```
jmbarr: Would prefer :attr:`measure()`

stonesoup/sensor/passive.py
rjgreen-dstl: These setters are used by the platform class in order to move the sensors with the platform, therefore not protected.

stonesoup/sensor/passive.py
rjgreen-dstl: good question, taken from the earlier work by @sglvladi . Because the measurement_model is passed for each detection i don't think it needs to be?

jmbarr: As it's (IMO an ugly) way of dealing with the issue of passing-by-reference changing the original variable, perhaps just remove it and check that `self.measurement_model` isn't affected?

stonesoup/sensor/base.py
```diff
@@ -19,3 +20,29 @@ class Sensor(Base):
     def gen_measurement(**kwargs):
```
jmbarr: For another time: 
Can we just call it `measure(**kwargs)` or `sense(**kwargs)`? Seems cleaner, more apposite.

gawebb-dstl:
For line 17/18 in the test_passive.py, can it be shortened from
`StateVector( np.array(([[1], [1], [0]])))`
to
`StateVector([[1], [1], [0]])`

stonesoup/sensor/base.py
```diff
@@ -19,3 +20,36 @@ class Sensor(Base):
...
+    orientation = Property(
+        StateVector,
+        doc="A 3x1 array of angles (rad), specifying the sensor \
+            orientation in terms of the counter-clockwise rotation \
+            around each Cartesian axis in the order :math:`x,y,z`. The \
+            rotation angles are positive if the rotation is in the \
+            counter-clockwise direction when viewed by an observer \
+            looking along the respective rotation axis, towards the \
+            origin")
+
```
gawebb-dstl: Does it matter what coordinate system is used? (East, North, Up or North, East, Down,  etc)

stonesoup/sensor/tests/test_passive.py
```diff
@@ -0,0 +1,59 @@
...
+from ..passive import PassiveElevationBearing
+
+
+def test_passive_sensor():
+    # Input arguments
+    # TODO: pytest parametarization
+    noise_covar = CovarianceMatrix(np.array([[np.deg2rad(0.015), 0],
+                                             [0, np.deg2rad(0.1)]]))
+    detector_position = StateVector(
+        np.array(([[1], [1], [0]])))
```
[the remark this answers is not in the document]
DaveKirkland: Actually - with the changes I made to the StateVector (previous update) you can do
`StateVector([1, 1, 0])`

Internally it converts it to a 3x1 array.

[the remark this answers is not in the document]
rjgreen-dstl: Thanks, i will make these modifications to the pull request before it gets merged.

stonesoup/sensor/passive.py
rjgreen-dstl: Removed

stonesoup/sensor/base.py
rjgreen-dstl: modified to measure - i think it makes it clearer and it was only a few modifications. Suggest people have a look and decide if this is a change we want to make or to remain with gen_measurement

stonesoup/sensor/passive.py
```diff
@@ -0,0 +1,69 @@
...
+        :class:`~.Detection`
+            A measurement generated from the given state. The timestamp of the\
+            measurement is set equal to that of the provided state.
+        """
+
+        measurement_vector = self.measurement_model.function(
+            ground_truth.state_vector, noise=noise, **kwargs)
+
+        return Detection(measurement_vector,
+                         measurement_model=self.measurement_model,
```
sdhiscocks: I'd suggest creating a new instance of the measurement model here, rather than copying or using `self.measurement_model`. This way you remove the need for the init method, as well as the _set_position_ and _set_orientation_ methods.

I also think we should remove the _measurement_model_ property from the sensor, as so far none of the sensors use it.

rjgreen-dstl: Changes made to remove _measurement_model_ from sensor.

stonesoup/sensor/passive.py
rjgreen-dstl: Removed the setters as part of the modification to remove the _measurement_model_ object in the sensor class

stonesoup/sensor/passive.py
```diff
@@ -0,0 +1,63 @@
...
+            model")
+    mapping = Property(
+        [np.array], doc="Mapping between the targets state space and the\
+                        sensors measurement capability")
+    noise_covar = Property(CovarianceMatrix,
+                           doc="The sensor noise covariance matrix. This is \
+                                utilised by (and follow in format) the \
+                                underlying \
+                                :class:`~.CartesianToElevationBearing`\
+                                model")
```
sdhiscocks: There is a mix of different styles being used here and other places which looks messy. Would be good to stick to one standard. Suggest one used for _ndim_state_ and _orientation_ as give you more room for _doc_.

stonesoup/sensor/radar.py
```diff
@@ -120,14 +91,11 @@ class RadarRotatingRangeBearing(RadarRangeBearing):
     fov_angle = Property(
         float, doc="The radar field of view (FOV) angle (in radians).")
 
-    def __init__(self, position, orientation, ndim_state, mapping, noise_covar,
-                 dwell_center, rpm, max_range, fov_angle, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
```
sdhiscocks: This is no longer required.

PR #148, opened 2019-11-21
Fix for KalmanFilter when using multiple models

sdhiscocks:
This was ignoring the measurement matrix when using measurement models per detection.


sdhiscocks:
> You're correct that line 112 needs changing, but to 'return self._check_measurement_model(measurement_model).matrix(**kwargs)'. Your addition will throw an error if measurement_model isn't passed as a parameter.
> 
> _check_measurement_model() is designed to return the measurement model passed unless it's None, at which point it uses the one held in the Updater.
> 
> Without line 262 you'll get an error when trying to use the Updater's measurement model

So looks like `_measurement_matrix()` is only called from `predicted_measurement()`, where `_check_measurement_model()` has already been used. So `measurement_model` in `_measurement_matrix()` will never be `None`

jmbarr:
> > You're correct that line 112 needs changing, but to 'return self._check_measurement_model(measurement_model).matrix(**kwargs)'. Your addition will throw an error if measurement_model isn't passed as a parameter.
> > _check_measurement_model() is designed to return the measurement model passed unless it's None, at which point it uses the one held in the Updater.
> > Without line 262 you'll get an error when trying to use the Updater's measurement model
> 
> So looks like `_measurement_matrix()` is only called from `predicted_measurement()`, where `_check_measurement_model()` has already been used. So `measurement_model` in `_measurement_matrix()` will never be `None`

Unless you want to query the matrix directly for other reasons (I can't think of right now.) In which case it's worth making the method robust to missing measurement models.

sdhiscocks:
> > > You're correct that line 112 needs changing, but to 'return self._check_measurement_model(measurement_model).matrix(**kwargs)'. Your addition will throw an error if measurement_model isn't passed as a parameter.
> > > _check_measurement_model() is designed to return the measurement model passed unless it's None, at which point it uses the one held in the Updater.
> > > Without line 262 you'll get an error when trying to use the Updater's measurement model
> > 
> > 
> > So looks like `_measurement_matrix()` is only called from `predicted_measurement()`, where `_check_measurement_model()` has already been used. So `measurement_model` in `_measurement_matrix()` will never be `None`
> 
> Unless you want to query the matrix directly for other reasons (I can't think of right now.) In which case it's worth making the method robust to missing measurement models.

So as the method is "private" I think we should wait for a use case rather than putting in additional defensive code for currently unlikely scenario. Removing it also makes the behaviour consistent between both Kalman and Extended Kalman versions of the method.

sdhiscocks:
I've updated this as I noticed that the Unscented Kalman Filter also had this issue.

stonesoup/updater/kalman.py
```diff
@@ -109,7 +109,7 @@ def _measurement_matrix(self, predicted_state=None, measurement_model=None,
             The measurement matrix, :math:`H_k`
 
         """
-        return self.measurement_model.matrix(**kwargs)
+        return measurement_model.matrix(**kwargs)
```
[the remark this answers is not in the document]
jmbarr: The reason it's keyword is to cater for those situations where you want to use a measurement model contained within the Updater, rather than having one attached to each detection.

jmbarr:
> > > > You're correct that line 112 needs changing, but to 'return self._check_measurement_model(measurement_model).matrix(**kwargs)'. Your addition will throw an error if measurement_model isn't passed as a parameter.
> > > > _check_measurement_model() is designed to return the measurement model passed unless it's None, at which point it uses the one held in the Updater.
> > > > Without line 262 you'll get an error when trying to use the Updater's measurement model
> > > 
> > > 
> > > So looks like `_measurement_matrix()` is only called from `predicted_measurement()`, where `_check_measurement_model()` has already been used. So `measurement_model` in `_measurement_matrix()` will never be `None`
> > 
> > 
> > Unless you want to query the matrix directly for other reasons (I can't think of right now.) In which case it's worth making the method robust to missing measurement models.
> 
> So as the method is "private" I think we should wait for a use case rather than putting in additional defensive code for currently unlikely scenario. Removing it also makes the behaviour consistent between both Kalman and Extended Kalman versions of the method.

Which line are you referring to?

sdhiscocks:
> > > > > You're correct that line 112 needs changing, but to 'return self._check_measurement_model(measurement_model).matrix(**kwargs)'. Your addition will throw an error if measurement_model isn't passed as a parameter.
> > > > > _check_measurement_model() is designed to return the measurement model passed unless it's None, at which point it uses the one held in the Updater.
> > > > > Without line 262 you'll get an error when trying to use the Updater's measurement model
> > > > 
> > > > 
> > > > So looks like `_measurement_matrix()` is only called from `predicted_measurement()`, where `_check_measurement_model()` has already been used. So `measurement_model` in `_measurement_matrix()` will never be `None`
> > > 
> > > 
> > > Unless you want to query the matrix directly for other reasons (I can't think of right now.) In which case it's worth making the method robust to missing measurement models.
> > 
> > 
> > So as the method is "private" I think we should wait for a use case rather than putting in additional defensive code for currently unlikely scenario. Removing it also makes the behaviour consistent between both Kalman and Extended Kalman versions of the method.
> 
> Which line are you referring to?

https://github.com/dstl/Stone-Soup/blob/4e98c0f3c4a334dcb6c097726114ccb378762a03/stonesoup/updater/kalman.py#L138-L145

stonesoup/updater/kalman.py
```diff
@@ -355,9 +328,12 @@ def predict_measurement(self, predicted_state, measurement_model=None):
             gauss2sigma(predicted_state.state_vector, predicted_state.covar,
                         self.alpha, self.beta, self.kappa)
 
+        def measurement_function_nonoise(state_vector, noise=0, **kwargs):
+            return measurement_model.function(state_vector, noise, **kwargs)
+
```
jmbarr: Is it intentional that this becomes not a private function?

jmbarr: Ah, missed the indent - it's a function that appears only in predict_measurement()

jmbarr: Course, it wouldn't be a problem if the default function() didn't add noise

sdhiscocks: Yes. Think we need to address that in another PR, as will require changing a number of models.

stonesoup/updater/kalman.py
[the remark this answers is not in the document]
sdhiscocks: See https://github.com/dstl/Stone-Soup/pull/148#issuecomment-557099891

PR #150, opened 2019-11-25
Video processing classes

sglvladi:
Summary
----------
This PR adds video processing classes to Stone Soup:

- Defined an `ImageFrame` class that encapsulates information about a frame (`pixels`, `timestamp`)
- Defined a `UrlReader` reader base class, to be used for readers that access abstract URLs.
- Defined a `FrameReader` base class, to be used for readers that return `ImageFrame` objects.
- Defined a `VideoClipReader` that replays video clips from files (e.g. `.avi`, `.mp4`) to generate frames.
- Defined a `FFmpegVideoStreamReader` that generates frames from stream urls (e.g. RTSP). 

To showcase how the introduced classes can be used, examples have temporarily been added in the [examples/video folder](https://github.com/sglvladi/Stone-Soup/tree/video_processing/examples/video).

TODO
-------
- [x] Add tests
- [x] Remove examples

Notes
------
The choice has been made to not use `OpenCV` as the basis for reading files and/or streams. Instead, [`MoviePy`](https://zulko.github.io/moviepy/) has been used for reading files and [`ffmpeg-python`](https://github.com/kkroening/ffmpeg-python) has been used for reading streams. This has been done for the following reasons:
-  Extracting the timestamp/duration of frames in replayed video from files with `OpenCV` is tricky. On the contrary, `MoviePy` allows for this to be done easily.
- Usage of `MoviePy` allows for the application of clip transformations and effects, as per the [MoviePy documentation](https://zulko.github.io/moviepy/getting_started/effects.html). Upon instantiation, the underlying MoviePy `VideoFileClip` instance can be accessed through the `VideoClipReader.clip` property. This can then be used as expected, e.g.:
    ```python
    # Rearrange RGB to BGR
    def arrange_bgr(image):
        return image[:, :, [2, 1, 0]]

    reader = VideoClipReader("path_to_file")
    reader.clip = reader.clip.fl_image(arrange_rgb)

    for timestamp, frame in reader:
        # The generated frame.pixels will now
        # be arranged in BGR format.
        ...
    ```
- When using `OpenCV` to read video streams (e.g. RTSP) a delay was observed in the replayed stream. This seems to be due to internal buffering performed by whatever decoder is used by `OpenCV`. To overcome this issue, the `FFmpegVideoStreamReader` makes direct use of `FFmpeg` to decode the stream, therefore allowing for direct parameterisation of `FFmpeg`. 
- By default, `FFmpeg` also performs internal buffering of frames. Disabling the buffer can be achieved by setting ``input_opts={'threads': 1, 'fflags': 'nobuffer'}`` when instantiating a `FFmpegVideoStreamReader` , e.g: .
    ``` python
        video_reader = FFmpegVideoStreamReader('rtsp://192.168.0.10:554/1/h264minor',
                                               input_opts={'threads': 1, 'fflags': 'nobuffer'})
        for timestamp, frame in video_reader:
            ....
    ```


stonesoup/reader/video.py
```diff
@@ -0,0 +1,218 @@
...
+        ------
+        : :class:`datetime.datetime`
+            Datetime of current time step
+        : set of :class:`~.ImageFrame`
+            Generated frame in the time step
+        """
+        raise NotImplementedError
+
+    @BufferedGenerator.generator_method
+    def sensor_data_gen(self):
```
sglvladi: I am not too sure if this is correct/desired. 

My reasoning is that `FrameReader` objects generate frames (i.e. `ImageFrames`), rather than generic sensor data (i.e. `SensorData`). Therefore it made sense to have a `frame` property and `frames_gen()` method, in the same way `DetectionReaders` have `detections` and `detections_gen()`. 

Based on the above definitions, readers derived from `FrameReader` would only need to defined the `frame_gen()` method (as is done in the newly introduced `VideoClipReader` and `FFmpegVideoStreamReader` classes).

setup.py
```diff
@@ -26,7 +26,8 @@
       ],
       packages=find_packages(exclude=('docs', '*.tests')),
       install_requires=[
-          'ruamel.yaml>=0.15.45', 'scipy', 'matplotlib', 'utm', 'pymap3d'],
+          'ruamel.yaml>=0.15.45', 'scipy', 'matplotlib', 'utm', 'pymap3d',
+          'ffmpeg-python', 'moviepy'],
```
sdhiscocks: I think we should make these optional dependencies, adding them to as a new category in `extra_required` below. 

Need to think about how we handle this on import when not installed. Probably just wrap import with `try`/`except ImportError` which prints out dependencies required.

stonesoup/reader/video.py
sdhiscocks: This makes sense, as sensor data is more unique compared to other elements of Stone Soup, as you need a matching detector to make "sense" of it.

stonesoup/reader/video.py
```diff
@@ -0,0 +1,220 @@
...
+        to) :py:meth:`~frames_gen`.
+
+        Yields
+        ------
+        : :class:`datetime.datetime`
+            Datetime of current time step
+        : set of :class:`~.ImageFrame`
+            Generated frame in the time step
+        """
+        return self.frames_gen()
```
sdhiscocks: I think this should be `yield from self.frames_gen()`.

sglvladi: Done

setup.py
sglvladi: Done 👍

sglvladi:
@sdhiscocks, @rjgreen-dstl 

I have tried to make all suggested changes. Please have another look and let me know what you think.

stonesoup/reader/video.py
```diff
@@ -0,0 +1,227 @@
...
+        to) :py:meth:`~frames_gen`.
+
+        Yields
+        ------
+        : :class:`datetime.datetime`
+            Datetime of current time step
+        : set of :class:`~.ImageFrame`
+            Generated frame in the time step
+        """
+        yield self.frames_gen()
```
sdhiscocks: I think this should be `yield from self.frames_gen()`

stonesoup/reader/video.py
```diff
@@ -0,0 +1,227 @@
...
+from abc import abstractmethod
+import datetime
+import numpy as np
+try:
+    import ffmpeg
+    import moviepy.editor as mpy
+except ImportError:
+    print("Usage of video processing classes requires that the optional"
+          "package dependencies 'moviepy' and 'ffmpeg-python' are installed. "
+          "This can be achieved by running 'python -m pip install .[video]'")
```
sdhiscocks: I'm thinking this  should be `python -m pip install stonesoup[video]`, as it'll allow those who aren't installing from local clone to install.

Also, I'm wondering if this message should be part of an "ImportError" rather than a separate `print` statement. Its a bit clearer that there's an error, and also goes to _stderr_ rather than _stdout_. Something like:
```python
except ImportError as error:
    raise ImportError(
          "Usage of video processing classes requires that the optional"
          "package dependencies 'moviepy' and 'ffmpeg-python' are installed. "
          "This can be achieved by running 'python -m pip install .[video]'") \
          from error

```

PR #151, opened 2019-11-25
Add Multi-detector Feeder class

sdhiscocks:
This enables multiple detectors to feed into single tracker, e.g. for a multi-sensor tracking scenario.

### DIFF
--- a/docs/source/stonesoup.feeder.rst
+++ b/docs/source/stonesoup.feeder.rst
@@ -19,6 +19,12 @@ Geo
 .. automodule:: stonesoup.feeder.geo
     :show-inheritance:
 
+Multi
+-----
+
+.. automodule:: stonesoup.feeder.multi
+    :show-inheritance:
+
 Time Based
 ----------
 
--- a/stonesoup/feeder/multi.py
+++ b/stonesoup/feeder/multi.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+from .base import Feeder
+from ..base import Property
+from ..buffered_generator import BufferedGenerator
+from ..detector import Detector
+
+
+class MultiDetectorFeeder(Feeder):
+    """Multi-detector Feeder
+
+    This returns detections from multiple detectors as a single stream,
+    yielding from the detector yielding the lowest timestamp first.
+    """
+    detector = None
+    detectors = Property([Detector], doc="Detectors to yield from")
+
+    @BufferedGenerator.generator_method
+    def detections_gen(self):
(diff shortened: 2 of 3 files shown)


DaveKirkland:
If we are feeding detections in from different sensors, they are likely to have different measurement covariances. Is there a way to connect the detections from a particular sensor to a particular Updater? This would also be necessary to handle the case where the detections/measurements have different components e.g A measurement from one sensor has Range, Bearing, Elevation while another sensor has Range, Bearing, Elevation, and Doppler. Each measurement would need it's own Updater, or we would need an Updater that can handle both measurements and likely just call the correct updater.
Not sure if that this is something that needs to be addressed within this pull request, but maybe an issue that needs to be addressed elsewhere?

sdhiscocks:
> If we are feeding detections in from different sensors, they are likely to have different measurement covariances. Is there a way to connect the detections from a particular sensor to a particular Updater? This would also be necessary to handle the case where the detections/measurements have different components e.g A measurement from one sensor has Range, Bearing, Elevation while another sensor has Range, Bearing, Elevation, and Doppler. Each measurement would need it's own Updater, or we would need an Updater that can handle both measurements and likely just call the correct updater.
> Not sure if that this is something that needs to be addressed within this pull request, but maybe an issue that needs to be addressed elsewhere?

So this is currently addressed by attaching a _measurement model_ to each detection. The _Updater_ will then use this model, falling back to the one provided to the _Updater_ if detection doesn't have it's own one. This feature is also required for example in a radar/bearing sensor, where it's position is changing, as each detection will have a measurement model with a different translations offset.

DaveKirkland:
> > If we are feeding detections in from different sensors, they are likely to have different measurement covariances. Is there a way to connect the detections from a particular sensor to a particular Updater? This would also be necessary to handle the case where the detections/measurements have different components e.g A measurement from one sensor has Range, Bearing, Elevation while another sensor has Range, Bearing, Elevation, and Doppler. Each measurement would need it's own Updater, or we would need an Updater that can handle both measurements and likely just call the correct updater.
> > Not sure if that this is something that needs to be addressed within this pull request, but maybe an issue that needs to be addressed elsewhere?
> 
> So this is currently addressed by attaching a _measurement model_ to each detection. The _Updater_ will then use this model, falling back to the one provided to the _Updater_ if detection doesn't have it's own one. This feature is also required for example in a radar/bearing sensor, where it's position is changing, as each detection will have a measurement model with a different translations offset.

Ok - that makes sense, I just hadn't seen that before or knew we could do that.

PR #155, opened 2019-12-13
Pass States not StateVectors

jmbarr:
Designed to modify the `function()` and `pdf()` methods in `Model` so that they take `State` rather than `StateVector` objects. This so that functions can access parameters not in the state vector.

### DIFF
--- a/stonesoup/dataassociator/tracktotrack.py
+++ b/stonesoup/dataassociator/tracktotrack.py
@@ -104,9 +104,9 @@ class EuclideanTrackToTrack(TrackToTrackAssociator):
 
                     distance = np.linalg.norm(
                         self.measurement_model_track1.function(
-                            state1.state_vector, noise=0)
+                            state1, noise=0)
                         - self.measurement_model_track2.function(
-                            state2.state_vector, noise=0))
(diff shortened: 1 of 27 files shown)


sdhiscocks:
There's a few places where we have test models that also need to be updated, as tests are passing where they shouldn't. For example: 
https://github.com/dstl/Stone-Soup/blob/48df0828ac6b43a5a178621f04b07c151315b64f/stonesoup/simulator/tests/conftest.py#L6-L32

jmbarr:
> There's a few places where we have test models that also need to be updated, as tests are passing where they shouldn't. For example:
> 
> https://github.com/dstl/Stone-Soup/blob/48df0828ac6b43a5a178621f04b07c151315b64f/stonesoup/simulator/tests/conftest.py#L6-L32

gah; who's writing bespoke transition models for tests? 

How to track these down?

sdhiscocks:
> How to track these down?

Looks like they are only present in the simulator tests.

jmbarr:
> > How to track these down?
> 
> Looks like they are only present in the simulator tests.

Done those. Any more?

PR #162, opened 2020-02-10
Avoid calculating measurement prediction for missed detections

sdhiscocks:
These aren't really required, as no comparison to measurement space is done. This also causes issues where no measurement model is provided for an updater, as reasonable use case when adding measurement models to detections.


sglvladi:
Looks good 👍

PR #163, opened 2020-02-11
Modify SinglePointInitiator to work with non-linear models

sdhiscocks:
This changes the `SinglePointInitiator` to use a Extended Kalman Updater, rather than the non-linear Kalman Filter. As the KF/EKF updater only differs on how the measurement matrix is fetch, this results in no change to the linear usage, but adds support to non-linear case.

### DIFF
--- a/stonesoup/initiator/tests/test_simple.py
+++ b/stonesoup/initiator/tests/test_simple.py
@@ -3,11 +3,12 @@ import datetime
 import numpy as np
 import pytest
 
+from ...models.base import LinearModel
 from ...models.measurement.linear import LinearGaussian
 from ...models.measurement.nonlinear import CartesianToBearingRange
 from ...models.transition.linear import \
     CombinedLinearGaussianTransitionModel, ConstantVelocity
-from ...updater.kalman import KalmanUpdater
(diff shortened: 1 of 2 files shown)


stonesoup/initiator/tests/test_simple.py
```diff
@@ -22,19 +23,24 @@
...
     prior_state = GaussianState(
         np.array([[0], [0]]),
         np.array([[100, 0], [0, 1]]))
 
-    # Define a measurement model
-    measurement_model = LinearGaussian(2, [0], np.array([[1]]))
-
     # Create the Kalman updater
-    kup = KalmanUpdater(measurement_model)
+    if isinstance(measurement_model, LinearModel):
```
jmbarr: You don't really need this though do you? In event of linear model `ExtendedKalmanUpdater` uses `KalmanUpdater` logic.

sdhiscocks: True, but as the test uses the `KalmanUpdater` and initiator uses `ExtendedKalmanUpdater` it helps to test that the output is the same and that is indeed True.

jmbarr: Ah, well done, +1 for test within a test.

Issue #161, opened 2020-02-08
Increase allowed line length


etfrogers-dstl:
Is it time to change to a longer line-length standard in Stone Soup?

The current 80 character lines seem to me (and a couple of others I have spoken) to negatively affect code readability. Many lines of code have to be wrapped several times and would be clearer on one or two lines. 

From PEP8: "Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters."

Bear in mind that was the position in 2001,and monitors have only got wider since then. 
Switching to 99 chars, or even pycharm's default of 120 seems like a good idea to me.

The switch is a few-line, single-file push (either `tox.ini`, `setup.cfg`, `.pep8` or `.flake8`) and is, of course, backward compatible. No old code needs to be changed, but new code would have more flexibility.

Do people have any comments?

sdhiscocks:
So even with screen real estate being larger, very long lines can be unreadable and can get messy are allows people to nest in more _if_ block, _for_ loops, etc. It also can become more difficult for side by side comparison, especially for 3-way merges (but these are rare to be fair).

We have a lot of equations and long variables names (which makes code more readable) that could really do with some more room, as they are becoming less readable by the time your forced to break them up over multiple lines. We also typically have the bulk of code in class methods, so lose 8 characters before you even get started.

There's always a balance to be had, but I think increasing to 99 characters would improve readability.

If we did make a change, I'd suggest we don't make one massive code change but allow the project to transition naturally as we make code changes.

jmbarr:
<meh>

jmbarr:
¯\_(ツ)_/¯

DaveKirkland:
I don't really have a problem with this. Can a local flake8 configuration file be distributed with the project, so that it passes the flake8 testing? Or include instructions for how to manually configure this ourselves,

etfrogers-dstl:
@DaveKirkland Yes, the flake8 configuration can be committed, if that what is wanted. I assume CircleCI would honour it (otherwise, we might have to configure that separately).

rjgreen-dstl:
closed as addressed in #166

PR #166, opened 2020-02-17
Increase allowed line length

etfrogers-dstl:
This PR addresses  #161. I appreciate that it has not been discussed at length and agreed, but this shows how it could be done (and will check that CircleCI honours the config). 

If people decide they are happy this can be pulled. If it gets objections, we can close this without pulling.


etfrogers-dstl:
Hmm, the failed checks are not due to flake8 errors, so CircleCI is working as I expected. They are due to a stochastic test failing (which it has not done before!). If re-run they will probably pass, but we will try and fix them so it doesn't re-occur. That will be a separate pull request.

PR #167, opened 2020-03-03
Property extensions

etfrogers-dstl:
This branch adds getter and setter options to Stone Soup `Property`s, analogous to those for Python `property`s. 

It also added the ability to make a `Property` read only, such that it is written once (in the parent object's constructor) and thereafter cannot be changed. 

Tests are also added to `test_declarative.py`

### DIFF
--- a/stonesoup/tests/test_declarative.py
+++ b/stonesoup/tests/test_declarative.py
@@ -180,3 +180,29 @@ def test_basic_getter():
     # now test when the default is used
     test_object_default = TestGetter()
     assert test_object_default.times_two == 10
+
+
+def test_readonly():
+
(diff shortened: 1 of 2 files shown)


stonesoup/tests/test_declarative.py
```diff
@@ -85,3 +85,124 @@ def test_non_base_property():
...
+
+
+def test_basic_setter():
+
+    class TestSetter(Base):
+        times_two = Property(float, default=5)
+
+        @times_two.setter
+        def set_times_two(self, value):
+            self._property_times_two = value * 2
```
dcrew-dstl: The decorators are a nice touch, very pythonic

PR #168, opened 2020-03-09
Added Longitude and Latitude types

sglvladi:
This PR adds Longitude and Latitude types that handle angle wrapping.

The types (and their respective mod functions) are essentially the same as the Bearing and Elevation types, but operate in degrees instead of radians.

Some minor typos have also been fixed in the ```mod_bearing()``` and ```mod_elevation()``` docstrings.


stonesoup/types/tests/test_angle.py
```diff
@@ -108,3 +109,92 @@ def test_misc(self, Class, func):
         assert b1 % 0.1 == b2 % 0.1
         assert +b1 == +b2
         assert approx(-b1) == -b2
+
+
+scenario3 = ('longitude', {'Class': Longitude, 'func': mod_longitude})
+scenario4 = ('latitude', {'Class': Latitude, 'func': mod_latitude})
+
+
+class TestAngleDegrees:
```
etfrogers-dstl: I would prefer `class_` here to `Class` here, as I think it may be less confusing.

sglvladi: @etfrogers-dstl Thanks for the review. Following discussions during the telecons, the PR will be modified so that it deals with radians, instead of degrees, so I would be grateful if you could have another look once this is done.

But, in relation to your comment, I have not preference on what keyword we use. This code was originally written by @DaveKirkland , so I have just copied and pasted his tests here. I don't mind correcting this if you/others find it appropriate.

stonesoup/types/tests/test_angle.py
```diff
@@ -23,10 +23,12 @@ def pytest_generate_tests(metafunc):
 
 scenario1 = ('bearing', {'Class': Bearing, 'func': mod_bearing})
 scenario2 = ('elevation', {'Class': Elevation, 'func': mod_elevation})
+scenario3 = ('longitude', {'Class': Longitude, 'func': mod_bearing})
```
etfrogers-dstl, line 26: I thought you were planning to change these to `class_`. Is that still the plan?

sglvladi: Correct, I just hadn't gotten around to doing it. Should be done now.

PR #170, opened 2020-03-09
Preserve State Vector types

sglvladi:
Summary
======

**NOTE: This PR is mostly intended as a point for discussion**.

The intention here was to ensure that state vector types are preserved during the recursion of predictors and updaters. For example, consider a State whose state vector that has an element of ```Bearing``` type, e.g.: 
```python
>>> sv = StateVector([Bearing(np.pi/2), 0.1])
>>> cov = CovarianceMatrix([[0.1, 0], [0, 1]])
>>> ts = datatime.now()
>>> prior = GaussianState(sv, cov, ts)
>>> prior.state_vector
StateVector([[Bearing(1.5707963267948966)],
             [0.1]], dtype=object)
```
Now, if the above state is currently passed to a predictor, the resulting state will no longer have the correct types, e.g.:
```python
>>> tx_model = ConstantVelocity(0.1)
>>> predictor = KalmanPredictor(tx_model)
>>> prediction = predictor.predict(prior, timestamp=ts+timedelta(seconds=1))
>>> prediction.state_vector
StateVector([[1.67079633],
             [0.1       ]])
```
The main reason this is happening is due the matrix multiplication performed in the ```tx_model.function()```. The same also applies to any calls to measurement model ```function()``` methods. 

Contribution
========

Therefore, the underlying contributions/ideas here are:
1. Introduce an optional property to all ```Models```, that will define the intended element types of the state vectors returned by such functions. This is the purpose of the introduced ```Model.sv_types``` property. 
2. Introduce a decorator that can be used to wrap all ```Model.function()``` methods to enforce casting of state vector element types. This is the purpose of the introduced ```@cast_sv_types``` decorator, which has been applied to ```LinearModel.function()```.
3. Introduce a function that can be used to match the element types of a state vector, to another (reference) vector. This is useful when the types are not directly available, but can be inferred from another state vector. An example of this is in the ```sigma2gauss()``` function, where the sigma points contain the correct types, but the resulting mean does not, due to the (matrix) multiplication necessary to compute the weighted sum. This is the purpose of the introduced ```match_sv_types()``` function, which has been applied in ```sigma2gauss()```.
4. Finally, make the necessary modifications to affected classes, e.g. ```KalmanPredictor``` and ```LinearGaussian```.

Usage
====
An example of how the above can be used is provided in the same context as above. The only difference is that the intended state vector types are now passed to the transition model on instantiation:

```python
>>> tx_model = ConstantVelocity(0.1, sv_types=[Bearing, np.float])
>>> predictor = KalmanPredictor(tx_model)
>>> prediction = predictor.predict(prior, timestamp=ts+timedelta(seconds=1))
>>> prediction.state_vector
StateVector([[Bearing(1.6707963267948962)],
             [0.1]], dtype=object)
```

and similarly in an updater:

```python
>>> mx_model = LinearGaussian(ndim_state=2, mapping=[0],
                              noise_covar=[[0.1]],
                              sv_types=[Bearing])
>>> updater = KalmanUpdater(mx_model)
>>> meas_prediction = updater.predict_measurement(prediction)
>>> meas_prediction.state_vector
StateVector([[Bearing(1.6707963267948962)]], dtype=object)

>>> dsv = StateVector([[1.6]])
>>> detection = Detection(dsv, timestamp=ts+timedelta(seconds=1))
>>> hyp = SingleHypothesis(prediction, detection, meas_prediction)
>>> posterior = updater.update(hyp)
>>> posterior.state_vector
StateVector([[Bearing(1.6057402427131002)],
             [0.03972745151245333]], dtype=object)
```

Conclusions
========
As stated in the summary, this is intended mostly as a point for discussion. I used the proposed approach to track ships that cross the +/- 180 Longitude boundary (using the Longitude/Latitude types of #168) and it seems to work well. Nevertheless, I suspect there may be suggestions for improvement and/or problems that I have not considered. 

Note
====
- The above does not introduce noticeable overheads when used in a Kalman context. However, if/when applied to a ParticleState, this will most likely not be very efficient. In this case, a vectorisation of the ```cast_sv_types``` would be necessary, if possible.
- The above may be affected by #155, as I expect it will most likely be merged first.


DaveKirkland:
I haven't taken a close look at this yet. One issue, sort of, is that the decorators aren't inherited. This means that when someone adds a new model as a subclass, they have to explicitly add in the decorators. It isn't necessarily a problem, but more of a gotcha.

I ran into the same sort of issue in adding a decorator to help with the model mixing for IMM and such.

sglvladi:
> I haven't taken a close look at this yet. One issue, sort of, is that the decorators aren't inherited. This means that when someone adds a new model as a subclass, they have to explicitly add in the decorators. It isn't necessarily a problem, but more of a gotcha.
> 
> I ran into the same sort of issue in adding a decorator to help with the model mixing for IMM and such.

@DaveKirkland that's a good point. I suppose that if we were to follow this route, we would have to clearly state somewhere that this is the case. I suspect that at some point we will add some DIY Tutorials to the StoneSoup docs, that will take users through a step-by-step process of defining custom components. If/when we do this, that would be a crucial point to mention.

oharrald-Dstl:
Is there a case in which we would NOT want state vector types to be preserved? Does it makes sense to have this type preservation in the base class `function`, rather than in decorators (consequently solving the decorator non-inheritance issue)? The option to not preserve state type could be introduced later.

sdhiscocks:
> Observation/comment: Recently I continue to experience learning with date aware time zones, the StoneSoup state datetime is not timezone aware, most would think UTC would be the standard is good. I am finding that I need to produce timezone aware for other reasons, database storage, etc., so possibly another state property to consider or option for state.
> 
> For now:
> state_timestamp_withtimezone = pytz.utc.localize(state.timestamp)

I don't think anywhere in the code (other than tests) we generate timestamps. So as long as the timestamps on your input data have timezones, it should work fine. We may want a way to explicitly add a timezone to  things like `CSVReader` however, as not always specified depending on string representation used. Suggest opening a separate issue for this @apiszcz.

DavidFCrouse:
Regarding time zones and the like,

> Observation/comment: Recently I continue to experience learning with date aware time zones, the StoneSoup state datetime is not timezone aware, most would think UTC would be the standard is good. I am finding that I need to produce timezone aware for other reasons, database storage, etc., so possibly another state property to consider or option for state.
> 
> For now:
> state_timestamp_withtimezone = pytz.utc.localize(state.timestamp)

I will disagree with this. While I see that Stone Soup makes extensive use of the datetime class, that is convenient, but not 100% ideal. It is fine for basic simulations, but you probably wouldn't want to model anything critical that way and timezones would just make it wore. As I understand it, the datetime class bases things on UTC. However, the leap seconds in UTC could potentially cause problems. For example, think of the timescale in GLONASS versus GPS. GPS reports a time and then a leap second offset. GLONASS reports UTC. However, when a leap second is added, that means that there is suddenly an extra second in the count from what you get from GLONASS. It can be handled, but it is inconvenient. The real issue is that if you use UTC, you must always keep your software up to date to keep track of the leap seconds.

I would say that an ideal timescale would be terrestrial time (TT), which is in the IERS standards. One can provide their own conversions to any other timescale as desired, included different timezones. There is a lot of overhead that comes with programming in timezones. Using a local time as a timestamp for something isn't great: how are you going to uniquely represent a time at the changes for daylight savings time? Daylight savings time is not at the same time in Europe and the United States. Some countries do not have it at all. Also, timezones themselves can be complicated by political boundaries.

The simplest thing would be to store the time in terrestrial time, which has no leap seconds. For conversion to other timescales, one can use something like the functions in the International Astronomical Union's Standards of Fundamental Astronomy. As for local time and the like, just let the user provide his own stuff. Do you really want to differentiate between things at a granular level, such as different counties in the U.S. state of Indiana observing different times? I think not.
https://en.wikipedia.org/wiki/Time_in_Indiana

That said, using terrestrial time would require bypassing the datetime class. For basic simulations, UTC suffices. Just don't be surprised if you get a blip when a leap second is added, depending on how you handle the times. I say, keep it the way it is and add a note about leapseconds and local time to the documentation.

apiszcz:
DavidFCrouse, thank you for the detailed description on options.

sdhiscocks:
This issue has come up again, and it would be good to get it resolved. Did you have time to rebase and remove WIP label?

PR #174, opened 2020-03-23
Add ID to Ground Truth Paths

sdhiscocks:


sdhiscocks:
> Not sure why conftest.py changes: it doesn't seem to touch the ids.

This is because the order of the paths is random, so they can switch order and such the id switches. making the test fail. Ideally need to update the test to support this, but for now shorting the test works, such only one path is written.

PR #177, opened 2020-03-30
Update to platform to allow it to control the positions of its sensors

etfrogers-dstl:
Update to platform to allow it to control the positions of its sensors rather than saving it to the sensor each time the platform moves.

As discussed with Rich Green (what seems like an age ago). 

Also an add_sensor method and tests (as requested by the todo in the code). 

Changes needed to Radar tests for the new platforms.


sdhiscocks:
One thing we could do to simplify things, is require that a sensor be mounted to a platform, and can't operate alone; and make the base Platform have sensors (having the option to set this to empty list).

The base platform could be stationary, and then have a subclass for a moving platform.

The position and velocity information should using mappings, such you have greater flexibility in models/dimensions that you use.

I think then you can also have position and velocity as an attribute that's dynamically returned (as appose to "get" methods) from the platform state using the position/velocity mappings on the platforms, and using the mappings and offsets on the sensor. Velocity can still be present on the base (stationary) platform, just returning _StateVector_ of zeros in the same dimensions as position mapping.

On mounting offsets, we should just make this a list of _StateVector_, that must be same length as number of sensors (default could still create list of all zero state vectors). The current description and logic is confusing.

Also, I'm not clear on the role of _add_sensor_? Are we expecting to add (remove?) sensors during simulation? Could just be covered by adding a sensor and offset to the platform if required?

etfrogers-dstl:
> One thing we could do to simplify things, is require that a sensor be mounted to a platform, and can't operate alone; and make the base Platform have sensors (having the option to set this to empty list).
The base platform could be stationary, and then have a subclass for a moving platform.

Intriguing. I don't see any reason it wouldn't work and I'd be happy to implement this if that is the consensus. 

> The position and velocity information should using mappings, such you have greater flexibility in models/dimensions that you use.

Agreed. I wondered about that, but it wasn't the approach used in the `Platform` originally, so I left as was. The original was less hacky, but when we needed velocity as well things got messy. I should have added a comment to the code that I thought my current approach was a bit of a hack. 

> I think then you can also have position and velocity as an attribute that's dynamically returned (as appose to "get" methods) from the platform state using the position/velocity mappings on the platforms, and using the mappings and offsets on the sensor. Velocity can still be present on the base (stationary) platform, just returning StateVector of zeros in the same dimensions as position mapping.

I think that should work. ~Overriding `property`s is not quite trivial, as I recall, but I managed it elsewhere.~ Sorry, only getting access to the superclass's `property` was the tricky bit. Only (minor) reason not too is that `property`s imply a cheap call, which is not necessarily true. 

> On mounting offsets, we should just make this a list of StateVector, that must be same length as number of sensors (default could still create list of all zero state vectors). The current description and logic is confusing.

Do you mean mounting mappings? Mounting offsets are implemented as you describe, but mappings are more complex. Assuming you do mean mappings, then yes, I can see the argument for simplifying this code. The tradeoff is more complicated client code slightly though, as I don't think zero defaults would work. The only sensible default I can see for _mappings_ is `[0::2]` which is pretty much as bad as my (hacky) code for `get_position`. 

Re-reading the comment, it also possible that you meant we should use `List[StateVector]` rather than nxN `ndarray`. In which case: Happy to go with that. It's always used by element anyway. 

> Also, I'm not clear on the role of add_sensor? Are we expecting to add (remove?) sensors during simulation? Could just be covered by adding a sensor and offset to the platform if required?

I have to say I added this because there was a TODO asking for it and it was easy when I was refactoring. The `add_sensor` code can be simplified if we have change `mounting_mapping` as described above. 

Without the `add_sensor()` method you would need to append to the list of sensors and concatenate properly to both the mapping and offset arrays. The risk of internal inconsistency would be high... You could end up with many errors on read, which might be hard to diagnose. I would therefore argue for keeping it. It's implemented (and tested) now. The only argument against is YAGNI, which is valid, but not sure it's worth deleting it just for that. Happy to yield the point though if others feel strongly

etfrogers-dstl:
On reflection: I think that `List[StateVector]` or single `StateVector` would be the way to go for `mounting_mappings` and `List[StateVector]` defaulting to zeros for the `mounting_offsets`. That's easier to understand and would avoid the nasty default problem for `mounting_mappings`.

etfrogers-dstl:
Still a WIP. Mostly needs documentation. Pushed to check the circle CI builds. I expected the docs not to build until #183 is merged.

stonesoup/models/measurement/nonlinear.py
```diff
@@ -90,9 +90,7 @@ def rvs(self, num_samples=1, **kwargs):
             return rvs_vectors.view(Matrix)
 
 
-class NonLinearGaussianMeasurement(MeasurementModel,
-                                   NonLinearModel,
-                                   GaussianModel):
+class NonLinearGaussianMeasurement(MeasurementModel, NonLinearModel, GaussianModel, ABC):
```
sdhiscocks, line 93: These are already abstract, so no need to inherit from `ABC` again.

stonesoup/platform/base.py
```diff
@@ -1,21 +1,143 @@
...
 from ..types.state import State
 from ..models.transition import TransitionModel
+import numpy as np
 
 
-class Platform(Base):
-    """Platform base class
+class Platform(Base, ABC):
+    state = Property(State, doc="The platform state at any given point")
+    mapping = Property(np.ndarray, doc="Mapping between platform position and state dims")
```
sdhiscocks, line 14: I'd call this `position_mapping`, as may want different mappings for velocity, acceleration or other elements of the state.

stonesoup/functions.py
```diff
@@ -566,3 +567,26 @@ def mod_elevation(x):
     elif N == 3:
         x = x - 2.0 * np.pi
     return x
+
+
+def coerce_to_valid_mapping(mapping: Union[List[int], np.ndarray, StateVector]):
```
sdhiscocks, line 572: Mappings really shouldn't be `StateVectors`. Should just be sequence of indexes.

stonesoup/platform/simple.py
```diff
@@ -43,90 +55,111 @@ def __init__(self, *args, **kwargs):
...
+                 isinstance(self.mounting_mappings[0], Sequence))):
+            # We were passed a non-empty list of arrays or lists
+            self.mounting_mappings = [coerce_to_valid_mapping(m) for m in self.mounting_mappings]
+        elif (isinstance(self.mounting_mappings, np.ndarray) or
+              isinstance(self.mounting_mappings, Sequence)):
+            # We were passed either list of non-arrays (assumed to be ints) or a single array, so
+            # coerce the single entry and then expand
+            # noinspection PyTypeChecker
+            single_mapping = coerce_to_valid_mapping(self.mounting_mappings)
+            self.mounting_mappings = [single_mapping] * len(self.sensors)
```
sdhiscocks, line 79: I still think it would be better to just required mountings to be list of mountings. How often would you actually have multiple sensors mounted at the exact same point?

stonesoup/models/measurement/nonlinear.py
etfrogers-dstl: Do you say "they are already abstract" because they have `@abstractmethod`s?

If so then I agree, but thought this was clearer to see if it was in the class definition as well. Happy to revert though.

stonesoup/platform/simple.py
etfrogers-dstl: Might you not have many sensors with negligibly different mountings, and hence want the all at (say) the nose of the platform? I agree that it complicates the code in this class though, so if it's a minor benefit, maybe better to enforce the List[mapping] approach.

stonesoup/models/measurement/nonlinear.py
sdhiscocks: The `BaseMeta` class inherits from `ABCMeta`, which `Base` and then all components are based from.

etfrogers-dstl: Yes, only reason is for readability then. If not useful (which it probably isn't) then I'll take it out again.

OTOH: to quote SO :)
> and also readability. Inheriting from ABC makes it clear to your readers what you're up to.

etfrogers-dstl: I'm easy on this one.

stonesoup/functions.py
etfrogers-dstl: Agreed. I think there are many issues with mappings in general that need to be thrashed out.

stonesoup/platform/simple.py
etfrogers-dstl: Actually: I think there is some confusion here (which is not a good sign!). `mounting_mapping` is a list of mappings between state vectors and positions (analogous to `mapping`/`position_mapping` in the platform). The single entry option was added because these would **almost always** be the same. 

I will review whether these are now necessary given the new architecture, and even if they are needed, perhaps the default that they are the same as the platform mapping is good enough, meaning the single StateVector option can be dropped.

`mounting_offsets` specify where on the platform the sensor is mounted, and must be a list (defaulting to a list of zero vectors).

stonesoup/platform/base.py
etfrogers-dstl: This is done and will be pushed

stonesoup/platform/simple.py
etfrogers-dstl: I have removed mounting mappings in a new version.

etfrogers-dstl:
Now ready for review :)

stonesoup/platform/base.py
```diff
@@ -1,21 +1,143 @@
...
+        # Return without moving static platforms
+        self.state.timestamp = timestamp
+
+
+class MovingPlatform(Platform):
+    """Moving platform base class
 
     A platform represents a random object defined as a :class:`~.State`
     that moves according to a given :class:`~.TransitionModel`.
     """
```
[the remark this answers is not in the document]
etfrogers-dstl: Good point. I will edit the docs to include this.

stonesoup/platform/base.py
```diff
@@ -1,21 +1,143 @@
...
-    state = Property(State, doc="The platform state at any given point")
     transition_model = Property(
         TransitionModel, doc="Transition model")
 
+    @property
+    def velocity(self):
+        # TODO docs
+        # TODO return zeros?
+        try:
+            return self.state_vector[self.mapping + 1]
```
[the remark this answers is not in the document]
etfrogers-dstl: Think this is on an outdated version of the code. Not sure how that happened, but this code has since changed.

stonesoup/platform/base.py
```diff
@@ -1,21 +1,143 @@
...
+        except IndexError:
+            raise AttributeError('Velocity is not defined for this platform')
+
+    @property
+    def acceleration(self):
+        # TODO docs
+        # TODO return zeros?
+        try:
+            return self.state_vector[self.mapping + 2]
```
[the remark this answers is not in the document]
etfrogers-dstl: Also outdated

stonesoup/platform/base.py
```diff
@@ -1,37 +1,223 @@
...
+
+        if self.ndim == 3:
+            _, bearing, elevation = cart2sphere(*velocity.flat)
+            return StateVector([0, bearing, elevation])
+        elif self.ndim == 2:
+            _, bearing = cart2pol(*velocity.flat)
+            return StateVector([0, bearing])
+
+    @property
+    def is_moving(self):
```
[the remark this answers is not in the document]
etfrogers-dstl: Done.

stonesoup/platform/base.py
```diff
@@ -1,37 +1,223 @@
...
+        """
+        # Note: a platform without a transition model can be given a velocity as part of it's
+        # StateVector. It just won't move
+        # This inconsistency is handled in the move logic
+        return np.any(self.velocity != 0)
+
+    def _set_position(self, value):
+        raise AttributeError('Cannot set the position of a moving platform')
+
     def move(self, timestamp=None, **kwargs):
```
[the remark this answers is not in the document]
etfrogers-dstl: Done

stonesoup/platform/base.py
```diff
@@ -1,37 +1,223 @@
...
+
+        Extracted from the state vector of the platform using the platform's
+        :attr:`position_mapping`. This property is settable for fixed platforms, but not for
+        movable ones, where the position must be set by moving the model with a transition model.
+        """
+        return self.state_vector[self.position_mapping]
+
+    @position.setter
+    def position(self, value: StateVector) -> None:
+        self._set_position(value)
```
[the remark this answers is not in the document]
sdhiscocks: Nevermind - I realise from reading comment where you've done this in Sensor also.

stonesoup/sensor/sensor.py
```diff
@@ -0,0 +1,51 @@
+from abc import ABC
+
+from ..platform.simple import FixedSensorPlatform
+from .base import BaseSensor
+from ..types.array import StateVector
+from ..types.state import State
+
+
+class Sensor(BaseSensor, ABC):
```
sdhiscocks, line 9: Why not remove `BaseSensor` and just have this?

stonesoup/functions.py
```diff
@@ -566,3 +567,26 @@ def mod_elevation(x):
     elif N == 3:
         x = x - 2.0 * np.pi
     return x
+
+
+def coerce_to_valid_mapping(mapping: Union[List[int], np.ndarray, StateVector]) -> np.ndarray:
+    """Function to take a mapping and convert it to a suitable form for indexing a
+    :class:`StateVector`. Needed because if you index a :class:`StateVector` with another
+    :class:`StateVector` (which would be the obvious technique) you get a 3d :class:`ndarray`.
```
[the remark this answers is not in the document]
etfrogers-dstl: Added a note to the StateVector docs.

stonesoup/platform/base.py
```diff
@@ -1,37 +1,223 @@
...
+
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def _set_position(self, value: StateVector) -> None:
+        raise NotImplementedError
+
+
+class FixedPlatform(Platform):
```
sdhiscocks, line 104: I'd be tempted to make this the base `Platform` class. Maybe even just having MovingPlatform as the base class, with transitional model optional?

stonesoup/platform/base.py
```diff
@@ -1,37 +1,223 @@
...
+
+        Equivalent (for this class) to ``all(v == 0 for v in self.velocity)``
+        """
+        # Note: a platform without a transition model can be given a velocity as part of it's
+        # StateVector. It just won't move
+        # This inconsistency is handled in the move logic
+        return np.any(self.velocity != 0)
+
+    def _set_position(self, value):
+        raise AttributeError('Cannot set the position of a moving platform')
```
sdhiscocks, line 203: Is there a reason to limit this?

stonesoup/platform/simple.py
```diff
@@ -183,16 +211,16 @@ def _get_rotation_matrix(vel):
...
+def _get_angle(vec: StateVector, axis: np.ndarray) -> float:
     """ Returns the angle between a pair of vectors. Used to determine the
     angle of rotation required between relative rectangular cartesian
     coordinate frame of reference and platform inertial frame of reference.
 
     Parameters
     ----------
-    vec : np.array
+    vec : StateVector
         1xD array denoting platform velocity
```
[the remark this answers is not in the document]
etfrogers-dstl: Fixed.

stonesoup/sensor/base.py
```diff
@@ -1,43 +1,87 @@
...
 
-    A sensor object that operates according to a given
-    :class:`~.MeasurementModel`.
+    .. warning::
+        This class is private and should not be used or subclassed directly. Instead use the
+        :class:`~.Sensor` class which is needed to achieve the functionality described in this
+        class's documentation.
+
     """
+    platform_system = Property(weakref.ref, default=None,
```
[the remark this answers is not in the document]
etfrogers-dstl: As in could it be a `property`? If so then I think so. There is no particular reason this needs to be set by a user or set in the `__init__`, so yes, it could be changed.

stonesoup/functions.py
```diff
@@ -566,3 +567,26 @@ def mod_elevation(x):
...
+    mapping: :class:`Union[List, np.ndarray, StateVector]`
+        A mapping that needs to be formatted to a valid mapping shape
+
+    Returns
+    -------
+    mapping: :class:`np.ndarray`
+        A flattened np.ndarray suitable for indexing a :class:`StateVector`
+    """
+
+    return np.asarray(mapping).flatten()
```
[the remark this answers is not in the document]
etfrogers-dstl: Yeah, that combines the two operations nicely (I didn't know it did `asarray()`). Will update.

stonesoup/platform/base.py
rjgreen-dstl: Looking at the waypoint navigation work it might be sensible to allow it to be set externally, for example if you want to take real world movement data and run it through a simulator

stonesoup/sensor/sensor.py
etfrogers-dstl: Because of a circular reference: This needs to know about the `FixedSensorPlatform` class, to build one as a default internal platform. All `SensorPlatforms` need to know about the basic `BaseSensor` functionality to know how to handle them. In order to resolve this I had to split the functionality into one class that can be imported by `platform.simple` and one that imports `platform.simple`. I tried first to make `BaseSensor` private (by calling it `_Sensor`) but that means you lose the documentation for it (unless you know a way of fixing that).

stonesoup/platform/base.py
etfrogers-dstl: I just had to look back at the code to find out why this is. The reason is that `orientation` is a `Property` in `FixedPlatform` but a `property` (note P/p) in `MovingPlatform`. I think overriding `Property`s with `property`s was problematic, but I can't remember for certain. It may just be that I thought it unintuitive. 

Moving platform with optional transition model could work, but there would be lots of checks for `self.transition_model is None` as they behave differently in e.g. velocity, orientation.

stonesoup/functions.py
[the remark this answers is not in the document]
etfrogers-dstl: I agree, but all I can say is it's what I tried first. Maybe that makes me the problem...

All I can say in my defence is: it just looks the right shape. Index with something the same shape as the thing you are indexing (as in logical indexing etc). In response to Rich's comment though I already modified the docs to make it clear that only muppets (me) would think this. 

Also added a note to the `StateVector` docs to advise against indexing with it.

stonesoup/models/measurement/base.py
```diff
@@ -1,18 +1,22 @@
...
     """Measurement Model base class"""
 
     ndim_state = Property(int, doc="Number of state dimensions")
-    mapping = Property(
-        sp.ndarray, doc="Mapping between measurement and state dims")
+    mapping = Property(np.ndarray, doc="Mapping between measurement and state dims")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.mapping = coerce_to_valid_mapping(self.mapping)
```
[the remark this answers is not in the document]
etfrogers-dstl: Point noted. I may try removing this and see what breaks (in my code)...

stonesoup/platform/base.py
etfrogers-dstl: The reason I limited it was the assumption that platforms with transition models should only move via the transition model. This may be faulty, and happy to change if sensible. 

You can always set the state_vector directly if required, but again very willing to change this if wanted.

stonesoup/models/measurement/base.py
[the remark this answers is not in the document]
etfrogers-dstl: Interesting. Do you expect `state_vector[mapping, 0]` and `state_vector[mapping, :]` to give different results?

stonesoup/platform/simple.py
```diff
@@ -43,116 +55,132 @@ def __init__(self, *args, **kwargs):
...
+            The sensor object to add
+        mounting_offset : :class:`StateVector`, optional
+            A StateVector with the mounting offset of the new sensor. If not supplied, defaults to
+            a zero vector
+        rotation_offset : :class:`StateVector`, optional
+            A StateVector with the rotation offset of the new sensor. If not supplied, defaults to
+            a zero vector.
+        """
+        self.sensors.append(sensor)
+        sensor.platform_system = weakref.ref(self)
```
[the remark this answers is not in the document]
etfrogers-dstl: Actioned

stonesoup/sensor/base.py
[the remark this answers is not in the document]
etfrogers-dstl: Actioned. I think you're right it makes more sense that way.

stonesoup/platform/base.py
```diff
@@ -1,37 +1,223 @@
...
+        """
+        return StateVector([0] * self.ndim)
+
+    @property
+    def is_moving(self) -> bool:
+        return False
+
+    def move(self, timestamp: datetime.datetime, **kwargs) -> None:
+        """For a fixed platform this method has no effect other than to update the timestamp."""
+        # TODO Is this a sensible implementation?
```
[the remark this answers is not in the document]
etfrogers-dstl: OK. No action needed then

stonesoup/models/measurement/base.py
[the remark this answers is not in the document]
sdhiscocks: `state_vector[mapping, 0]` is a `(n, )` and `state_vector[mapping, :]` is `(n, 1)`.

stonesoup/sensor/sensor.py
sdhiscocks: Circular references are a right pain 😩

stonesoup/models/measurement/nonlinear.py
etfrogers-dstl: I think we have decided to leave this as is.

stonesoup/models/measurement/base.py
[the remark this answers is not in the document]
etfrogers-dstl: I'm impressed (if you knew that off the top of your head with testing, that is)!

Yes, you are right. Anyway, I have removed this code now, so marking as resolved.

stonesoup/functions.py
etfrogers-dstl: I think I have accepted this point now, so will resolve this.

stonesoup/platform/simple.py
```diff
@@ -1,17 +1,22 @@
...
 from ..types.state import StateVector
-from ..sensor import Sensor
-from ..functions import cart2pol
-from .base import Platform
+from .base import Platform, MovingPlatform, FixedPlatform
 
 
-class SensorPlatform(Platform):
-    """A simple Platform that can carry a number of different sensors and is
+class SensorPlatformMixin(Platform, ABC):
```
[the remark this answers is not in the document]
etfrogers-dstl: Agreed after discussion with Rich. This simplifies things considerably, and will be pushed shortly.

stonesoup/platform/base.py
etfrogers-dstl: I think the compromise solution is to check for `transition_model is None` at this point, and raise an error only if the transition model is set. This is what I have implemented in a new version to be pushed shortly.

stonesoup/platform/base.py
```diff
@@ -1,69 +1,516 @@
...
+        """Return the ``True`` if the platform is moving, ``False`` otherwise.
+
+        Equivalent (for this class) to ``all(v == 0 for v in self.velocity)``
+        """
+        # Note: a platform without a transition model can be given a velocity as part of it's
+        # StateVector. It just won't move
+        # This inconsistency is handled in the move logic
+        return np.any(self.velocity != 0)
+
+    def _set_position(self, value: StateVector):
```
rjgreen-dstl, line 362: If pushing a platform state vector from real-world data there is a corresponding need for _set_velocity, or possibly just one _set_state which allows the state vector to be set externally if no transition model exists?

etfrogers-dstl: There is nothing stopping you setting `platform.state` or `platform.state.statevector` directly at the moment.
The leaves two questions:
1. Does this address your use case?
2. Is there any reason I should forbid either of those?

My guess is 1) yes, 2) no, but I'd like to know your thoughts.

stonesoup/sensor/sensor.py
```diff
@@ -0,0 +1,48 @@
...
+    """Sensor Base class for general use.
+
+    Most properties and methods are inherited from :class:`~.BaseSensor`, but this class includes
+    crucial functionality and so should be used in preference.
+
+    All sensors must be mounted on a platform to calculate their position and orientation. To
+    make this easier, if the sensor has a position and/or orientation specified in the constructor,
+    and no :attr:`platform_system`, then the default is to create an internally held "private"
+    platform for the Sensor. This restricts the later setting of the :attr:`platform_system` but
+    does allow the Sensor to control (and set) its own position and orientation.
```
sdhiscocks, line 19: I think these should be `Property`s (defaulting to `None`), as this can't be serialised into configuration file.

stonesoup/platform/base.py
```diff
@@ -1,69 +1,516 @@
...
+                                doc="A list of StateVectors containing the sensor rotation "
+                                    "offsets from the platform's primary axis (defined as the "
+                                    "direction of motion). Defaults to a zero vector with the "
+                                    "same length as the Platform's :attr:`position_mapping`")
+    mounting_offsets = Property(List[StateVector], default=None,
+                                doc="A list of StateVectors containing the sensor translation "
+                                    "offsets from the platform's reference point. Defaults to "
+                                    "a zero vector with the same length as the Platform's "
+                                    ":attr:`position_mapping`")
+    sensors = Property(List["BaseSensor"], doc="A list of N mounted sensors", default=[])
```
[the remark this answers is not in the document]
etfrogers-dstl: Good point. I knew this about functions. Hadn't realised it applied to `Property`s, though I suppose I should've. Will fix.

stonesoup/sensor/sensor.py
etfrogers-dstl: Which things should be `Property`s? Position and orientation? These are never saved simply used to initialise a platform, so it seems odd to make them `Property`s. Or maybe I misunderstood?

stonesoup/platform/base.py
```diff
@@ -1,69 +1,516 @@
...
+    state = Property(State, doc="The platform state at any given point. For a static platform, "
+                                "this would usually contain its position coordinates in the form"
+                                "``[x, y, z]``. For a moving platform it would contain position "
+                                "and velocity interleaved: ``[x, vx, y, vy, z, vz]``")
+    position_mapping = Property(Sequence[int],
+                                doc="Mapping between platform position and state vector. For a "
+                                    "position-only 3d platform this might be ``[0, 1, 2]``. For a "
+                                    "position and velocity platform: ``[0, 2, 4]``")
```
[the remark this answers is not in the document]
etfrogers-dstl: Fixed.

stonesoup/sensor/sensor.py
etfrogers-dstl: After our discussion this morning, I accept this is an issue, and have added a note to the docs and a warning to the serialisation code, but fixing this properly seems pretty painful. As it is a edge case, we are probably OK with a warning, but if I have a brainwave on how to do it, I will update the code in a later commit. Not promising anything!

stonesoup/platform/base.py
```diff
@@ -1,69 +1,516 @@
...
+        if not self.is_moving:
+            raise AttributeError('Orientation of a zero-velocity moving platform is not defined')
+        velocity = self.velocity
+
+        if self.ndim == 3:
+            _, bearing, elevation = cart2sphere(*velocity.flat)
+            return StateVector([0, bearing, elevation])
+        elif self.ndim == 2:
+            _, bearing = cart2pol(*velocity.flat)
+            return StateVector([0, bearing])
```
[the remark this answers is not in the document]
etfrogers-dstl: After discussion, the cost/benefit here is too great to make the change. 

The returned vector is much less useful if it has mixed objects in it, and `Bearing` and `Elevation`, although correct, are not necessary for the obvious use cases of an orientation vector.

stonesoup/sensor/sensor.py
etfrogers-dstl: OK. I think I fixed it up. The problem I was having before was trying to test it too generically, but the simple case is much easier and works now.

etfrogers-dstl:
Just adding a few more tests to improve coverage...

etfrogers-dstl:
> Just adding a few more tests to improve coverage...

All done!

Issue #164, opened 2020-02-12
Create Detection Simulator which uses Platforms/Sensors


sdhiscocks:
Create a simulator, which takes in ground truth of targets, and detection simulator which has one/multiple platforms with one/multiple sensors, producing detections.

sdhiscocks:
I believe with the sensors `measure` method, this would be fairly easy to implement, using current detection simulator as an initial guide.

PR #182, opened 2020-04-03
Added platform detection simulators.

oharrald-Dstl:
Previously, simulated detections were provided by measurement models, which did not result in an accurate representation of detection possibilities for each individual sensor/collection of sensors on a platform.

Resolves: #164

### DIFF
--- a/stonesoup/simulator/tests/conftest.py
+++ b/stonesoup/simulator/tests/conftest.py
@@ -1,12 +1,10 @@
 # -*- coding: utf-8 -*-
 
-import datetime
-
 import pytest
 import numpy as np
 
 from ...types.detection import Detection
-from ...types.state import State
+from ...sensor import Sensor
 
 
 @pytest.fixture()
@@ -40,7 +38,7 @@ def measurement_model():
 
 @pytest.fixture()
 def sensor_model1():
-    class TestSensor():
+    class TestSensor(Sensor):
 
         @staticmethod
         def measure(ground_truth):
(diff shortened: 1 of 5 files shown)


stonesoup/simulator/tests/conftest.py
```diff
@@ -30,3 +36,52 @@ def function(state, noise):
...
+
+        @staticmethod
+        def measure(ground_truth):
+            return Detection(ground_truth.state_vector,
+                             timestamp=ground_truth.timestamp)
+    return TestSensor()
+
+
+@pytest.fixture()
+def platform_class():
```
etfrogers-dstl, line 64: Is there a need to use a custom platform class rather than one of the existing ones? There may well be, but can you explain what value it adds?

oharrald-Dstl: The test platform and sensor classes make detections predictable and hence easier to test.

stonesoup/simulator/tests/conftest.py
```diff
@@ -30,3 +36,52 @@ def function(state, noise):
...
+                             timestamp=ground_truth.timestamp)
+    return TestSensor()
+
+
+@pytest.fixture()
+def sensor_model2():
+    class TestSensor:
+
+        @staticmethod
+        def measure(ground_truth):
```
etfrogers-dstl, line 57: The problem with this is that you are not actually testing your simulator against the platform/sensor classes that people will use. 

Implementing your own `measure()` function is fine, but I would subclass Sensor
```suggestion
class TestSensor(Sensor):
   def measure(self, ground_truth):  # Note not static as in Sensor
````

Suggestion for Platform below

File: stonesoup/simulator/tests/conftest.py [hunk 3]
```diff
@@ -30,3 +36,52 @@ def function(state, noise):
...
+            self.state.timestamp = timestamp
+
+        @property
+        def state_vector(self):
+            return self.state.state_vector
+
+        @property
+        def timestamp(self):
+            return self.state.timestamp
+    return TestPlatform
```
etfrogers-dstl, line 87: Suggest removing this

stonesoup/simulator/tests/test_platform.py
```diff
@@ -0,0 +1,97 @@
...
+import datetime
+
+import numpy as np
+
+from ...types.state import State
+
+from ..platform import PlatformDetectionSimulator
+from ..simple import SingleTargetGroundTruthSimulator
+
+
```
etfrogers-dstl, line 12: Suggest convenience method with similar functionality to your platform class.

```suggestion
def build_platform(sensors, x_velocity):

    state = State(state_vector=StateVector([[0], [x_velocity], [0], [0]]),
                  timestamp=datetime.datetime(2020, 4, 1))
    model_1d = ConstantVelocity(0.0)  # zero noise so pure movement
    trans_model = CombinedLinearGaussianTransitionModel([model_1d] * 2)
    mounting_offsets = np.zeros((len(sensors), 2))
    mounting_mappings = np.array([[0, 2]])
    platform = SensorPlatform(state=state, sensors=sensors, transition_model=trans_model,
                              mounting_offsets=mounting_offsets,
                              mounting_mappings=mounting_mappings)
    return platform

```

Then replace `platform_class` with `build_platform` throughout the below.

etfrogers-dstl:
Fixed it :)

Issue #186, opened 2020-04-06
StateVector indexing leads to unnecessarily complicated code


etfrogers-dstl:
Firstly, let me say that I think the StateVector type generally useful and the removes confusion (keeping vectors as column vectors is a good idea).

Having said that, in the code I have been writing recently, I have been find some 'gotchas' in the behaviour of StateVectors and I think it could be changed to improve things.

The main issue is that `my_state_vector[1]` returns a 1x1 Statevector rather than a number (`int`, `float` etc). If you have 
````python
my_list = [0, 1, 2]
my_array = np.array([0, 1, 2])
my_sv = StateVector([0, 1, 2])
````
then the following are true (and expected)
````python
my_list[1] == 1 and isinstance(my_list[1], int)
my_array[1] == 1 and isinstance(my_array[1], int)
````
However the following is NOT true
````python
my_sv[1] == 1 and isinstance(my_sv[1], int)
````
In fact (in psuedocode) `my_sv[1] == StateVector([[1]])` i.e. a 1x1 StateVector with the right value, but unexpected type. To work around this, my code (and a lot of code in the code base) has to use the pattern `my_sv[1, 0]`. If we do that then:
````python
my_sv[1, 0] == 1 and isinstance(my_sv[1, 0], int)
````
as expected.

Current behaviour also means that iterating through a StaveVector produces a sequence of StateVector rather than the expected sequence of numbers. 

There is a fairly simple fix to this, overriding the behaviour of `StateVector.__getitem__()` and I will submit a pull request shortly demonstrating this. The change does **not** break the existing pattern `m_sv[1, 0]`, but it does break a few tests that rely on iteration. However, I have fixed those in the pull request, and I would argue that the fixes reduce complexity and increase readability.

I acknowledge that this change may be personal preference and would appreciate other thoughts on this. To me, though, it seems a simple improvement, without much downside.

sdhiscocks:
I'd be hesitant to do this, as it's unknown what impact this might have when passing a StateVector to _numpy_, _scipy_ or other 3rd party libraries. It also have different behaviour if in some cases you have a mix of StateVector and standard column vectors in your code.

etfrogers-dstl:
Hmm, that's an interesting point about the interface with other libraries. I hadn't considered that. 

In the existing codebase, it works as expected, but I suppose maybe it could cause subtle bug elsewhere. I would argue that you "shouldn't" be using other column vectors, but not sure if we can assume it. 

[...]

PR #187, opened 2020-04-06
Change StateVector indexing 

etfrogers-dstl:
This pull request updates StateVector indexing and code that uses it to demonstrate the change suggested in issue #186. I think this is a simple and useful update, but if people feel this over-complicates the Statevector class for little benefit, then the PR can be shelved.

### DIFF
--- a/stonesoup/feeder/geo.py
+++ b/stonesoup/feeder/geo.py
@@ -91,8 +91,8 @@ class LongLatToUTMConverter(Feeder):
     """
 
     mapping = Property(
-        (float, float), default=(0, 1),
-        doc="Indexes of long, lat. Default (0, 1)")
+        (float, float), default=[0, 1],
+        doc="Indexes of long, lat. Default [0, 1]")
     zone_number = Property(
         int, default=None,
         doc="UTM zone number to carry out conversion. Default `None`, where it"
@@ -109,7 +109,7 @@ class LongLatToUTMConverter(Feeder):
             utm_detections = set()
             for detection in detections:
                 easting, northing, zone_num, northern = utm.from_latlon(
-                    *detection.state_vector[self.mapping[::-1], 0],
+                    *detection.state_vector[self.mapping[::-1]],
                     self.zone_number)
                 if self.zone_number is None:
                     self.zone_number = zone_num
--- a/stonesoup/feeder/geo.py
+++ b/stonesoup/feeder/geo.py
@@ -91,8 +91,8 @@ class LongLatToUTMConverter(Feeder):
     """
 
     mapping = Property(
-        [float, float], default=[0, 1],
-        doc="Indexes of long, lat. Default [0, 1]")
+        (float, float), default=(0, 1),
+        doc="Indexes of long, lat. Default (0, 1)")
     zone_number = Property(
         int, default=None,
         doc="UTM zone number to carry out conversion. Default `None`, where it"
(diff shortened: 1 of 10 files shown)


stonesoup/feeder/geo.py
```diff
@@ -91,8 +91,8 @@ class LongLatToUTMConverter(Feeder):
     """
 
     mapping = Property(
-        (float, float), default=(0, 1),
```
etfrogers-dstl, line 94: There is a subtlety here: The old default of mapping as a tuple means that simple indexing into a StateVector using the tuple of the form `my_statevector[self.mapping]` fails, as it interprets the tuple as indexes over successive dimensions. 

Using the (previously necessary) pattern of `my_statevector[self.mapping, 0]` makes numpy interpret the tuple as a series of indices into the first dimension. Switching from tuple to list restores the expected behaviour in both cases. 

In essence, this only used to work because of a side-effect of the odd indexing being used before.

jmbarr:
Doesn't this PR have much the same intent as #170? Would it be worth combining?

etfrogers-dstl:
> How does this work with types which aren't `int` but could usefully be returned directly (e.g. `float`, `long`, `angle`, `myOwnVerySpecialType`, etc)?

This works the same regardless of the type `dtype` of the `StateVector` (which I think is what you are asking). The check for `int` in `__getitem__` and `__setitem__` is whether the **index** supplied is an `int`: i.e. we did `my_state_vector[1]`. The check is necessary so that all other index modes (`my_state_vector[:]`, `my_state_vector[1, 0]`, `my_state_vector[[1,2]]` etc) behave as expected.

stonesoup/feeder/geo.py
sdhiscocks: I think these should be left as is, otherwise risk having mutable defaults.

stonesoup/models/measurement/nonlinear.py
```diff
@@ -244,23 +244,20 @@ def function(self, state, noise=False, **kwargs):
...
-        xyz = [[x], [y], [z]]
         inv_rotation_matrix = inv(self._rotation_matrix)
-        xyz_rot = inv_rotation_matrix @ xyz
-        xyz = [xyz_rot[0][0], xyz_rot[1][0], xyz_rot[2][0]]
-        x, y, z = xyz + self.translation_offset[:, 0]
+        xyz = inv_rotation_matrix @ xyz
 
         res = np.zeros((self.ndim_state, 1)).view(StateVector)
-        res[self.mapping, 0] = x, y, z
+        res[self.mapping] = xyz + self.translation_offset
```
sdhiscocks, line 260: This may break if mapping is tuple.

stonesoup/models/measurement/nonlinear.py
```diff
@@ -343,23 +340,22 @@ def ndim_meas(self):
...
         inv_rotation_matrix = inv(self._rotation_matrix)
-        xyz_rot = inv_rotation_matrix @ xyz
-        xy = [xyz_rot[0][0], xyz_rot[1][0]]
-        x, y = xy + self.translation_offset[:, 0]
+        xyz = inv_rotation_matrix @ xyz
+        xy = xyz[0:2]
 
         res = np.zeros((self.ndim_state, 1)).view(StateVector)
-        res[self.mapping, 0] = x, y
+        res[self.mapping] = xy + self.translation_offset
```
sdhiscocks, line 358: This may break if mapping is tuple.

etfrogers-dstl: Fixed

stonesoup/models/measurement/nonlinear.py
etfrogers-dstl: Fixed

stonesoup/feeder/geo.py
etfrogers-dstl: Good catch. Reverted.

I was in the mode of think tuples were not ideal for `mapping`s at that point. You have since convinced me...

PR #189, opened 2020-04-07
Added StatesLengthLimiter

sjablett-dstl:
By default Stone Soup stores the track history for all tracks in memory.  If
running Stone Soup on very large data your application may run out of memory
and the process terminated - often by the operating system.

This wrapper converts the states space list to a collections.deque data
type with the maximum length specified.

### DIFF
--- a/docs/source/stonesoup.initiator.rst
+++ b/docs/source/stonesoup.initiator.rst
@@ -10,4 +10,8 @@ Initiators
 .. automodule:: stonesoup.initiator.simple
     :show-inheritance:
 
+Wrappers
+--------
+.. automodule:: stonesoup.initiator.wrapper
+    :show-inheritance:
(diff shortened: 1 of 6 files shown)


sdhiscocks:
> Looks sensible in general. I wonder if you have considered the approach of using a Mixin rather than a wrapper like the below:
> 
One issue with Mixin, is that is doesn't work well with serialisation to create configuration files. For that reason I'd say having a wrapper is preferable.

sglvladi:
Looks good 👍

etfrogers-dstl:
> One issue with Mixin, is that is doesn't work well with serialisation to create configuration files. For that reason I'd say having a wrapper is preferable.

OK. Fine by me, just wanted to be sure it was considered.

PR #191, opened 2020-04-09
Update AESARadar to avoid mutating measurement model

etfrogers-dstl:
The AESA radar `measure` function was previously written so that saved a reference to a measurement model with each detection that is later mutated. This code change it to use a deepcopy instead.

### DIFF
--- a/stonesoup/sensor/radar/radar.py
+++ b/stonesoup/sensor/radar/radar.py
@@ -492,8 +492,8 @@ class AESARadar(Sensor):
         # Is the state detected?
         if np.random.rand() <= det_prob:
             measurement_model = copy.deepcopy(self.measurement_model)
-            measurement_model.translation_offset = self.translation_offset
-            measurement_model.rotation_offset = self.rotation_offset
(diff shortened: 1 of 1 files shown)


stonesoup/sensor/radar/radar.py
```diff
@@ -489,9 +491,10 @@ def measure(self, sky_state, noise=True, **kwargs):
         det_prob = self.gen_probability(sky_state)[0]
         # Is the state detected?
         if np.random.rand() <= det_prob:
-            self.measurement_model.translation_offset = self.translation_offset
-            self.measurement_model.rotation_offset = self.rotation_offset
+            measurement_model = copy.deepcopy(self.measurement_model)
+            measurement_model.translation_offset = self.translation_offset
+            measurement_model.rotation_offset = self.rotation_offset
```
rjgreen-dstl, line 496: Will you not have the same issue here with the two offsets being pointers to properties which will evolve over time?

etfrogers-dstl: A good point...

PR #192, opened 2020-04-15
corrected rotx roty rotz for n x theta array. Now returns 3 x 3 x n a…

dcherrie-dstl:
Corrected rotx roty rotz for n x theta array. Now returns 3 x 3 x n array of rotation matrices. Scalar theta input and output is unchanged.

### DIFF
--- a/stonesoup/functions.py
+++ b/stonesoup/functions.py
@@ -406,10 +406,11 @@ def rotx(theta):
     """
 
     c, s = np.cos(theta), np.sin(theta)
-
-    return np.array([[1, 0, 0],
-                     [0, c, -s],
-                     [0, s, c]])
+    zero = np.zeros_like(theta) 
+    one = np.ones_like(theta)
+    return np.array([[one, zero, zero],
+                     [zero, c, -s],
+                     [zero, s, c]])
 
 
 def roty(theta):
@@ -441,10 +442,11 @@ def roty(theta):
     """
 
     c, s = np.cos(theta), np.sin(theta)
-
-    return np.array([[c, 0, s],
-                     [0, 1, 0],
-                     [-s, 0, c]])
+    zero = np.zeros_like(theta) 
+    one = np.ones_like(theta)
+    return np.array([[c, zero, s],
+                     [zero, one, zero],
+                     [-s, zero, c]])
 
 
 def rotz(theta):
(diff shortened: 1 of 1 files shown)


dcherrie-dstl:
> Code looks good.
> 
> I think the docs might need updating in all three. Perhaps something like.
> 
> ```
> ...
> theta: Union[float, np.ndarray]
>         Rotation angle specified as a real-valued number or an :class:`np.ndarray` of reals. ...
> ...
> ```
> 
> I think you should also explain in the docs what shape the output is for a given input array (if I pass in nxm, do I get nxmx3x3?)

Alternatively, how about a more subtle doc correction below? Prior to this PL the rotation functions would still accept an input array of thetas returning a counter intuitive output shape of (3 x 3). I'd consider this PL a correction to expected behaviour for array inputs.

    Returns
    -------
    : :class:`numpy.ndarray` of shape (3, 3) or (3, 3, n) for `n x theta` input
        Rotation matrix around x-axis of the form :eq:`Rx`.

etfrogers-dstl:
> Alternatively, how about a more subtle doc correction below? Prior to this PR the rotation functions would still accept an input array of thetas returning a counter intuitive output shape of (3 x 3). I'd consider this PR a correction to expected behaviour for array inputs.
> 
> ```
> Returns
> -------
> : :class:`numpy.ndarray` of shape (3, 3) or (3, 3, n) for `n x theta` input
>     Rotation matrix around x-axis of the form :eq:`Rx`.
> ```

That's OK for the outputs. (Though I don't think the input is `n x theta`: isn't it an length `n` array containing theta values).

I still think that the statement `Rotation angle specified as a real-valued number` for the input docs is misleading if we also accept an array.

dcherrie-dstl:
> > Alternatively, how about a more subtle doc correction below? Prior to this PR the rotation functions would still accept an input array of thetas returning a counter intuitive output shape of (3 x 3). I'd consider this PR a correction to expected behaviour for array inputs.
> > ```
> > Returns
> > -------
> > : :class:`numpy.ndarray` of shape (3, 3) or (3, 3, n) for `n x theta` input
> >     Rotation matrix around x-axis of the form :eq:`Rx`.
> > ```
> 
> That's OK for the outputs. (Though I don't think the input is `n x theta`: isn't it an length `n` array containing theta values).
> 
> I still think that the statement `Rotation angle specified as a real-valued number` for the input docs is misleading if we also accept an array.

Thanks very much for the suggested docstring corrections, the last commit should reflect these.

PR #194, opened 2020-04-17
Add support for array/number types and parser type in YAML

sdhiscocks:
- Add support for Array, Angle and Numeric types in YAML serialiser
- Allow setting of YAML parser type, so faster C library can be used for parsing


sdhiscocks:
Having some trouble with Python 3.5, which I believe is to do with dictionary ordering. Will have to set up Python 3.5 environment to debug properly.

PR #200, opened 2020-04-20
Added multi-deleter methods

oharrald-Dstl:
The methods apply a list of deleters to input tracks

### DIFF
--- a/docs/source/stonesoup.deleter.rst
+++ b/docs/source/stonesoup.deleter.rst
@@ -18,3 +18,9 @@ Time Based
 .. automodule:: stonesoup.deleter.time
     :show-inheritance:
     :inherited-members:
+
+Multi
+-----
+.. automodule:: stonesoup.deleter.multi
+    :show-inheritance:
+    :inherited-members:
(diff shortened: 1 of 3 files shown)


docs/source/stonesoup.deleter.rst
```diff
@@ -18,3 +18,9 @@ Time Based
 .. automodule:: stonesoup.deleter.time
     :show-inheritance:
     :inherited-members:
+
+Multi
```
rjgreen-dstl, line 22: Suggest Composite, or multiple method, something more descriptive so that the reader understands immediately

stonesoup/deleter/multi.py
```diff
@@ -0,0 +1,73 @@
...
+"""Contains the multi-deleter feeders"""
+
+from ..base import Property
+from .base import Deleter
+
+
+class MultiDeleterIntersect(Deleter):
+    """ Track deleter composed of multiple deleters.
+
+    Deletes tracks if they satisfy the conditions of all deleters applied.
```
rjgreen-dstl, line 11: I suspect that logically there should be an option for the intersection or union of all deleter's to make more generic. Could be a logical input along lines of default intersection = True

rjgreen-dstl: @rjgreen-dstl should read to the end of the file before writing comments

oharrald-Dstl: I suppose there could be a second argument to a more generic "CompositeDeleter" class which determines whether it intersects or unifies?

PR #201, opened 2020-04-20
external sgp4 propagator into transition model

dcherrie-dstl:
Provided initial implementation of sgp4 transition model + test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,8 @@ jobs:
             python3 -m venv venv
             . venv/bin/activate
             pip install -e .[dev]
+            pip install astropy
+            pip install sgp4
       - save_cache:
           paths:
             - ./venv
--- a/stonesoup/models/transition/orbital/orbit.py
+++ b/stonesoup/models/transition/orbital/orbit.py
@@ -5,8 +5,7 @@ from numpy import matlib
 from scipy.stats import norm
 from scipy.stats import multivariate_normal
 from datetime import datetime, timedelta
-from ....orbital_functions import stumpf_c, stumpf_s, universal_anomaly_newton, \
-    lagrange_coefficients_from_universal_anomaly
+from ....orbital_functions import lagrange_coefficients_from_universal_anomaly
 
 from ....base import Property
 from ....types.orbitalstate import OrbitalState, TLEOrbitalState
@@ -15,6 +14,7 @@ from ...base import LinearModel, NonLinearModel
 from .base import OrbitalTransitionModel
 
 from sgp4.api import jday, Satrec
+
 
 class SimpleMeanMotionTransitionModel(OrbitalTransitionModel, LinearModel):
     r"""This simple transition model uses the mean motion to update the
--- a/stonesoup/models/transition/orbital/orbit.py
+++ b/stonesoup/models/transition/orbital/orbit.py
@@ -455,6 +455,7 @@ class CartesianTransitionModel(OrbitalTransitionModel, NonLinearModel):
                                            time_interval=time_interval))
         # ravel appears to be necessary here.
 
+
 class SGP4TransitionModel(OrbitalTransitionModel, NonLinearModel):
     """This class wraps https://pypi.org/project/sgp4/
 
@@ -467,12 +468,6 @@ class SGP4TransitionModel(OrbitalTransitionModel, NonLinearModel):
         CovarianceMatrix, default=CovarianceMatrix(np.zeros((6, 6))),
         doc=r"Transition noise covariance :math:`\Sigma` per unit time "
             r"interval (assumed seconds)")
-
-    _uanom_precision = Property(
-        float, default=1e-8, doc="The stopping point in the progression of "
-                                 "ever smaller f/f' ratio in the the Newton's "
-                                 "method calculation of universal anomaly."
-    )
 
     @property
     def ndim_state(self):
@@ -532,7 +527,7 @@ class SGP4TransitionModel(OrbitalTransitionModel, NonLinearModel):
         return self.transition(orbital_state, time_interval)
 
     def transition(self, orbital_state, time_interval=timedelta(seconds=0)):
-        r"""The transition proceeds as algorithm 3.4 in [1]
+        r"""
 
         Parameters
         ----------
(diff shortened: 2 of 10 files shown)


stonesoup/models/transition/orbital/tests/test_orbit.py
```diff
@@ -0,0 +1,49 @@
+from ......types.groundtruth import GroundTruthPath
+from ......types.orbitalstate import TLEOrbitalState
+from ..orbit import SGP4TransitionModel
+
+# https://pypi.org/project/sgp4/
```
jmbarr, line 5: This means that we need to go and install the sgp4 package I assume. There is a way of having this flagged as a dependency such that the flake8 tests work automatically.

dcherrie-dstl: Yes the external sgp4 propagator library is required. Could please advise on how to add such a flag?

stonesoup/models/transition/orbital/tests/test_orbit.py
```diff
@@ -0,0 +1,49 @@
...
+    state_vector = np.zeros(shape=(6,1))
+    # A state vector MUST be passed; however it will be ignored
+    state_vector[0,0] = tle_ext.inclo
+    state_vector[1,0] = tle_ext.nodeo
+    state_vector[2,0] = tle_ext.ecco
+    state_vector[3,0] = tle_ext.argpo
+    state_vector[4,0] = tle_ext.no
+    state_vector[5,0] = tle_ext.no_kozai
+
+    test_tle  = GroundTruthPath(TLEOrbitalState(state_vector,metadata = tle_dict))
```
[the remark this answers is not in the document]
dcherrie-dstl: I our discussions you mentioned I should use readers to input TLE states. There are three Stone Soup readers, Detection, Truth and Sensor. I require Truth, hence I'm testing the transition model  
by generating the output of a feeder class.

stonesoup/models/transition/orbital/orbit.py
```diff
@@ -370,6 +371,205 @@ def transition(self, orbital_state, time_interval=timedelta(seconds=0)):
...
+            The orbital state vector returned by the transition function
+
+        Warning
+        -------
+        If noisy samples from the transition function are required, use
+        the :attr:`rvs` method.
+        """
+        # Get the position and velocity vectors
+        # Evaluated at initial timestamp
+        tle_ext = Satrec.twoline2rv(orbital_state.state.metadata['line_1'],
```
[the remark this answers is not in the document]
dcherrie-dstl: This works because I have input a ground truth state. I'm unsure if the transition model should or should not be able to work with GroundTruthState object. I could always extract the Orbital State prior to entering the transition function.

stonesoup/models/transition/orbital/tests/test_orbit.py
```diff
@@ -0,0 +1,49 @@
...
+    # a: Evaluate position at initial date using external SGP4
+    
+    jd, fr = jday(initial_date.year, initial_date.month, initial_date.day, 
+                  initial_date.hour, initial_date.minute, initial_date.second) 
+    e, r, v = tle_ext.sgp4(jd, fr)
+	
+	# b: Evaluate position at initial date using wrapped SGP4 library
+    
+    state_vector = np.zeros(shape=(6,1))
+    # A state vector MUST be passed; however it will be ignored
```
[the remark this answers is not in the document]
dcherrie-dstl: I have tried to follow your advice and work around the fact StateVector objects have to be initialised with cartesian elements. Right now the metadata does not get updated, but this meant to be an example of wrapping not a fully worked solution.

stonesoup/models/transition/orbital/tests/test_orbit.py
jmbarr: @sdhiscocks, can you advise? I think required packages have to be added to a list somewhere.

stonesoup/models/transition/orbital/orbit.py
```diff
@@ -370,6 +371,205 @@ def transition(self, orbital_state, time_interval=timedelta(seconds=0)):
...
+                                    orbital_state.state.metadata['line_2'])
+        jd, fr = jday(orbital_state.state.timestamp.year,
+                      orbital_state.state.timestamp.month,
+                      orbital_state.state.timestamp.day,
+                      orbital_state.state.timestamp.hour,
+                      orbital_state.state.timestamp.minute,
+                      orbital_state.state.timestamp.second)
+        e, bold_r_0, bold_v_0 = tle_ext.sgp4(jd, fr)
```
jmbarr, line 565: These values never used...?

stonesoup/models/transition/orbital/orbit.py
```diff
@@ -370,6 +371,205 @@ def transition(self, orbital_state, time_interval=timedelta(seconds=0)):
...
+
+        tt = orbital_state.state.timestamp + time_interval
+        jd, fr = jday(tt.year,tt.month,tt.day,
+                      tt.hour,tt.minute,tt.second)
+        e, bold_r, bold_v = tle_ext.sgp4(jd, fr)
```
jmbarr, line 570: Warning here. These values are in km and km/s. OrbitalState does not require specification of particular units and so will break if used with (e.g.) a method which is initialised with other units. 

At least need to include a comment. Better would be some sort of checking...

jmbarr: Further, default initialisation of OrbitalState assumes units in m.

jmbarr:
Let us know what you think

jmbarr:
CircleCI errors are to do with the fact that astropy and sgp4 don't yet exist in the test schedule.

stonesoup/models/transition/orbital/tests/test_orbit.py
sdhiscocks: You can add dependencies to `setup.py`. Suggest you add a _orbital_ dependency set (you'll see examples for _dev_ and _video_ on _master_).

If you want to run them on CircleCI you'll also need to update `.circleci/config.yml` file. You'll see _dev_ there already, so can add more by delimiting them with comma e.g.:
```
pip install -e .[dev,orbital]
```

stonesoup/models/transition/orbital/orbit.py
dcherrie-dstl: Good spot. I should probably add another test to help detect incorrect units.

stonesoup/models/transition/orbital/tests/test_orbit.py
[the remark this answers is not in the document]
dcherrie-dstl: > What happens when the state vector is updated by the orbitalstate.transition() function? Does the metadata get updated?

Technically with my implementation the state does not get updated, simply evaluated at different times. Really I just wanted to use this transition implementation for the Truth Simulator.

.circleci/config.yml
```diff
@@ -24,6 +24,8 @@ jobs:
             python3 -m venv venv
             . venv/bin/activate
             pip install -e .[dev]
+            pip install astropy
+            pip install sgp4
```
jmbarr, line 28: That seems to work

stonesoup/models/transition/orbital/tests/test_orbit.py
jmbarr: In fact I modified .circleci/config.yml. Was that bad?

.circleci/config.yml
jmbarr: Or maybe you made your changes at the same time...

stonesoup/models/transition/orbital/tests/test_orbit.py
sdhiscocks: It'd be better to change them as per my suggestion, as then it'll be easier if/when you add more dependencies, you just to it in _setup.py_ and CircleCI. Will also need astropy in setup.py, such when others pip install with _orbital_ extras, it'll pick the dependencies.

Issue #185, opened 2020-04-04
SimplePDA implementation is overly complicated


sglvladi:
It seems like the ``associate_highest_probability_hypotheses()`` method called is over-complicated and possibly unnecessary. Instead the class should simply return a copy the hypotheses returned by ``PDAHypothesiser``, but with normalised weights for each track.

PR #202, opened 2020-04-21
Simplify SimplePDA

sglvladi:
This PR addresses #185 and removes unnecessary processing from ``SimplePDA``. Also, ``SimplePDA`` is renamed to ``PDA``.

### DIFF
--- a/stonesoup/dataassociator/probability.py
+++ b/stonesoup/dataassociator/probability.py
@@ -12,17 +12,11 @@ from ..types.numeric import Probability
 import itertools
 
 
-class SimplePDA(DataAssociator):
-    """Simple Probabilistic Data Association (PDA)
+class PDA(DataAssociator):
+    """Probabilistic Data Association (PDA)
 
-    Given a set of detections and a set of tracks, each detection has a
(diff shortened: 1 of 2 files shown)


stonesoup/dataassociator/probability.py
```diff
@@ -52,14 +46,18 @@ def associate(self, tracks, detections, time):
             track: self.hypothesiser.hypothesise(track, detections, time)
             for track in tracks}
 
-        return associate_highest_probability_hypotheses(tracks, hypotheses)
+        # Ensure association probabilities are normalised
+        for track, hypothesis in hypotheses.items():
+            hypothesis.normalise_probabilities(total_weight=1)
+
```
sglvladi, line 52: This normalisation is not currently necessary, given that multi-hypotheses are already normalised on instantiation in PDAHypothesiser and (J)PDA requires the use of PDAHypothesiser. However, PDA should always produce normalised weights, whereas normalising weights is not technically necessary inside PDAHypothesiser. Therefore, I have included this normalisation to make sure normalisation is satisfied, in case users define some custom hypothesiser that does not normalise.

PR #205, opened 2020-04-27
Fix issue with logpdf not handling custom types (e.g. Angle)

sdhiscocks:


stonesoup/models/measurement/tests/test_models.py
```diff
@@ -226,10 +227,8 @@ def fun(x):
...
     assert approx(prob) == multivariate_normal.pdf(
-        meas_pred_wo_noise.T,
-        mean=np.array(h(state_vec,
-                        model.translation_offset,
-                        model.rotation_offset)).ravel(),
+        (meas_pred_wo_noise
+         - np.array(h(state_vec, model.translation_offset, model.rotation_offset))).ravel(),
         cov=R)
 
     # Propagate a state vector through the model
```
jmbarr, line 234: through

stonesoup/models/base.py
```diff
@@ -224,9 +224,10 @@ def pdf(self, state1, state2, **kwargs):
             The likelihood of ``state1``, given ``state2``
         """
 
+        # Calculate difference before to handle custom types (mean defaults to zero)
+        # This is required as log pdf coverts arrays to floats
         likelihood = multivariate_normal.logpdf(
```
jmbarr, line 229: I didn't realise this did this in log space by default (indeed, always). This is likely well justified but we should ensure that the users are aware that this is happening by way of a note in the documentation. 

Also, include a switch to allow the the calculation to proceed without logs?

sdhiscocks: The output is wrapped in `Probability` type, so transparently acts as float.

jmbarr: What I meant was that there's a potential choice here between `.logpdf()` and `.pdf()` and whereas `.logpdf()` might be a better, indeed obvious, choice, it would be worth documenting that this is what's happening, and perhaps giving users the option of using `.pdf()`.

sdhiscocks: I'm not clear what the advantage would be, as both using `logpdf` wrapped in `Probability` type is effectively same as `pdf` as a float, but using pure float risks issues with values close to zero.

sdhiscocks: I realise that I wasn't very clear about the behaviour. `pdf` is just a wrapper around `logpdf` that calls `exp(log_prob)`. `Probability` type is just a wrapper around a float that is `log_prob`. You use it like a non-logged probability, but if at some point you try to do an operation that doesn't work with a log probability, it'll call `exp(log_prob)`. This gives you advantages of both without having to worry about it.

stonesoup/models/measurement/tests/test_models.py
```diff
@@ -243,10 +242,8 @@ def fun(x):
...
     assert approx(prob) == multivariate_normal.pdf(
-        meas_pred_w_inoise.T,
-        mean=np.array(h(state_vec,
-                        model.translation_offset,
-                        model.rotation_offset)).ravel(),
+        (meas_pred_w_inoise
+         - np.array(h(state_vec, model.translation_offset, model.rotation_offset))).ravel(),
         cov=R)
 
     # Propagate a state vector throught the model
```
DaveKirkland, line 249: throught -> through

PR #212, opened 2020-05-04
Change Property to use type hints and drop Python 3.5 support

sdhiscocks:
This overall doesn't change behaviour of the code or how Properties on classes are defined, but enabled type hints to be used instead. At the moment, this change is backwards compatible.

This is useful to aid IDE's, using more familiar syntax for typing, and possibly allow future use for type checking.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,6 @@ workflows:
   version: 2
   test:
     jobs:
-      - test-35
       - test-36
       - test-37
       - test-38
@@ -11,9 +10,9 @@ workflows:
     jobs:
       - docs
 jobs:
-  test-35: &test-template
+  test-36: &test-template
     docker:
-      - image: circleci/python:3.5
+      - image: circleci/python:3.6
     steps:
       - checkout
       - restore_cache:
@@ -45,10 +44,6 @@ jobs:
               -f test-reports/coverage.xml \
               -F unittests \
               -n ${CIRCLE_BUILD_NUM}
-  test-36:
-    <<: *test-template
-    docker:
-      - image: circleci/python:3.6
   test-37:
     <<: *test-template
     docker:
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Stone Soup uses the following dependencies:
 
 | Name | License |
 | ---- | ------- |
-| [Python](https://www.python.org/) (v3.5+) | PSFL |
+| [Python](https://www.python.org/) (v3.6+) | PSFL |
 | [SciPy](https://www.scipy.org/) | BSD |
 | [matplotlib](https://matplotlib.org/) | [PSF/BSD-compatible](https://matplotlib.org/users/license.html) |
 | [ruamel.yaml](https://yaml.readthedocs.io/) | MIT |
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(name='stonesoup',
           'Topic :: Scientific/Engineering',
       ],
       packages=find_packages(exclude=('docs', '*.tests')),
+      python_requires='>=3.6',
       install_requires=[
           'ruamel.yaml>=0.15.45', 'numpy>=1.17', 'scipy', 'matplotlib', 'utm', 'pymap3d'],
       extras_require={
--- a/stonesoup/base.py
+++ b/stonesoup/base.py
@@ -222,7 +222,7 @@ class BaseMeta(ABCMeta):
                 cls._properties.move_to_end(name)
         for name, prop in cls._properties.items():
             if not (isinstance(prop.cls, type) or
-                    str(prop.cls).startswith('typing.')):
+                    (hasattr(prop.cls, '__module__') and prop.cls.__module__ == 'typing')):
                 raise ValueError(f'Invalid type specification ({str(prop.cls)}) '
                                  f'for property {name} of class {cls.__name__}')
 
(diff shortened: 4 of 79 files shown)


sdhiscocks:
This requires minimum of Python 3.6. As Python 3.5 is supported until September 2020, we should ideally still support it for now.

sdhiscocks:
> Looks neat. My only worry would be that this will add more overheads to `BaseMeta`, which already seems to be somewhat costly .

As it's only in the meta class, this just affects during class definition, not when creating instances of classes.

etfrogers-dstl (approved):
Looks good to me. Not sure if we can merge until the 13th?

Please tell you me you managed to automate most of the changes... 😊

etfrogers-dstl:
Does this (or can it) disallow types of the form `[int]`, which is't actually a valid type (it is a list of types).  It should be `List[int]` (or `Sequence[int]` etc).

When you check for types (e.g. when displaying `Property`s in a GUI, the `[int]` form doesn't really work (you can work around it, but it's very hacky, and easier to go back to the source and fix it...)

sdhiscocks:
> Does this (or can it) disallow types of the form `[int]`, which is't actually a valid type (it is a list of types). It should be `List[int]` (or `Sequence[int]` etc).
> 
> When you check for types (e.g. when displaying `Property`s in a GUI, the `[int]` form doesn't really work (you can work around it, but it's very hacky, and easier to go back to the source and fix it...)

So I checked, and it works with bad form as well unfortunately. I thought Python would raise an error as invalid type specification. Hopefully this change has got everything as the right types now, and we'll review types more closely going forward.

etfrogers-dstl:
Should we remove the check `if sys.version_info <= (3, 6)` on line 225 of `base.py` in this PR?

sdhiscocks:
> Should we remove the check `if sys.version_info <= (3, 6)` on line 225 of `base.py` in this PR?

Yeah. In fact there are a few version checks in other places. I'll remove those as well.

etfrogers-dstl:
> > Does this (or can it) disallow types of the form `[int]`, which is't actually a valid type (it is a list of types). It should be `List[int]` (or `Sequence[int]` etc).
> > When you check for types (e.g. when displaying `Property`s in a GUI, the `[int]` form doesn't really work (you can work around it, but it's very hacky, and easier to go back to the source and fix it...)
> 
> So I checked, and it works with bad form as well unfortunately. I thought Python would raise an error as invalid type specification. Hopefully this change has got everything as the right types now, and we'll review types more closely going forward.

How about this: https://github.com/dstl/Stone-Soup/tree/type_hint_checking

This adds a few lines: 
```
+        for name, prop in cls._properties.items():
+            if not (isinstance(prop.cls, type) or
+                    str(prop.cls).startswith('typing.')):
+                raise ValueError(f'Invalid type specification ({str(prop.cls)}) '
+                                 f'for property {name} of class {cls.__name__}')
```

And also fixes the one case left in the codebase of an invalid type (`value: any = Property(doc='Value of the metric')` should be `value: typing.Any = Property(doc='Value of the metric')`. Good to see that it caught it though. If the idea is of use, I would add a test or two.

P.S. it uses f-strings, so needs to be added to this PR (if you like it), rather than making a new one

sdhiscocks:
> > > Does this (or can it) disallow types of the form `[int]`, which is't actually a valid type (it is a list of types). It should be `List[int]` (or `Sequence[int]` etc).
> > > When you check for types (e.g. when displaying `Property`s in a GUI, the `[int]` form doesn't really work (you can work around it, but it's very hacky, and easier to go back to the source and fix it...)
> > 
> > 
> > So I checked, and it works with bad form as well unfortunately. I thought Python would raise an error as invalid type specification. Hopefully this change has got everything as the right types now, and we'll review types more closely going forward.
> 
> How about this: https://github.com/dstl/Stone-Soup/tree/type_hint_checking
> 
> This adds a few lines:
> 
> ```
> +        for name, prop in cls._properties.items():
> +            if not (isinstance(prop.cls, type) or
> +                    str(prop.cls).startswith('typing.')):
> +                raise ValueError(f'Invalid type specification ({str(prop.cls)}) '
> +                                 f'for property {name} of class {cls.__name__}')
> ```
> 
> And also fixes the one case left in the codebase of an invalid type (`value: any = Property(doc='Value of the metric')` should be `value: typing.Any = Property(doc='Value of the metric')`. Good to see that it caught it though. If the idea is of use, I would add a test or two.
> 
> P.S. it uses f-strings, so needs to be added to this PR (if you like it), rather than making a new one

Great. May as well get it in this PR as well if you wanna go ahead and fast-forward merge your branch onto this one.

etfrogers-dstl:
Another thought: the current logic has the argument winning if both a type hint and argument are specified. 

Is that the right behaviour? Should we warn/raise an error/ignore/something else?

etfrogers-dstl:
> Another thought: the current logic has the argument winning if both a type hint and argument are specified.
> 
> Is that the right behaviour? Should we warn/raise an error/ignore/something else?

Oh, and what if the type is NOT specified either way?

etfrogers-dstl:
I also added checks that raise errors in both the following cases:
``` python
i: float = Property(int, doc='Test')  # type specified twice
i = Property(doc='Test')  # type not specified
```
Is this the behaviour we want? 

I can't think of a use case where either of the above would be useful. If you really don't want to specify the type, you can use `typing.Any`.

oharrald-Dstl (approved):

PR #213, opened 2020-05-04
Moved sqrt logic into the KalmanUpdater controlled by a flag. Wrote t…

jmbarr:
…est case suggesting that sqrt version indeed more stable than naive version.

### DIFF
--- a/stonesoup/updater/kalman.py
+++ b/stonesoup/updater/kalman.py
@@ -151,7 +151,8 @@ class KalmanUpdater(Updater):
                                              predicted_state.timestamp,
                                              cross_covar=meas_cross_cov)
 
-    def update(self, hypothesis, force_symmetric_covariance=False, **kwargs):
+    def update(self, hypothesis, force_symmetric_covariance=False,
+               sqrt_form=False, **kwargs):
         r"""The Kalman update method. Given a hypothesised association between
         a predicted state or predicted measurement and an actual measurement,
         calculate the posterior state.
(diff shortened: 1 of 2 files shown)


stonesoup/updater/kalman.py
```diff
@@ -151,7 +151,8 @@ def predict_measurement(self, predicted_state, measurement_model=None,
                                              predicted_state.timestamp,
                                              cross_covar=meas_cross_cov)
 
-    def update(self, hypothesis, force_symmetric_covariance=False, **kwargs):
+    def update(self, hypothesis, force_symmetric_covariance=False,
+               sqrt_form=False, **kwargs):
```
sdhiscocks, line 155: There isn't a simple way to exploit these additional arguments in components.

This could be done by having a subclass that just has default `sqrt_form=True` and call `super()`. Or have `sqrt_form` as a `Property` of the Kalman class.

jmbarr: Not sure I follow first sentence. Example of how these arguments can't be exploited?

But then adding sqrt_form = Property(bool, default=False, ...) is fairly trivial, so happy if that's the preferred option.

sglvladi: I think I share @sdhiscocks's view on this one. When the updater is used inside a tracker, say [``SingleTargetTracker``](https://github.com/dstl/Stone-Soup/blob/7f6b95c0eb0642fe32628ac4bb404b238440e559/stonesoup/tracker/simple.py#L65), currently no kwargs are passed in. I suppose we could add kwargs support to ``tracks_gen()``, which would enable one to selectively run it with ``sqrt_form=True`` at chosen timesteps. The question is whether this is a valid use case and how that would work with configuration files . 

Although I'm I am conscious that ``force_symmetric_covariance`` is not new, another question is whether ``force_symmetric_covariance`` should also be defined as a ``Property`` for the same reasons.

jmbarr: My expertise in these matters is waning somewhat but I would have thought that variables which are only used in a single method (e.g. `sqrt_form` or `force_symmetric_covariance` are only used in `.update()`) are most sensibly (perhaps only for reasons of compactness) passed as keywords/arguments. Whereas stuff that's required across multiple methods can only work as a `Property`.

I note that the UKF doesn't do this...

PR #215, opened 2020-05-06
Fix divide by zero in SIAP Metrics when no truth/tracks

### LINKED ISSUE #209: self.metric_manager.generate_metrics() : ZeroDivisionError: division by zero
Probably my data, however FYI.

```
  File "lib\_mot\mot.py", line 302, in run
    self.metrics_generate()
  File "lib\_mot\mot.py", line 253, in metrics_generate
    self.metrics = self.metric_manager.generate_metrics()
  File "C:\g\testapp\lib\lib\site-packages\stonesoup\metricgenerator\manager.py", line 87, in generate_metrics
    metric = generator.compute_metric(self)
  File "C:\g\testapp\lib\lib\site-packages\stonesoup\metricgenerator\tracktotruthmetrics.py", line 45, in compute_metric
    self.A_time_range(manager),
  File "C:\g\testapp\lib\lib\site-packages\stonesoup\metricgenerator\tracktotruthmetrics.py", line 169, in A_time_range
    manager, timestamps)
ZeroDivisionError: division by zero
```

sdhiscocks:
Fixes #209

### DIFF
--- a/stonesoup/metricgenerator/tracktotruthmetrics.py
+++ b/stonesoup/metricgenerator/tracktotruthmetrics.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import datetime
+import warnings
 from operator import attrgetter
 
 import numpy as np
@@ -47,6 +48,16 @@ class SIAPMetrics(MetricGenerator):
                    self.LT(manager),
                    self.LS(manager)]
         return metrics
+
+    @staticmethod
+    def _warn_no_truth(manager):
+        if len(manager.groundtruth_paths) == 0:
+            warnings.warn("No truth to generate SIAP Metric", stacklevel=2)
+
+    @staticmethod
+    def _warn_no_tracks(manager):
+        if len(manager.tracks) == 0:
+            warnings.warn("No tracks to generate SIAP Metric", stacklevel=2)
 
     def C_single_time(self, manager, timestamp):
         r"""SIAP metric C at a specific time
@@ -130,7 +141,7 @@ class SIAPMetrics(MetricGenerator):
             C = self._jt_sum(manager, timestamps) / self._j_sum(
                 manager, timestamps)
         except ZeroDivisionError:
-            # No truth
+            self._warn_no_truth(manager)
             C = 0
         return TimeRangeMetric(
             title="SIAP C",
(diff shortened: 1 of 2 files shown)


stonesoup/metricgenerator/tracktotruthmetrics.py
```diff
@@ -207,7 +214,11 @@ def S_time_range(self, manager):
             self._n_t(manager, timestamp) - self._na_t(manager, timestamp)
             for timestamp in timestamps)
 
-        S = numerator / self._n_sum(manager, timestamps)
+        try:
+            S = numerator / self._n_sum(manager, timestamps)
+        except ZeroDivisionError:
+            # No tracks
+            S = 0
```
rjgreen-dstl, line 221: Its not as clear cut as just returning Zero when the division fails. S is a measure of percentage of spurious tracks, returning 0 indicates perfect performance where every true track is associated with an object which is being tracked. In this case it should probably return an error message to say that there were no tracks

stonesoup/metricgenerator/tracktotruthmetrics.py
```diff
@@ -277,9 +288,14 @@ def LS(self, manager):
...
         timestamps = manager.list_timestamps()
+        try:
+            LS = numerator / denominator
+        except ZeroDivisionError:
+            # No truth
+            LS = 0
         return TimeRangeMetric(
             title="SIAP LS",
-            value=numerator / denominator,
+            value=LS,
```
rjgreen-dstl, line 298: Returning Zero here is more reasonable (0 = worst performance), however the metric demoninator is is the sum of the total time for which there is a true object in the scene. Therefore a 0 indicates that there is no true objects within the data being evaluated so again an error message is probably more appropriate.

stonesoup/metricgenerator/tracktotruthmetrics.py
sdhiscocks: So I think raising an error needs to be avoided here, as wouldn't want to have to handle that when processing a batch run. Given you'd be looking at set of metrics as a whole, I don't see an issue with reporting that number of spurious tracks is zero, when number of tracks is also zero (which seems like valid value for the metric in that scenario).

sdhiscocks: I've updated the code to produce warnings, and hopefully the outputs are sensible in context of the warnings.

oharrald-Dstl (approved):

jswright-dstl (approved):

rjgreen-dstl (approved):

PR #220, opened 2020-05-13
Platform state mutable

rjgreen-dstl:
This merge changes the state property within the Platform object to be a StateMutableSequence rather than an individual State. All original behaviours remain the same however Platforms now maintain history of States. 

This has application in simulations, plotting and metric assessment (when using platforms rather than ground truth simulations). A Dummy simulation class has also been added which returns an empty set, this can be used in conjunction with the PlatformDetectionSimulator to produce simulations with only defined platform objects.

### DIFF
--- a/stonesoup/platform/tests/test_platform_base.py
+++ b/stonesoup/platform/tests/test_platform_base.py
@@ -543,7 +543,7 @@ def test_multi_transition():
     assert len(platform.transition_models) == 2
     assert len(platform.transition_times) == 2
 
-    #  Check that platform states lenght increases as platform moves
+    #  Check that platform states length increases as platform moves
     assert len(platform) == 1
     time = datetime.datetime.now()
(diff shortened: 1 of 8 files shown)


stonesoup/platform/tests/test_platform_base.py
```diff
@@ -534,19 +536,32 @@ def test_multi_transition():
...
     time = datetime.datetime.now()
+    time += datetime.timedelta(seconds=1)
+    platform.move(timestamp=time)
+    assert len(platform) == 2
+    time += datetime.timedelta(seconds=1)
+    platform.move(timestamp=time)
+    assert len(platform) == 3
+    time += datetime.timedelta(seconds=1)
+    platform.move(timestamp=time)
+    assert len(platform) == 4
```
oharrald-Dstl, line 557: For the sake of having less lines of code, is it worth putting these checks in a FOR loop?

PR #223, opened 2020-05-15
Sqrt kalma filter

jmbarr:
Several classes embodying a couple of versions of the 'Square Root Kalman' filter. Specifically, a `SqrtGaussianState` which inherits from `GaussianState` and is equivalent save for storing the Covariance matrix as W rather than P where P=WW^T.

`SqrtKalmanPredictor` and `SqrtKalmanUpdater` inherit from `KalmanPredictor` and `KalmanUpdater` respectively. I've had to re-engineer these latter slightly to enable more efficient inheritance, but I think the effect is positive. Each class has an option to do the predict or update step via a QR decomposition (due to @paulhorridge) or a more 'traditional' method (i.e. the first one in the textbooks, which is deprecated due to its inefficiency). 

The test cases demonstrate that nothing is broken and that they offer some benefit for cases where P becomes ill-conditioned. 

I've ignored the simple_sqrt_kalman.py file in the examples folder. I'm in favour of such things in general, but it's probably better as a notebook in a different repository. Thoughts on this?


stonesoup/predictor/kalman.py
```diff
@@ -65,7 +65,7 @@ def _transition_matrix(self, **kwargs):
         Parameters
         ----------
         **kwargs : various, optional
-            These are passed to :meth:`~.LinearGaussianTransitionModel.matrix`
+            These are passed to :math:`~.LinearGaussianTransitionModel.matrix`
```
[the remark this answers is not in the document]
jmbarr: ok

stonesoup/predictor/kalman.py
```diff
@@ -133,6 +132,56 @@ def _predict_over_interval(self, prior, timestamp):
...
         return predict_over_interval
 
+    def _predicted_covariance(self, trans_m, prior_cov, trans_cov, ctrl_mat,
+                              ctrl_noi):
+        """Private function to return the predicted covariance. Useful in that
+        it can be overwritten in children.
+
+        Parameters
+        ----------
+        trans_m : np.array
```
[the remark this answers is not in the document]
jmbarr: In all instances I assume.

stonesoup/predictor/kalman.py
```diff
@@ -376,3 +426,96 @@ def predict(self, prior, timestamp=None, **kwargs):
...
+            :math:`k`
+
+        Returns
+        -------
+        : :class:`~.SqrtGaussianState`
+            The predicted state (with the covariance encoded in lower triangular
+            form
+
+        """
+        return SqrtGaussianState(pred_mean, pred_covar, timestamp=timestamp)
```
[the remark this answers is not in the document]
jmbarr: So, if I get you right I create:

_prediction_class = Property(classmethod, default=GaussianStatePrediction,
                                 doc="trala") 

as a class attribute in KalmanPredictor, and equivalents as necessary in the child classes. No need then for extra methods.

Nice.

stonesoup/types/state.py
```diff
@@ -141,6 +141,50 @@ def mean(self):
         return self.state_vector
 
 
+class SqrtGaussianState(GaussianState):
```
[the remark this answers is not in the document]
jmbarr: ok, can give that a try...

stonesoup/predictor/kalman.py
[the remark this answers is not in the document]
sdhiscocks: Don't think it needs to be a `Property`. So just `_prediction_class = GaussianStatePrediction`

stonesoup/types/state.py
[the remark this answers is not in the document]
jmbarr: Doesn't seem to like the fact that `covar` is an attribute in the parent class, but now a method in the child.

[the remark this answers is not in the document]
jmbarr: I suppose it could just inherit from `State`, rather than `GaussianState`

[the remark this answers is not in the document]
jmbarr: Done that then. See what you think.

stonesoup/predictor/kalman.py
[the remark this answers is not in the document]
jmbarr: Done

jmbarr:
Hm, why the new flake8 errors?

sdhiscocks:
> Hm, why the new flake8 errors?

Due to update of flake8. You can ignore them for now, as proposed fix in #225

sdhiscocks:
@jmbarr - Could you rebase onto _master_ (and probably squash some commits as well), and migrated the example to `docs/examples` for building with sphinx-gallery.

jmbarr:
> @jmbarr - Could you rebase onto _master_ (and probably squash some commits as well), and migrated the example to `docs/examples` for building with sphinx-gallery.

Ok, that seems to have worked now. Not a rebase (because I don't really know what that is), rather a merge from master.

stonesoup/types/state.py
```diff
@@ -135,14 +135,48 @@ def __init__(self, state_vector, covar, *args, **kwargs):
...
+
+        Returns
+        -------
+        : :class:`~.CovarianceMatrix`
+            The covariance matrix calculated via :math:`W W^T`, where :math:`W` is a
+            :class:`~.SqrtCovarianceMatrix`
+
+        """
+        return self.sqrt_covar @ self.sqrt_covar.T
+
```
sdhiscocks, line 178: Could include a `GaussianState.register(SqrtGaussianState)` here, such that it qualifies as type of `GaussianState` in any instance checks.

stonesoup/predictor/kalman.py
```diff
@@ -376,3 +393,78 @@ def predict(self, prior, timestamp=None, **kwargs):
...
+        if hasattr(self.control_model, 'sqrt_control_noise'):
+            sqrt_ctrl_noi = self.control_model.sqrt_control_noise
+        else:
+            sqrt_ctrl_noi = la.sqrtm(self.control_model.control_noise)
+
+        if self.qr_method:
+            # Note that the control matrix aspect of this hasn't been tested
+            m_sq_trans_cov = np.block([[trans_m @ sqrt_prior_cov, sqrt_trans_cov,
+                                        ctrl_mat@sqrt_ctrl_noi]])
+            [_, pred_sqrt_cov] = np.linalg.qr(m_sq_trans_cov.T)
```
sdhiscocks, line 465: Brackets shouldn't be needed here
```suggestion
            _, pred_sqrt_cov = np.linalg.qr(m_sq_trans_cov.T)
```

stonesoup/predictor/kalman.py
```diff
@@ -133,6 +133,34 @@ def _predict_over_interval(self, prior, timestamp):
...
+        trans_m = self._transition_matrix(prior=prior, time_interval=predict_over_interval,
+                                          **kwargs)
+        trans_cov = self.transition_model.covar(time_interval=predict_over_interval, **kwargs)
+
+        # As this is Kalman-like, the control model must be capable of
+        # returning a control matrix (B)
+        ctrl_mat = self._control_matrix
+        ctrl_noi = self.control_model.control_noise
+
+        return trans_m @ prior_cov @ trans_m.T + trans_cov + ctrl_mat @ ctrl_noi @ ctrl_mat.T
```
sdhiscocks, line 162: Should group operations by order
```suggestion
        return trans_m@prior_cov@trans_m.T + trans_cov + ctrl_mat@ctrl_noi@ctrl_mat.T
```

stonesoup/predictor/kalman.py
```diff
@@ -376,3 +393,78 @@ def predict(self, prior, timestamp=None, **kwargs):
...
+        if self.qr_method:
+            # Note that the control matrix aspect of this hasn't been tested
+            m_sq_trans_cov = np.block([[trans_m @ sqrt_prior_cov, sqrt_trans_cov,
+                                        ctrl_mat@sqrt_ctrl_noi]])
+            [_, pred_sqrt_cov] = np.linalg.qr(m_sq_trans_cov.T)
+            return pred_sqrt_cov.T
+        else:
+            return np.linalg.cholesky(trans_m @ sqrt_prior_cov @ sqrt_prior_cov.T @ trans_m.T +
+                                      sqrt_trans_cov @ sqrt_trans_cov.T +
+                                      ctrl_mat @ sqrt_ctrl_noi @ sqrt_ctrl_noi.T @ ctrl_mat.T)
```
sdhiscocks, line 470: Would be better to group and lead with operators on line break.
```suggestion
            return np.linalg.cholesky(trans_m@sqrt_prior_cov@sqrt_prior_cov.T@trans_m.T
                                      + sqrt_trans_cov@sqrt_trans_cov.T
                                      + ctrl_mat@sqrt_ctrl_noi@ sqrt_ctrl_noi.T@ctrl_mat.T)
```

stonesoup/predictor/tests/test_kalman.py
```diff
@@ -72,3 +73,41 @@ def test_kalman(PredictorClass, transition_model,
...
+    sqrt_predictor = SqrtKalmanPredictor(transition_model=transition_model)
+    # Can swap out this method
+    sqrt_predictor = SqrtKalmanPredictor(transition_model=transition_model, qr_method=True)
+
+    # Perform and assert state prediction
+    prediction = predictor.predict(prior=prior, timestamp=new_timestamp)
+    sqrt_prediction = sqrt_predictor.predict(prior=sqrt_prior,
+                                             timestamp=new_timestamp)
+
+    assert(np.allclose(prediction.mean, sqrt_prediction.mean, 0, atol=1.e-14))
```
sdhiscocks, line 108: Minor issue, but asserts don't need brackets.

stonesoup/types/prediction.py
```diff
@@ -75,8 +83,8 @@ def __init__(self, state_vector, covar, timestamp=None,
                  cross_covar=None, *args, **kwargs):
         if(cross_covar is not None
            and cross_covar.shape[1] != state_vector.shape[0]):
-            raise ValueError("cross_covar should have the same number of "
-                             "columns as the number of rows in state_vector")
+            raise ValueError("cross_covar should have the same number of \
+                             columns as the number of rows in state_vector")
```
sdhiscocks, line 87: This was better before. This change add loads of white space to error message.

stonesoup/updater/kalman.py
```diff
@@ -339,3 +409,153 @@ def predict_measurement(self, predicted_state, measurement_model=None):
...
+    the 'full' covariance matrix :math:`P_{k|k-1} = L_{k|k-1} L^T_{k|k-1}`
+    [Eq1].
+
+    In its basic form :math:`L` is the lower triangular matrix returned via
+    Cholesky factorisation. There's no reason why other forms that satisfy Eq 1
+    above can't be used.
+
+    References
+    ----------
+    1. (to be added)
```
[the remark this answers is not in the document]
jmbarr: As this method is due to @paulhorridge, I should leave him to confirm where he got the method from. In the mean time I can guess where it came from.

stonesoup/types/state.py
jmbarr: Example of how that would work please. And what would it give?

sdhiscocks: The `ParticleState` does a similar thing: https://github.com/dstl/Stone-Soup/blob/4718a0f78050277b00d54b967997fa5516afa586/stonesoup/types/state.py#L203

This means checks like `isinstance(particle_state, State)` return `True`.

So this would mean that `isinstance(sqrt_state, GaussianState)` would return True as expected, and you have the `covar` method implemented already.

stonesoup/types/prediction.py
jmbarr: Hm, don't remember doing that one. Undone!

stonesoup/types/state.py
jmbarr: I think I see. This just means that now SqrtGaussianState is 'registered' as a subclass of GaussianState, even though it's not actually built that way.

stonesoup/predictor/kalman.py
```diff
@@ -133,6 +133,34 @@ def _predict_over_interval(self, prior, timestamp):
 
         return predict_over_interval
 
+    def _predicted_covariance(self, prior, predict_over_interval, **kwargs):
```
sglvladi, line 139: I feel like the name ``_get_predicted_covariance()`` seems more sensible, seeing as we are not extracting a property of the class (or the underlying models), but rather performing an computation. At the same time it makes the name quite lengthy so happy to leave as is.

stonesoup/updater/kalman.py
```diff
@@ -64,6 +65,14 @@ class KalmanUpdater(Updater):
             "specified on construction, or in the measurement, then error "
             "will be thrown.")
 
+    force_symmetric_covariance = Property(
+        bool, default=False, doc="A flag to force the output covariance matrix"
+                                 "to be symmetric by way of a simple geometric"
+                                 "combination of the matrix and transpose."
+                                 "Default is False.")
+
+    _update_class = GaussianStateUpdate
```
sglvladi, line 74: Same comment here as for ``KalmanPredictor._prediction_class``

stonesoup/updater/kalman.py
```diff
@@ -112,15 +121,83 @@ def _measurement_matrix(self, predicted_state=None, measurement_model=None,
         return self._check_measurement_model(
             measurement_model).matrix(**kwargs)
 
+    def _measurement_cross_covariance(self, predicted_state, measurement_matrix):
```
sglvladi, line 124: As before maybe ``_get_measurement_cross_covariance()``?

stonesoup/updater/kalman.py
```diff
@@ -112,15 +121,83 @@ def _measurement_matrix(self, predicted_state=None, measurement_model=None,
...
+
+        Returns
+        -------
+        : numpy.ndarray
+            The innovation covariance
+
+        """
+        return meas_mat @ m_cross_cov + meas_mod.covar()
+
+    def _posterior_covariance(self, hypothesis):
```
sglvladi, line 164: Maybe ``_get_posterior_covariance()``?

stonesoup/updater/kalman.py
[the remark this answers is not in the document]
jmbarr: Added something but feel free to change, @paulhorridge.

stonesoup/predictor/kalman.py
```diff
@@ -376,3 +393,78 @@ def predict(self, prior, timestamp=None, **kwargs):
...
+    the Gaussian state, :class:`~.SqrtGaussianState`.
+
+    The prediction is undertaken in one of two ways. The default is to work in exactly the same
+    way as the parent class, with the exception exception that the predicted covariance is
+    subject to a Cholesky factorisation prior to initialisation of the :class:`~.SqrtGaussianState`
+    output. The alternative, accessible via the :attr:`qr_method = True` flag, is to predict via a
+    modified Gram-Schmidt process. See [1] for details.
+
+    If transition and control models are possessed of the square root form of the covariance (as
+    :attr:`sqrt_covar` in the case of the transition model and :attr:`sqrt_control_noi` for
```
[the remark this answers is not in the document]
jmbarr: I'd be in favour of having a proper look at control models as a separate issue. 

They (control models) exist because they have to in the general case of the Kalman predictors, but they're largely untested and, as you've pointed out, their analogy with other models is a little tenuous. I think more eyes on the issue would help, as well as considering uses they might have outside the Kalman family.

stonesoup/predictor/kalman.py
jmbarr: I thought class.get_something() was considered somewhat non-pythonic, but I may be confused about why that is (something to do with properties). And I prefer the shorter names.

But maybe you want to discourage someone doing `class._predicted_covariance = AMatrix`.

Happy to receive guidance on getters/setters.

stonesoup/predictor/kalman.py
```diff
@@ -376,3 +393,78 @@ def predict(self, prior, timestamp=None, **kwargs):
...
+        sqrt_prior_cov = prior.sqrt_covar
+
+        trans_m = self._transition_matrix(prior=prior, time_interval=predict_over_interval,
+                                          **kwargs)
+        if hasattr(self.transition_model, 'sqrt_covar'):
+            sqrt_trans_cov = self.transition_model.sqrt_covar(time_interval=predict_over_interval,
+                                                              **kwargs)
+        else:
+            sqrt_trans_cov = la.sqrtm(self.transition_model.covar(
+                time_interval=predict_over_interval, **kwargs))
```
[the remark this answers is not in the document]
jmbarr: Dunno. Do you mean that we'll only have to change one instance of `sqrt_covar` if that's to be renamed?

stonesoup/predictor/kalman.py
sglvladi: > But maybe you want to discourage someone doing `class._predicted_covariance = AMatrix`.

Correct. Personally, when I seen these namings my head thinks it's an attribute. But maybe that's just me.

In the case that ``_get_smth()`` is problematic, ``_calc_smth()`` could be used instead but this makes the names even longer.

stonesoup/predictor/kalman.py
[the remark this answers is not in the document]
sglvladi: Ah, I thought I had deleted that last line. I suppose it doesn't make a massive difference, but yeah that was my initial though.

I also remember that in the pre-beta era we were strongly advocating the use of EAFP (i.e. try-catch) over type checking (isinstance, hasattr, etc.), but not too sure were we stand on this now, as a lot of the code that did EAFP has been phased out. I thought I'd mention it just in case.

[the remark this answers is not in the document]
sdhiscocks: Yeah, EAFP is usually preferred style in Python, so should go with that. Also better fits with duck typing approach in Python.

stonesoup/predictor/kalman.py
sdhiscocks: I think @sglvladi makes a good point, and name does look more like a property, than a method. I think _calculate_ is better than _get_ to make it clear your calculating something, rather than just getting something. I think it's better to have clearer names, than worry about the length.

stonesoup/predictor/kalman.py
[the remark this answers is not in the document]
jmbarr: Ok, hereby asking for forgiveness. (4 times as far as I could see.)

stonesoup/predictor/kalman.py
jmbarr: Can we leave this for now? At least until we've agreed a 'corporate policy'. This argument also applies to `._transition_matrix()` in the current class, but could easily extend to stuff like the `.matrix()` or `.covar()` methods in the general models, neither of which are properties but return things that look like them. It feels like we should think of a consistent approach before I set off applying a smattering of semi-consistency.

PR #225, opened 2020-05-18
Fix flake8 errors introduced with recent flake8 update

sdhiscocks:


stonesoup/predictor/tests/test_kalman.py
```diff
@@ -51,17 +51,12 @@ def test_kalman(PredictorClass, transition_model,
...
-                                time_interval=time_interval)@prior.mean,
-        transition_model.matrix(timestamp=new_timestamp,
-                                time_interval=time_interval)
-        @prior.covar
-        @transition_model.matrix(timestamp=new_timestamp,
-                                 time_interval=time_interval).T
-        + transition_model.covar(timestamp=new_timestamp,
-                                 time_interval=time_interval))
+        transition_model_matrix @ prior.mean,
+        transition_model_matrix@prior.covar@transition_model_matrix.T + transition_model_covar)
```
oharrald-Dstl, line 59: Is there a reason that in some places '@' is surround by space, and in others it is not?

sdhiscocks: It's from [PEP-8 recommendation](https://www.python.org/dev/peps/pep-0008/#other-recommendations) to with operator priorities.

PR #226, opened 2020-05-26
Add dynamic UTM

sdhiscocks:
This allows each detection to belong to a different zone.

This also allows tracks across multiple UTM zones to be compared to detections, and will then also allow tracks to transition UTM zones via the Updater, due to prediction being in UTM zone of detection.

This is dependent on Turbo87/utm#46 and Turbo87/utm#47 being fixed in cases where tracks pass equator and anti-meridian.

TODO:
- [ ] Add tests for hypothesiser wrapper.

### DIFF
--- a/stonesoup/feeder/tests/test_geo.py
+++ b/stonesoup/feeder/tests/test_geo.py
@@ -16,7 +16,7 @@ def detector():
         @BufferedGenerator.generator_method
         def detections_gen(self):
             for i in range(-3, 4):
-                detections = {Detection([[i], [50], [5000 + i*10]])}
+                detections = {Detection([[i], [50 + i], [5000 + i*10]])}
                 yield None, detections
 
     return Detector()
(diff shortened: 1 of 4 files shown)


stonesoup/feeder/tests/test_geo.py
```diff
@@ -43,15 +45,76 @@ def test_utm_converter(detector):
...
+
+        assert pytest.approx((lat, long), rel=1e-2, abs=1e-4) == utm.to_latlon(
+            *detection.state_vector[0:2], zone_number=31, zone_letter='U', strict=False)
+
+        # Correct zone shouldn't match, as conversion is wrong
+        assert pytest.approx((lat, long)) != utm.to_latlon(
+            *detection.state_vector[0:2], zone_number=30, zone_letter='U', strict=False)
+
+
+def test_utm_converter_dynamic(detector):
```
oharrald-Dstl, line 92: Adding a test for the case when `dynamic`, `zone_number` and `zone_letter` attributes are all pre-set, raising `ValueError("Cannot have both dynamic and fixed zone") ` will increase code coverage to 100%

oharrald-Dstl (changes requested):
Looks good so far. Just waiting on hypothesiser tests and Turbo87/utm#46 to be merged.

PR #228, opened 2020-06-01
Documentation updates

sdhiscocks:
A number of small documentation fixes, added doc string to _Property_ instances (so they populate attributes in documentation), added to _Contributing_ page and rearrange table of contents, including renaming data flow to design.


sdhiscocks:
> Change looks good to me, makes the docs a lot neater.
> 
> Only comment is that we probably need to think about making the amendment to the framework design section to demonstrate the State Estimation rather than purely tracking. That probably requires a little more thought so should be a separate pull request.

Yeah, I forgot to add that whilst I've renamed data flow section, I've not added anything else to it yet, but intend to in future PR. It would be good to also have an example of state estimation we can link to.

PR #234, opened 2020-06-09
Add bearing-only measurement model

oharrald-Dstl:
This model takes 2D positional cartesian coordinates and outputs bearing-only measurements.
Appropriate modifications to base class and measurement initiator were made to accommodate a 1D output.

### DIFF
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -554,6 +554,103 @@ class CartesianToElevationBearing(NonLinearGaussianMeasurement):
         return out
 
 
+class Cartesian2DToBearing(NonLinearGaussianMeasurement):
+    r"""This is a class implementation of a time-invariant measurement model, where measurements \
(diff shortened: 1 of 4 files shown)


stonesoup/models/measurement/nonlinear.py
```diff
@@ -554,6 +554,103 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+
+      v_t \sim \mathcal{N}(0,\sigma_{\phi}^2)
+
+    The :py:attr:`mapping` property of the model is a 2 element vector, whose first \
+    (i.e. :py:attr:`mapping[0]`) and second (i.e. :py:attr:`mapping[1]`) elements contain the \
+    state index of the :math:`x` and :math:`y` coordinates, respectively.
+
+    """  # noqa:E501
+
+    translation_offset = Property(StateVector, default=StateVector(np.array([[0], [0]])),
```
sdhiscocks, line 585: default here should be `None`

PR #236, opened 2020-06-11
Add Data Association Tutorials

oharrald-Dstl:
Clutter and multi-target tracking tutorials with diagrams for each in the source/_static folder.


docs/tutorials/05_DataAssociation-Clutter.py
```diff
@@ -0,0 +1,226 @@
...
+    # Plot actual detections.
+    ax.scatter([state.state_vector[0] for state in set_ if isinstance(state, TrueDetection)],
+               [state.state_vector[1] for state in set_ if isinstance(state, TrueDetection)],
+               color='b')
+    # Plot clutter.
+    ax.scatter([state.state_vector[0] for state in set_ if isinstance(state, Clutter)],
+               [state.state_vector[1] for state in set_ if isinstance(state, Clutter)],
+               color='y',
+               marker='2')
+fig
```
sdhiscocks, line 117: Creating figure and having fig in same block causes plot to be shown twice in notebooks.

docs/tutorials/05_DataAssociation-Clutter.py
```diff
@@ -0,0 +1,217 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+5 - Data Association - clutter tutorial
+=======================================
+"""
+
+# %%
```
jmbarr, line 9: Wonder if we might want to talk a bit more about data association in general (analogously to the 'bit' on Bayesian estimation in the Kalman filter tutorial).

docs/tutorials/05_DataAssociation-Clutter.py
```diff
@@ -0,0 +1,217 @@
...
+# Simulating clutter
+# ^^^^^^^^^^^^^^^^^^
+# Next, generate some measurements and add in some clutter
+# at each time-step. We use the :class:`~.TrueDetection` and :class:`~.Clutter` subclasses of
+# :class:`~.Detection` to help with discerning data types later in plotting. We'll introduce the
+# possibility that, at any time-step, our sensor receives no detection from the target (i.e.
+# :math:`p_d < 1`). A fixed number of clutter points are generated uniformly distributed across
+# a region around the target.
```
jmbarr, line 51: Could be a bit more explicit about how uniform the clutter actually is (i.e. +-20 around the target).

docs/tutorials/06_DataAssociation-MultiTargetTutorial.py
```diff
@@ -0,0 +1,224 @@
...
+# Tracking multiple targets through clutter
+# -----------------------------------------
+#
+# More often than not, the difficult part of state estimation concerns the ambiguous association of
+# predicted states with measurements. This happens whenever there is more that one target under
+# consideration, there are false alarms or clutter, targets can appear and disappear. That is to say
+# it happens everywhere.
+#
+# In this tutorial we introduce **global nearest neighbour** data association, which
+# attempts to find a globally-consistent collection of hypotheses such that some overall probability
```
jmbarr, line 19: Shouldn't really have used 'probability' here. 'Score' or 'measure' would be better.

docs/tutorials/05_DataAssociation-Clutter.py
jmbarr: I only had a brief look but I think that, in this instance, clutter is generated by sampling uniformly from a rectangular area around the true position of the target a specified number of times. What the sensor is comes later.

Perhaps whoever wrote originally can confirm? @sdhiscocks @oharrald-Dstl ?

docs/tutorials/05_DataAssociation-Clutter.py
oharrald-Dstl: I'm a bit unclear on that. Could you make any changes you see appropriate?

sdhiscocks: @jmbarr and @oharrald-Dstl: Agree, but let's raise another PR for this.

PR #239, opened 2020-06-12
Add (J)PDA Tutorial and Diagrams

oharrald-Dstl:
This tutorial covers PDA and JPDA examples.

Notes: Covariances for JPDA example are quite large (maybe calculated wrong somewhere, or just naturally big in JPDA?)


docs/tutorials/07_PDATutorial.py
```diff
@@ -0,0 +1,385 @@
...
+    orient = np.arctan2(v[max_ind, 1], v[max_ind, 0])
+    ellipse = Ellipse(xy=state.state_vector[(0, 2), 0],
+                      width=np.sqrt(w[0])*2, height=np.sqrt(w[1])*2,
+                      angle=np.rad2deg(orient),
+                      alpha=0.2)
+    ax.add_artist(ellipse)
+fig
+
+# %%
+# Joint Probabilistic Data Association
```
sdhiscocks, line 215: @oharrald-Dstl - I know we discussed have this as one tutorial, but I'm thinking now it's worth splitting into two, as it's a little long like this.

jmbarr:
> This tutorial covers PDA and JPDA examples.
> 
> Notes: Covariances for JPDA example are quite large (maybe calculated wrong somewhere, or just naturally big in JPDA?)

We'd expect them to be bigger, to account for the ambiguity. The size will depend on clutter rate, position, size of gate. They didn't look outrageous on my plot.

PR #240, opened 2020-06-15
Modify sensors to measure sets of ground truth states

oharrald-Dstl:
Changes to sensor tests to account for ground truths being passed-in, and measurements being received, as sets.
As well as a bit of adjustment in state definitions in test files to account for larger character limit.

### DIFF
--- a/stonesoup/sensor/passive.py
+++ b/stonesoup/sensor/passive.py
@@ -52,8 +52,9 @@ class PassiveElevationBearing(Sensor):
         Returns
         -------
         Set[:class:`~.Detection`]
-            A measurement generated from the given state. The timestamp of the\
-            measurement is set equal to that of the provided state.
+            A set of measurements generated from the given states. The timestamps of the
+            measurements are set equal to that of the corresponding states that they were
+            calculated from.
         """
 
         measurement_model = CartesianToElevationBearing(
--- a/stonesoup/sensor/passive.py
+++ b/stonesoup/sensor/passive.py
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
 import numpy as np
 
-from stonesoup.sensor.sensor import Sensor
+from typing import Set, Union
+
 from ..base import Property
 from ..models.measurement.nonlinear import CartesianToElevationBearing
+from ..sensor.sensor import Sensor
 from ..types.array import CovarianceMatrix
 from ..types.detection import Detection
+from ..types.groundtruth import GroundTruthState
 
 
 class PassiveElevationBearing(Sensor):
@@ -33,12 +36,13 @@ class PassiveElevationBearing(Sensor):
             (and follow in format) the underlying \
             :class:`~.CartesianToElevationBearing` model")
 
-    def measure(self, ground_truths, noise=True, **kwargs):
+    def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray, bool] = True,
+                **kwargs) -> Set[Detection]:
         """Generate a measurement for a given state
 
         Parameters
         ----------
-        ground_truths : :class:`~.State`
+        ground_truths : Set[:class:`~.GroundTruthState`]
             A set of :class:`~.GroundTruthState`
         noise: :class:`numpy.ndarray` or bool
             An externally generated random process noise sample (the default is
@@ -47,7 +51,7 @@ class PassiveElevationBearing(Sensor):
 
         Returns
         -------
-        :class:`~.Detection`
+        Set[:class:`~.Detection`]
             A measurement generated from the given state. The timestamp of the\
             measurement is set equal to that of the provided state.
         """
--- a/stonesoup/sensor/passive.py
+++ b/stonesoup/sensor/passive.py
@@ -33,13 +33,13 @@ class PassiveElevationBearing(Sensor):
             (and follow in format) the underlying \
             :class:`~.CartesianToElevationBearing` model")
 
-    def measure(self, ground_truth, noise=True, **kwargs):
+    def measure(self, ground_truths, noise=True, **kwargs):
         """Generate a measurement for a given state
 
         Parameters
         ----------
-        ground_truth : :class:`~.State`
-            A ground-truth state
+        ground_truths : :class:`~.State`
+            A set of :class:`~.GroundTruthState`
         noise: :class:`numpy.ndarray` or bool
             An externally generated random process noise sample (the default is
             `True`, in which case :meth:`~.Model.rvs` is used
@@ -59,9 +59,13 @@ class PassiveElevationBearing(Sensor):
             translation_offset=self.position,
             rotation_offset=self.orientation)
 
-        measurement_vector = measurement_model.function(
-            ground_truth, noise=noise, **kwargs)
+        detections = set()
+        for truth in ground_truths:
 
-        return Detection(measurement_vector,
-                         measurement_model=measurement_model,
-                         timestamp=ground_truth.timestamp)
+            measurement_vector = measurement_model.function(truth, noise=noise, **kwargs)
+            detection = Detection(measurement_vector,
+                                  measurement_model=measurement_model,
+                                  timestamp=truth.timestamp)
+            detections.add(detection)
+
+        return detections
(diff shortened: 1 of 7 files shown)


stonesoup/sensor/radar/radar.py
```diff
@@ -645,13 +663,13 @@ def gen_probability(self, sky_state):
         return det_prob, snr, rcs, directed_power,\
             10 * np.log10(spoiled_gain), spoiled_width
 
-    def measure(self, sky_state, noise=True, **kwargs):
+    def measure(self, sky_states, noise=True, **kwargs):
```
sdhiscocks, line 666: `sky_states` should be renamed to `ground_truths` like other classes.

stonesoup/sensor/radar/radar.py
```diff
@@ -665,13 +683,20 @@ def measure(self, sky_state, noise=True, **kwargs):
...
+
+        detections = set()
+        for sky_state in sky_states:
+            det_prob = self.gen_probability(sky_state)[0]
+            # Is the state detected?
+            if np.random.rand() <= det_prob:
+                measurement_model = copy.deepcopy(self.measurement_model)
+                measurement_model.translation_offset = self.position.copy()
+                measurement_model.rotation_offset = self.rotation_offset.copy()
+                measured_pos = self.measurement_model.function(sky_state, noise=noise)
```
sdhiscocks, line 695: I think this shouldn't have `self` prefixed.

stonesoup/simulator/platform.py
```diff
@@ -28,14 +28,8 @@ def detections_gen(self):
...
-                    for truth in truths.union(self.platforms):
-
-                        # Make sure platform's sensors do not measure itself
-                        if truth is platform:
-                            continue
-
-                        detection = sensor.measure(truth)
-                        if detection is not None:
-                            detections.add(detection)
+                    truths_to_be_measured = truths.union([other_platform for other_platform in
```
sdhiscocks, line 31: Could use `truths.union(self.platforms) - {platform}`

stonesoup/sensor/passive.py
```diff
@@ -33,13 +33,13 @@ class PassiveElevationBearing(Sensor):
             (and follow in format) the underlying \
             :class:`~.CartesianToElevationBearing` model")
 
-    def measure(self, ground_truth, noise=True, **kwargs):
+    def measure(self, ground_truths, noise=True, **kwargs):
```
etfrogers-dstl, line 36: Is it worth type hinting these?
```suggestion
    def measure(self, ground_truths: Set[State], noise: Union[np.ndarray, bool] = True, **kwargs) -> Set[Detection]:
```
or (depending on whether the text or the annotation is correct (see comment below)
```suggestion
    def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray, bool] = True, **kwargs) -> Set[Detection]:
```

Of course, there are a lot of places to apply this change...

stonesoup/sensor/passive.py
```diff
@@ -33,13 +33,13 @@ class PassiveElevationBearing(Sensor):
...
         """Generate a measurement for a given state
 
         Parameters
         ----------
-        ground_truth : :class:`~.State`
-            A ground-truth state
+        ground_truths : :class:`~.State`
+            A set of :class:`~.GroundTruthState`
```
etfrogers-dstl, line 42: These two seem inconsistent. Is it meant to be:

```suggestion
        ground_truths : :class:`~.GroundTruthState`
            A set of :class:`~.GroundTruthState`
```

Again, many places to update this...


Also, we need to update the return docs
``` 
        Returns
        -------
        Set[:class:`~.Detection`]
            A set of measurements generated from the given states. The timestamp of the
            measurement is set equal to that of the provided state.
```

etfrogers-dstl (commented):
I can see the argument for this change, but it feels like it makes the single case more complicated (though I guess you just add a set literal `sensor.measure({truth}, ...)`)

Is it more natural this way (that you would always have a lot of targets)? Not sure.

An alternative approach (which would require much less code) might be to edit the `BaseSensor` class:
```python
class BaseSensor(Base, ABC):
   ...
   @abstractmethod
    def measure(self, truth: State, **kwargs) -> Detection:  # Updated signature
        raise NotImplementedError

    def measure_set(truths: Set[State], **kwargs) -> Set[Detection]:
        return {self.measure(truth) for truth in truths}
```

Is there something I haven't considered here?

oharrald-Dstl:
> I can see the argument for this change, but it feels like it makes the single case more complicated (though I guess you just add a set literal `sensor.measure({truth}, ...)`)
> 
> Is it more natural this way (that you would always have a lot of targets)? Not sure.
> 
> An alternative approach (which would require much less code) might be to edit the `BaseSensor` class:
> 
> ```python
> class BaseSensor(Base, ABC):
>    ...
>    @abstractmethod
>     def measure(self, truth: State, **kwargs) -> Detection:  # Updated signature
>         raise NotImplementedError
> 
>     def measure_set(truths: Set[State], **kwargs) -> Set[Detection]:
>         return {self.measure(truth) for truth in truths}
> ```
> 
> Is there something I haven't considered here?

I'm uncertain whether the majority of scenarios would involve measuring more than one truth or not when using a tracking framework. The addition of a 'measure_set' method sounds like a better idea. What are people's opinions on this?

sdhiscocks:
> > I can see the argument for this change, but it feels like it makes the single case more complicated (though I guess you just add a set literal `sensor.measure({truth}, ...)`)
> > Is it more natural this way (that you would always have a lot of targets)? Not sure.
> > An alternative approach (which would require much less code) might be to edit the `BaseSensor` class:
> > ```python
> > class BaseSensor(Base, ABC):
> >    ...
> >    @abstractmethod
> >     def measure(self, truth: State, **kwargs) -> Detection:  # Updated signature
> >         raise NotImplementedError
> > 
> >     def measure_set(truths: Set[State], **kwargs) -> Set[Detection]:
> >         return {self.measure(truth) for truth in truths}
> > ```
> > 
> > 
> > Is there something I haven't considered here?
> 
> I'm uncertain whether the majority of scenarios would involve measuring more than one truth or not when using a tracking framework. The addition of a 'measure_set' method sounds like a better idea. What are people's opinions on this?

I think it's much better to have a single interface that supports multiple (like most things in Stone Soup). This change was proposed as sensor performance may vary depending on number of targets, like ability to differentiate close objects. I think single is the special case, and multi-targets is much more common.

etfrogers-dstl:
> > > I can see the argument for this change, but it feels like it makes the single case more complicated (though I guess you just add a set literal `sensor.measure({truth}, ...)`)
> > > Is it more natural this way (that you would always have a lot of targets)? Not sure.
> > > An alternative approach (which would require much less code) might be to edit the `BaseSensor` class:
> > > ```python
> > > class BaseSensor(Base, ABC):
> > >    ...
> > >    @abstractmethod
> > >     def measure(self, truth: State, **kwargs) -> Detection:  # Updated signature
> > >         raise NotImplementedError
> > > 
> > >     def measure_set(truths: Set[State], **kwargs) -> Set[Detection]:
> > >         return {self.measure(truth) for truth in truths}
> > > ```
> > > 
> > > 
> > > Is there something I haven't considered here?
> > 
> > 
> > I'm uncertain whether the majority of scenarios would involve measuring more than one truth or not when using a tracking framework. The addition of a 'measure_set' method sounds like a better idea. What are people's opinions on this?
> 
> I think it's much better to have a single interface that supports multiple (like most things in Stone Soup). This change was proposed as sensor performance may vary depending on number of targets, like ability to differentiate close objects. I think single is the special case, and multi-targets is much more common.

I yield to @sdhiscocks expertise on the the philosophy. We **could** try and tweak the functions so they take a single target or a set, but that would probably over complicate the code in far too many places. OK, leave this bit as is (though my other comments still stand).

stonesoup/sensor/passive.py
etfrogers-dstl: Actually, should this be:
```suggestion
        ground_truths : :class:Set [:class:`~.GroundTruthState`]
            A set of :class:`~.GroundTruthState`
```
or something similar (I haven't checked how that renders).

stonesoup/sensor/passive.py
oharrald-Dstl: Would the type hints be necessary if the argument types are documented?

etfrogers-dstl (approved):
Looks good here

PR #242, opened 2020-06-17
Add tutorial to highlight simulation and tracking components

sdhiscocks:
This is added to direct users to the existing components that implement the simulation and tracking functionality built up in other tutorials.

### DIFF
--- a/docs/tutorials/10_Simulation_&_Tracking_Components.py
+++ b/docs/tutorials/10_Simulation_&_Tracking_Components.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+10 - Simulation & Tracking Components
+=====================================
+Running through the tutorials, the simulation code and tracking code has developed as new elements
+have been added. In the last tutorial, this ended with a simulation of varying number of targets,
+and a multi-target tracking handling initialisation and deletion of tracks. This tutorial just
+wraps up these elements by showing the Stone Soup components that implement the similar simulation
+and tracking features in existing components.
+"""
+
+# %%
+# Creating Simulation
+# -------------------
+# First create models as seen in previous tutorials
+import numpy as np
+
+from stonesoup.models.transition.linear import (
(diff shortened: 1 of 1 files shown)


docs/tutorials/10_Simulation_&_Tracking_Components.py
```diff
@@ -0,0 +1,123 @@
...
+groundtruth_sim = MultiTargetGroundTruthSimulator(
+    transition_model=transition_model,
+    initial_state=GaussianState(
+        StateVector([[0], [0], [0], [0]]),
+        CovarianceMatrix(np.diag([4, 0.5, 4, 0.5]))),
+    timestep=datetime.timedelta(seconds=5),
+    number_steps=20,
+    birth_rate=0.3,
+    death_probability=0.05
+)
```
rjgreen-dstl, line 48: I think it may be worth adding code snippets which show the original code which is now wrapped up here (or pointing to the specific tutorial section). That would allow the reader to directly compare without having two files open

docs/tutorials/10_Simulation_&_Tracking_Components.py
```diff
@@ -0,0 +1,123 @@
...
+    updater=updater,
+    min_points=3,
+)
+
+# %%
+# Creating and Running the Tracker
+# --------------------------------
+# With the components created, the multi-target tracker component will be created, constructed from
+# the components created above. This is logically the same as tracking code in the previous
+# tutorial.
```
rjgreen-dstl, line 95: as above

docs/tutorials/10_Simulation_&_Tracking_Components.py
sdhiscocks: I've added links to sections (note these will be broken until #238 is also merged.

sdhiscocks:
> What's this one for? Just a consolidation of previous tutorials, or is the plan to add more to it?

Yes, just that.

PR #246, opened 2020-06-23
Range velocity binning

jjosborne-dstl:
Added a Cartesian to elevation, bearing, range and range rate measurement model that bins the range and range rate based on respective resolutions. The model is called RangeRangeRateBinning.

### DIFF
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -1111,10 +1111,9 @@ class RangeRangeRateBinning(CartesianToElevationBearingRangeRate):
 
         The probability for the binned dimensions, the last 2, can be written as:
 
-        ..math::
+        .. math::
 
-            p = P(a $\leq$ \mathcal{N} $\leq$ b)
-
+            p = P(a \leq \mathcal{N} \leq b)
 
         In this equation a and b are the edges of the bin.
 
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -1134,7 +1134,6 @@ class RangeRangeRateBinning(CartesianToElevationBearingRangeRate):
                 and ((state1.state_vector[3, 0]-self.range_rate_res/2) /
                      self.range_rate_res).is_integer()):
             mean_vector = self.function(state2, noise=False, **kwargs)
-
             # pdf for the angles
             az_el_pdf = multivariate_normal.pdf(
                 state1.state_vector[:2, 0],
@@ -1152,7 +1151,6 @@ class RangeRangeRateBinning(CartesianToElevationBearingRangeRate):
                 mean_vector[3, 0],
                 self.range_rate_res,
                 self.covar()[3])
-
             return Probability(range_pdf * velocity_pdf * az_el_pdf)
         else:
             return Probability(0)
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -998,8 +998,8 @@ class RangeRangeRateBinning(CartesianToElevationBearingRangeRate):
     .. math::
 
       h(\vec{x}_t,\vec{v}_t) = \begin{bmatrix}
-                asin(\mathcal{z}/\sqrt{\mathcal{x}^2 + \mathcal{y}^2 +\mathcal{z}^2}) \\
-                atan2(\mathcal{y},\mathcal{x}) \\
+                \textrm{asin}(\mathcal{z}/\sqrt{\mathcal{x}^2 + \mathcal{y}^2 +\mathcal{z}^2}) \\
+                \textrm{atan2}(\mathcal{y},\mathcal{x}) \\
                 \sqrt{\mathcal{x}^2 + \mathcal{y}^2 + \mathcal{z}^2} \\
                 (x\dot{x} + y\dot{y} + z\dot{z})/\sqrt{x^2 + y^2 + z^2}
                 \end{bmatrix} + \vec{v}_t
@@ -1026,7 +1026,7 @@ class RangeRangeRateBinning(CartesianToElevationBearingRangeRate):
 
     .. math::
 
-        x = \textrm{floor(x/\Delta x)*\Delta x
+        x = \textrm{floor}(x/\Delta x)*\Delta x
 
     The :py:attr:`mapping` property of the model is a 3 element vector, \
     whose first (i.e. :py:attr:`mapping[0]`), second (i.e. \
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -1080,7 +1080,7 @@ class RangeRangeRateBinning(CartesianToElevationBearingRangeRate):
             if noise:
                 out[2] = np.floor(out[2] / self.range_res) * self.range_res + self.range_res/2
                 out[3] = np.floor(out[3] / self.range_rate_res) * \
-                         self.range_rate_res + self.range_rate_res/2
+                    self.range_rate_res + self.range_rate_res/2
 
         return out
 
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -1079,13 +1079,13 @@ class RangeRangeRateBinning(CartesianToElevationBearingRangeRate):
         if isinstance(noise, bool) or noise is None:
             if noise:
                 out[2] = np.floor(out[2] / self.range_res) * self.range_res + self.range_res/2
-                out[3] = np.floor(out[3] / self.range_rate_res) * self.range_rate_res + self.range_rate_res/2
+                out[3] = np.floor(out[3] / self.range_rate_res) * \
+                         self.range_rate_res + self.range_rate_res/2
 
         return out
 
     def _gaussian_integral(self, a, b, mean, cov):
         # the cumlative probability ranging from a to b for a normal distribution
-        # \frac{1}{\sigma\sqrt{2\pi}}\int_{a}^{b}e^{-\frac{1}{2}\left(\frac{x-\mu}{\sigma}\right)^{2}}
         return (multivariate_normal.cdf(a, mean=mean, cov=cov)
                 - multivariate_normal.cdf(b, mean=mean, cov=cov))
 
@@ -1133,7 +1133,8 @@ class RangeRangeRateBinning(CartesianToElevationBearingRangeRate):
         # state1 is in measurement space
         # state2 is in state_space
         if (((state1.state_vector[2, 0]-self.range_res/2) / self.range_res).is_integer()
-                and ((state1.state_vector[3, 0]-self.range_rate_res/2) / self.range_rate_res).is_integer()):
+                and ((state1.state_vector[3, 0]-self.range_rate_res/2) /
+                     self.range_rate_res).is_integer()):
             mean_vector = self.function(state2, noise=False, **kwargs)
 
             # pdf for the angles
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -5,8 +5,10 @@ from typing import Sequence, Tuple, Union
 
 import numpy as np
 from scipy.linalg import inv, pinv, block_diag
+from scipy.stats import multivariate_normal
 
 from ...base import Property
+from ...types.numeric import Probability
 
 from ...functions import cart2pol, pol2cart, \
     cart2sphere, sphere2cart, cart2angles, \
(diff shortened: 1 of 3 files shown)


stonesoup/models/measurement/nonlinear.py
```diff
@@ -965,3 +967,192 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+                    out[1] / self.range_rate_res) * self.range_rate_res
+
+        return out
+
+    def _gaussian_integral(self, a, b, mean, cov):
+        # the cumlative probability ranging from a to b for a normal distribution
+        # \frac{1}{\sigma\sqrt{2\pi}}\int_{a}^{b}e^{-\frac{1}{2}\left(\frac{x-\mu}{\sigma}\right)^{2}}
+        return (multivariate_normal.cdf(a, mean=mean,
+                                        cov=cov) - multivariate_normal.cdf(
+            b, mean=mean, cov=cov))
```
etfrogers-dstl, line 1091: The indentation here can probably be improved

```suggestion
        return (multivariate_normal.cdf(a, mean=mean, cov=cov) 
               - multivariate_normal.cdf(b, mean=mean, cov=cov))
```

stonesoup/models/measurement/nonlinear.py
```diff
@@ -965,3 +967,192 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+        -------
+        :class:`numpy.ndarray` of shape (:py:attr:`~ndim_state`, 1)
+            The model function evaluated given the provided time interval.
+        """
+
+        out = super().function(state, noise, **kwargs)
+
+        if isinstance(noise, bool) or noise is None:
+            if noise:
+                out[0] = np.floor(out[0] / self.range_res) * self.range_res
```
etfrogers-dstl, line 1080: If there is no `noise`, or `noise` is supplied then don't bin. Is that logical? Why would whether you bin depend on whether noise is added?

stonesoup/models/measurement/nonlinear.py
```diff
@@ -965,3 +967,192 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+        : :class:`~.Probability`
+            The likelihood of ``state1``, given ``state2``
+        """
+
+        # state_vector1 is in measurement space
+        # state_vector2 is in state_space
+        if ((state1.state_vector[0, 0] / self.range_res).is_integer()
+                and (state1.state_vector[
+                         1, 0] / self.range_rate_res).is_integer()):
+            mean_vector = self.function(state2, noise=False, **kwargs)
```
etfrogers-dstl, line 1139: Do I understand correctly that if the state has not been binned the PDF is zero. Is that correct? And should it be that way?

stonesoup/models/measurement/tests/test_models.py
```diff
@@ -562,3 +563,67 @@ def test_inverse_function():
...
+                                              mapping=[0, 2, 4],
+                                              velocity_mapping=[1, 3, 5],
+                                              noise_covar=np.array([100,
+                                                                   10,
+                                                                   np.pi/18,
+                                                                   np.pi/18]))
+
+    measured = measurement_model.function(real_state, noise=True)
+    pdf = measurement_model.pdf(State(measured), real_state)
+    assert pdf != 0
```
etfrogers-dstl, line 601: Should there be a test (not necessarily here) that check the pdf value, maybe against a manual calculation?

stonesoup/models/measurement/nonlinear.py
```diff
@@ -965,3 +967,192 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+                \dot{r}
+            \end{bmatrix}
+
+    * :math:`h` is a non-linear model function of the form:
+
+    .. math::
+
+      h(\vec{x}_t,\vec{v}_t) = \begin{bmatrix}
+                asin(\mathcal{z}/\sqrt{\mathcal{x}^2 + \mathcal{y}^2 +\mathcal{z}^2}) \\
+                atan2(\mathcal{y},\mathcal{x}) \\
```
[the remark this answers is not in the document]
jjosborne-dstl: I'm going to leave these as they are so that it specifies the type of arctan() as atan2() and I'll leave asin() as well for consistency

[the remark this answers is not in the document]
etfrogers-dstl: Happy with that. Might be worth formatting as `\textrm{asin}` and `\textrm{atan2}` in any case.

stonesoup/models/measurement/nonlinear.py
jjosborne-dstl: It's important for the PDF. this is because "state2" is passed into self.function with noise = false. Then that's used as the mean in multivariance.norm.cdf (pdf for other models). If it was bin it would move state2 which would move the center of the normal distribution.

stonesoup/models/measurement/nonlinear.py
jjosborne-dstl: I'm trying to check if the measured range and range rate is in one of the bins by seeing if it is a multiple of the bin size. let me know if there's a better way to do that.

jjosborne-dstl: I have spotted an error. It's looking at the 2 angles. I'll change it to the range and range rate.

stonesoup/models/measurement/nonlinear.py
etfrogers-dstl: Would the check be better as `if not noise == False` then? Why check for if it is None? Again, there might be a good answer. 

I would also add the justification for this to either the docstring or a comment.

stonesoup/models/measurement/nonlinear.py
etfrogers-dstl: So the check is: is state1.state_vector at a bin edge in both range and range rate. If not, then this is an impossibility and so the probability of it occurring is 0. Is that the logic?

stonesoup/models/measurement/nonlinear.py
jjosborne-dstl: I'm not sure why it's done the way it is but it's what the other measurement models use. I'll ask Rich

stonesoup/models/measurement/nonlinear.py
jjosborne-dstl: yes. I might be wrong but my understanding is pdf returns the likeliness of state1 if self.function was given state2. Does it seem right to you?

stonesoup/models/measurement/nonlinear.py
sdhiscocks: If I remember rightly, this is because combination of `0 == False`, and `if not np.array([[1], [2], [3]]) == False:` would raise _ValueError_.

etfrogers-dstl: Would `if noise is False` work? Seems to salve those two issues.

stonesoup/models/measurement/nonlinear.py
etfrogers-dstl: I think you are right that the logic is correct then, but I had some issues with calculations when trying to work with delta-function-like PDFs. It means the area under the PDF is quite limited and, depending on how the function is used, it might never check that *specific* value. I think we would have a lot more confidence if there was a test/example that worked here.

stonesoup/models/measurement/nonlinear.py
oharrald-Dstl: To save a line, could you do: 'if isinstance(noise, bool) and noise:' ?

stonesoup/models/measurement/nonlinear.py
```diff
@@ -965,3 +967,192 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+            0 & \sigma_{\phi}^2 & 0 & 0\\
+            0 & 0 & \sigma_{r}^2 & 0\\
+            0 & 0 & 0 & \sigma_{\dot{r}}^2
+            \end{bmatrix}
+
+    The range and range rate are binned using
+
+    .. math::
+
+        x = floor(x/\Delta x)*\Delta x
```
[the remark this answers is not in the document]
DaveKirkland: The binning output gives the value at the starting edge of the bin; e.g. x=0, del_x = 10, then x_bin =0. Are there times when it makes sense to return the value in the middle of the bin?  Additional thought - The range gating of the radar may not align with starting from zero. If the radar starts binning the range at 2025.0 meters with delta range of 10, then the binning won't line up actual radar. @sdhiscocks Perhaps we want some binning functions that can be used here as well as in the radar sim classes?

stonesoup/models/measurement/nonlinear.py
```diff
@@ -965,3 +967,192 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+
+    The :py:attr:`velocity_mapping` property of the model is a 3 element vector, \
+    whose first (i.e. :py:attr:`velocity_mapping[0]`), second (i.e. \
+    :py:attr:`velocity_mapping[1]`) and third (i.e. :py:attr:`velocity_mapping[2]`) elements \
+    contain the state index of the :math:`\dot{x}`, :math:`\dot{y}` and :math:`\dot{z}`  \
+    coordinates, respectively.
+
+    Note
+    ----
+    This class implementation assuming at 3D cartesian space, it therefore \
```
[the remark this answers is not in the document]
DaveKirkland: Change to: This class implementation assumes a 3D cartesian space, it therefore expects a 6D state space.
Can't this also be used with an acceleration model?  Would it work properly? Not sure if this just a documentation issue or there are issues involved with using a constant acceleration model (or other models e.g. mixtures etc).

stonesoup/models/measurement/nonlinear.py
DaveKirkland: Just minor issue with the grammar on the documentation:
The covariances for radar **are** determined by different factors. The angle error
    is **affected** by the radar beam width. Range error is **affected** by the SNR and pulse bandwidth.
    The error for the range rate **is dependent** on the dwell time.
    The range and range rate are binned using

stonesoup/models/measurement/nonlinear.py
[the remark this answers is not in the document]
Abell1436: It could uses a acceleration model and the mapping and velocity mapping gets the position and velocity from the state.

stonesoup/models/measurement/nonlinear.py
Abell1436: Thank you for checking the grammar

stonesoup/models/measurement/nonlinear.py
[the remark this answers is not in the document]
jjosborne-dstl: An acceleration model can be used. The mapping and velocity mapping gets the position and velocity from the state so the state can have more than 6 dimensions.

stonesoup/models/measurement/nonlinear.py
jjosborne-dstl: Thank you for checking the grammar.

stonesoup/models/measurement/nonlinear.py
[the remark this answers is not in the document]
DaveKirkland: So I think the documentation just needs to be modified - right now it explicitly states a 6D state model. I think this part of the documentation is being grabbed from the parent class.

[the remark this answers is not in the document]
jjosborne-dstl: Should the documentation instead say it accepts a state with a minimum 6D? Should the other measurement model also be changed too because most measurement model have a minimum number of dimensions needed in a state but no maximum?

etfrogers-dstl (approved):
Loos good. Think you've addressed all the issues.

gawebb-dstl (commented):
All looks good, I've added a couple of comments but that just might be lack of understanding on my part.

stonesoup/models/measurement/nonlinear.py
```diff
@@ -962,3 +963,194 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+    contain the state index of the :math:`\dot{x}`, :math:`\dot{y}` and :math:`\dot{z}`  \
+    coordinates, respectively.
+
+    Note
+    ----
+    This class implementation assumes a 3D cartesian space, it therefore \
+    expects a 6D state space.
+    """
+
+    range_res = Property(float, doc="Size of the range bins in m/s")
```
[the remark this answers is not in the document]
jjosborne-dstl: Yes they are. Thank you for spotting it.

gawebb-dstl (approved):

PR #247, opened 2020-06-23
UAV Tutorial_V2

DaveKirkland:
Updated UAV Demo. Addressed previous comments.


docs/demos/UAV_tutorial.py
```diff
@@ -0,0 +1,279 @@
...
+
+# %%
+# Setup Deletor for the Tracker
+# -----------------------------
+# In the simple case of 1 target, we never want to delete the track. Because
+# this Deletor is so simple I haven't bothered using a subtype/inheritance.
+# Instead I make use of Python's duck typing. 
+class MyDeleter:
+    def delete_tracks(self, tracks):
+        return {}
```
sdhiscocks, line 191: `{}` is empty dictionary. Should be `set()`. Still works however, as set operation will use dictionary keys (which is an empty set :smile:). But to avoid confusion should change to `return set()`.

DaveKirkland: Done.

docs/demos/UAV_tutorial.py
```diff
@@ -0,0 +1,279 @@
...
+detector = PlatformDetectionSimulator(ground_truth_reader, [platform])
+
+
+# %%
+# Setup Initiator class for the Tracker
+# ---------------------------------------
+# This is just an heuristic initiation:
+# Assume most of the deviation is caused by the Bearing measurement error.
+# This is then converted into x, y components using the target bearing. For z,
+# we simply use range*elev_std_dev (ignore any bearing or range related components).
```
jmbarr, line 131: :math:`r \times \lamba_{\sigma}` or something? Make it look fancier...

docs/demos/UAV_tutorial.py
```diff
@@ -0,0 +1,279 @@
...
+            
+   
+prior_state = GaussianState(
+        np.array([[0], [0], [0], [0], [0], [0]]),
+        np.diag([0, 30.0, 0, 30.0, 0, 30.0])**2)
+initiator = Initiator(prior_state, meas_model)
+#initiator = SimpleMeasurementInitiator(prior_state, meas_model)
+
+# %%
+# Setup Deletor for the Tracker
```
jmbarr, line 184: Deleter

docs/demos/UAV_tutorial.py
```diff
@@ -0,0 +1,279 @@
...
+# - Add clutter.
+# - Handle :math:`P_d` behaviour. Could be based on radial velocity or other
+#   parameters.
+# - Handle radar revisit times.
+# - Add unknown number & multiple targets.
+#
+from stonesoup.reader.generic import CSVGroundTruthReader
+
+ground_truth_reader = CSVGroundTruthReader(
+    path='UAV_Rot.csv',
```
sdhiscocks, line 89: Need to provide a download link to the data. You can see example here using the `:download:` directive: https://github.com/dstl/Stone-Soup/blob/b6df2598989b1d03e27a6ab47e64897ba566b6d2/docs/demos/AIS_Solent_Tracker.py#L20-L31

Path should be the same as AIS demo CSV, so just need to change filename to `UAV_Rot.csv`. Sphinx will keep sure this is packaged up with documentation then.

docs/demos/UAV_tutorial.py
```diff
@@ -0,0 +1,279 @@
...
+ax1.set_xlabel('X (m)')
+ax1.set_ylabel('Y (m)')
+
+fig = plt.figure(figsize=(10, 6))
+ax2 = fig.add_subplot(1, 1, 1)
+ax2.plot(true_X, true_Y, 'd-k', label='Truth', markerfacecolor='None')
+ax2.plot(est_X, est_Y, 'r.', label='Estimates')
+ax2.set_xlabel('X (m)')
+ax2.set_ylabel('Y (m)')
+ax2.legend()
```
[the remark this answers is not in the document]
DaveKirkland: I tried overlaying the measurements and the estimates, but the graph is too busy. The measurements are very noisy and because the track comes back on itself, you're likely to infer the wrong measurement with the estimate. This works better when you only show a few measurements at a time. Using an animated sequence would be better, but that also brings other difficulties with it. Similar problem with the uncertainties - it's just too busy with so many measurements.

DaveKirkland:
Added download link and addressed other editorial issues.

PR #249, opened 2020-06-24
Fix Sphinx-Gallery scraper animations

sdhiscocks:
Apply upstream changes to _matplotlib_ scraper to enable animation support.

Longer term would be good to work with Sphinx-Gallery project so we don't need this custom scraper.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -65,6 +65,7 @@ sphinx_gallery_conf = {
      'reference_url': {'stonesoup': None},
      'remove_config_comments': True,
      'ignore_repr_types': r'matplotlib.figure',
+     'matplotlib_animations': True,
 }
 
 intersphinx_mapping = {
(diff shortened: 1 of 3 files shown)


sglvladi:
@sdhiscocks, I pulled this branch and copied the script from the [the Sphinx Gallery animation example](https://sphinx-gallery.github.io/stable/auto_examples/plot_8_animations.html) inside "docs/source/auto_tutorials", then re-run sphinx, but I am still getting no animation 🤕 

Just for reference, this is what I see:
ANIMATIONS

Am I missing something?

sdhiscocks:
> Am I missing something?

Did you set `matplotlib_animations: True,` in `doc/conf.py`?

sglvladi:
> Did you set `matplotlib_animations: True,` in `docs/source/conf.py`?

Ahaa! No I hadn't. After doing that, running ``make clean`` and deleting the ``source/auto_tutorial`` folder I got a happy animation! 

image

The radio buttons seem to be oddly positioned vertically, but I fixed that by adding the following css to ``docs/source/_static/css/custom.css``:

``` css
.anim-state label {
    display: inline-block;
}
```

PR #250, opened 2020-06-24
Tweaks to the single target tutorials. 

jmbarr:
Added some key learning points at the end, and the ability to set a random seed.

### DIFF
--- a/docs/tutorials/01_KalmanFilterTutorial.py
+++ b/docs/tutorials/01_KalmanFilterTutorial.py
@@ -340,6 +340,8 @@ for state in track:
     ax.add_artist(ellipse)
 fig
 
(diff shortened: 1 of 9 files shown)


docs/tutorials/01_KalmanFilterTutorial.py
```diff
@@ -102,6 +103,13 @@
 # And the clock starts
 start_time = datetime.now()
 
+# %%
+# We note that it can sometimes be useful to fix our random number generator in order to probe a
+# particular example repeatedly. That option is available by uncommenting the next lines.
+
+# import random
+# random.seed(1991)
```
sdhiscocks, line 111: Just tested this and it's doesn't work. Need to use NumPy random seed. I'll change it and merge this PR.

PR #184, opened 2020-04-04
Add TensorFlow detector

sglvladi:
Summary
----------

This PR adds the following:
- The `TensorFlowBoxObjectDetector` class, an object detector that generates detections of objects, in the form of bounding boxes, from image/video frames using a [TensorFlow object detection model](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/detection_model_zoo.md).
- A minor bugfix in the `VideoClipReader` class, which ensures that init parameter defaults are setup and processed correctly.

An example script can be found under `examples/tensorflow`. I have also included a sample video file. Both files will be removed before the PR is finilised for merging.

Note
-----
TensorFlow must be installed to run the example. Installation instructions for TensorFlow can be found [here](https://tensorflow2objectdetectioninstallation.readthedocs.io/en/latest/).

TODO
------- 
- [x] Remove examples
- [x] Add tests
- [x] Decide how to add TensorFlow to the dependencies.

PR #251, opened 2020-06-24
WIP: Video processing demo

sglvladi:
This PR adds the Video Processing demonstration for Fusion 2020.

Note
------
The branch is based on #184 and therefore includes a lot of changes from there. Only the file [Video_Processing.py](https://github.com/dstl/Stone-Soup/pull/251/files#diff-a6efc3f85f322a78ff3700817b85c8ff) is added here.


sdhiscocks:
Something that I haven't tried, is with the more complex demos, we may want to pre-compute the content, as may be too intensive to run on CircleCI and Read-The-Docs, or setup/dependencies may be complex, or may require 3rd party content we can't guarantee (API calls, etc.).

I've not tried running this, but I'm assuming that NN will be slow on those platforms?

I'm hoping that if you also commit the following after running on your system:
 - `docs/source/auto_demos/Video_Processing.ipynb`
 - `docs/source/auto_demos/Video_Processing.py.md5`
 - `docs/source/auto_demos/Video_Processing.rst`
 - `docs/source/auto_demos/images/sphx_glr_Video_Processing*`
 - `docs/source/auto_demos/images/thumb/sphx_glr_Video_Processing_thumb*`

that the page will be rendered okay.

sglvladi:
@sdhiscocks OK, I've committed the above files. Not sure, but I think CI is still trying to run the file, seeing as it fails we importing the ``VideoClipReader``.

sdhiscocks:
> @sdhiscocks OK, I've committed the above files. Not sure, but I think CI is still trying to run the file, seeing as it fails when importing the `VideoClipReader`.

Hope you don't mind @sglvladi, but thought I'd trying pushing something to see if it worked. I realised that md5sum will depend on line endings, and CircleCI is Linux based. So I converted the file to Unix line endings and calculated md5sum, and pushing that seems to have worked 👍 

Way I generated the md5sum file using Git Bash:
```bash
echo -n $(cat docs/demos/Video_Processing.py | dos2unix | md5sum | cut -d\  -f1) > docs/source/auto_demos/Video_Processing.py.md5   
```

docs/demos/Video_Processing.py
```diff
@@ -0,0 +1,607 @@
...
+# For confirmed tracks we used again a :class:`~.UpdateTimeStepsDeleter`, but this time configured
+# to delete tracks after they have not bee associated to a measurement in the last 15 frames.
+deleter = UpdateTimeStepsDeleter(time_steps_since_update=15)
+#%%
+# .. note::
+#
+#   For more information on the above classes and how they operate you can refer to the Stone
+#   `Initiators & Deleters <https://stonesoup.readthedocs.io/en/latest/auto_tutorials/09_Initiators_&_Deleters.html>`_
+#   tutorial.
+#
```
sdhiscocks, line 495: Think this # needs to be blank line, as causes heading below to break.

sglvladi: @sdhiscocks Thank you Steve, I have fixed that in the latest version.

sdhiscocks:
Not sure about the RST, as it's 58MB. May need to think about better approach.

sdhiscocks:
> Not sure about the RST, as it's 58MB. May need to think about better approach.

I replaced the animated sections (which had to store raw image for every frame), with the GIFs. Not quite as good as lack animation controls, but more reasonable size file.

PR #252, opened 2020-06-25
Single sensor platform example

rjgreen-dstl:
Documents example of a single sensor platform being deployed within a simulator

### DIFF
--- a/docs/examples/Sensor_Platform_Simulation.py
+++ b/docs/examples/Sensor_Platform_Simulation.py
@@ -215,7 +215,7 @@ groundtruth_sim = MultiTargetGroundTruthSimulator(
     initial_state=initial_target_state,  # add our initial state for targets
     timestep=timedelta(seconds=1),  # time between measurements
     number_steps=120,  # 2 minute
-    birth_rate=0.10,  # 5% chance of a new target being birthed
+    birth_rate=0.10,  # 10% chance of a new target being birthed
     death_probability=0.01  # 1% chance of a target being killed
 )
 
--- a/docs/examples/Sensor_Platform_Simulation.py
+++ b/docs/examples/Sensor_Platform_Simulation.py
@@ -1,0 +1,381 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+Sensor Platform Simulation Example
+==================================
+This example looks at how platforms and sensors can be used within the Stone Soup simulation capability.
+"""
+
+# %%
+# Building a Simulated Sensor Platform
+# ------------------------------------
+# This example is going to look at how *Platforms* and *Sensors* can be combined with the simulation capabilites within
+# Stone Soup.
+#
+# The focus of this example is to show how to setup and configure simulations, as such the application of a tracker
+# will not be covered in detail. For more information about trackers and how to configure them review of the
+# tutorials and demonstrations is recommended.
+#
+# This example makes use of Stone Soup :class:`~.FixedPlatform` and :class:`~.Sensor` objects.
+#
+# In order to configure platforms, sensors and the simulation we will need to import some specific Stone Soup objects.
+# As these have been introduced in previous tutorials they are imported upfront. New functionality within this example
+# will be imported at the relevant point in order to draw attention to the new features.
+
+# Some general imports and set up
+from datetime import datetime
+from datetime import timedelta
+
+import numpy as np
+
+# Stone Soup imports:
+from stonesoup.types.state import State, GaussianState
+from stonesoup.types.array import StateVector, CovarianceMatrix
+from stonesoup.models.transition.linear import (
+    CombinedLinearGaussianTransitionModel, ConstantVelocity)
+from stonesoup.models.measurement.nonlinear import CartesianToElevationBearingRange
+from stonesoup.updater.kalman import UnscentedKalmanUpdater
+from stonesoup.predictor.kalman import UnscentedKalmanPredictor
+from stonesoup.deleter.time import UpdateTimeStepsDeleter
+from stonesoup.tracker.simple import MultiTargetTracker
+from matplotlib import pyplot as plt
+
+# Define the simulation start time
+start_time = datetime.now()
+
+# %%
+# Create a Platform
+# -----------------
+# The first element we need to create is a platform. For this first example we will build a static (or *fixed*) platform
+# which is located at the origin. For this example we are going to work in a 6-dimensional state space our platform will
+# have the following :math:`\vec{x}`.
+#
+# .. math::
+#           \begin{align}\vec{x} = \begin{bmatrix}
+#                                   x\\ \dot{x}\\ y\\ \dot{y}\\ z\\ \dot{z} \end{bmatrix}
+#                                = \begin{bmatrix}
+#                                   0\\ 0\\ 0\\ 0\\ 0\\ 0 \end{bmatrix}\end{align}
+#
+# Because the platform is static we only need to define :math:`(x, y, z)`, any internal interaction with the platform
+# which requires knowledge of platform velocity :math:`(\dot{x}, \dot{y}, \dot{z})` will be returned :math:`(0, 0, 0)`.
+
+# First import the fixed platform
+from stonesoup.platform.base import FixedPlatform
+
+# Define the initial platform position, in this case the origin
+platform_state_vector = StateVector([[ 0], [0], [0]])
+position_mapping = (0, 1, 2)
+
+# Create the initial state (position, time), notice it is set to the simulation start time defined earlier
+platform_state = State(platform_state_vector, start_time)
+
+# create our fixed platform
+platform = FixedPlatform(states=platform_state,
+                         position_mapping=position_mapping)
+
+# %%
+# We have now created a platform within Stone Soup and located it at the origin of our state space. As previously stated
+# the platform will have a velocity :math:`(\dot{x}, \dot{y}, \dot{z})` of :math:`(0, 0, 0)` which we can check:
+platform.velocity
+
+# %%
+# We can also query the platform orientation:
+platform.orientation
+
+# %%
+# Create a Sensor
+# ---------------
+# Now that we have a platform the next step is to create a sensor which can be added to the platform. In this example a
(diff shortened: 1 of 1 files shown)


docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,381 @@
...
+platform.orientation
+
+# %%
+# Create a Sensor
+# ---------------
+# Now that we have a platform the next step is to create a sensor which can be added to the platform. In this example a
+# Radar will be created which is capable of measuring the range, bearing and elevation of the target relative to the
+# sensor.
+#
+# The :class:`~.RadarElevationBearingRange`  provides a sensor wrapper around the
```
sdhiscocks, line 93: Wrong class name here.

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,381 @@
...
+#   * :math:`x` is Gaussian distributed around the platform location with variance of 2km
+#   * :math:`y` is Gaussian distributed around the platform location with variance of 2km
+#   * :math:`z` is Gaussian distributed around an altitude of 9km with variance of 0.1km
+#   * :math:`\dot{x}` is Gaussian distributed around 100 :math:`ms^{-1}` with variance of 50 :math:`ms^{-1}`
+#   * :math:`\dot{y}` is Gaussian distributed around 100 :math:`ms^{-1}` with variance of 50 :math:`ms^{-1}`
+#   * :math:`\dot{z}` is Gaussian distributed around 0 :math:`ms^{-1}` with variance of 1 :math:`ms^{-1}`
+#
+# We will also configure our simulator to randomly create and delete targets, based on a birth rate and death rate we
+# specify. In this example we set the birth rate to be 30%, i.e. on any given time step there is a 30% chance of a new
+# target being initiated. We have set the death rate to 5%, i.e. on any given time step that is the chance that a target
```
sdhiscocks, line 202: Doesn't match numbers below.

docs/examples/Sensor_Platform_Simulation.py
rjgreen-dstl: Hangover from separation of this from proposed naming convention changes to radar, good spot

docs/examples/Sensor_Platform_Simulation.py
rjgreen-dstl: done

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+"""
+Sensor Platform Simulation Example
+==================================
+This example looks at how platforms and sensors can be used within the Stone Soup simulation capability.
+"""
+
+# %%
+# Building a Simulated Sensor Platform
+# ------------------------------------
+# This example is going to look at how *Platforms* and *Sensors* can be combined with the simulation capabilites within
```
jmbarr, line 13: Repeats line above almost verbatim. Suggest remove.

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+
+# Define the simulation start time
+start_time = datetime.now()
+
+# %%
+# Create a Platform
+# -----------------
+# The first element we need to create is a platform. For this first example we will build a static (or *fixed*) platform
+# which is located at the origin. For this example we are going to work in a 6-dimensional state space our platform will
+# have the following :math:`\vec{x}`.
```
jmbarr, line 52: For consistency with tutorials and to give the impression we talk with one voice we should use :math:`\mathbf{x}` to describe the state.

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+#
+# The :class:`~.RadarRangeBearingElevation`  provides a sensor wrapper around the
+# :class:`~.CartesianToElevationBearingRange` measurement model. The measurement model provides a time-invariant
+# measurement model, where measurements are assumed to be received in the form of elevation (:math:`\theta`),
+# bearing (:math:`\phi`) and range (:math:`r`) with Gaussian noise in each dimension.
+#
+# The model is described by the following equations:
+#
+# .. math::
+#           \begin{align} \vec{y}_t = h(\vec{x}_t, \vec{v}_t) \end{align}
```
jmbarr, line 101: use :math:`\mathbf{z}` for observations/measurements

jmbarr: also suggest :math:`\dot{\mathbf{x}}` instead of v

jmbarr: and $k$ instead of $t$

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+# where :math:`\vec{y}_t` is a measurement vector of the form:
+#
+# .. math::
+#           \begin{align} \vec{y}_t = \begin{bmatrix} \theta \\ \phi \\ r \end{bmatrix} \end{align}
+#
+# and :math:`h` is a non-linear model function of the form:
+#
+# .. math::
+#           \begin{align} h(\vec{x}_t,\vec{v}_t) = \begin{bmatrix}
+#                                               asin(\mathcal{z}/\sqrt{\mathcal{x}^2 + \mathcal{y}^2 +\mathcal{z}^2}) \\
```
jmbarr, line 112: I think you can use \arcsin and \arctan (the latter is the same as atan2 except without solving the quadrant ambiguity - but that point usually gets made in the text because I think atan2 was invented by programmers!)

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+# Building a simulation
+# ---------------------
+# Now that we have created a sensor platform we need to build a simulation which generates targets for the sensor to
+# detect and track. For this example we are going to use a :class:`~.MultiTargetGroundTruthSimulator`. this simulator
+# enables multiple ground truth targets to be created based on a number of user defined parameters.
+#
+# In this example targets are initiated with values based upon a mean state and a covariance, using a Gaussian
+# assumption. This is done by creating a :class:`~.GaussianState` object which describes the distribution from which we
+# want our targets to be drawn from. For this example targets will be drawn from:
+#   * :math:`x` is Gaussian distributed around the platform location with variance of 2km
```
jmbarr, line 193: This could probably be written more compactly as :math:`\mathcal{N}(\mathbf{x}_p, \diag([2,2,0.1,50,50,1])` (or similar)

Units should be roman, not italic (\mathrm{})

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+# :class:`~.Simulator` work within Stone Soup. We will therefore quickly build an Unscented Kalman Filter which
+# initiates measurements using a simple heuristic initiation and deletes any track where no detection is associated for
+# 2 consecutive time steps. There are a number of tutorials for how to build the tracking components provided in the
+# :ref:`auto_tutorials`.
+
+# Create an Unscented Kalman Predictor
+predictor = UnscentedKalmanPredictor(transition_model)
+
+# Create an Unscented Kalman Updater, note our sensor adds a measurement model to detections
+updater = UnscentedKalmanUpdater(measurement_model=None)
```
jmbarr, line 252: It's probably worth pausing to mention that we can do measurement_model=None because I don't think that will have come up before

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+
+# %%
+# Setup Initiator class for the Tracker
+# -------------------------------------
+# We will now build a simple heuristic initiator.
+# This assumes most of the deviation is caused by the Bearing measurement error. It converts the bearing error error
```
jmbarr, line 258: bearing rather than Bearing

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+# into x,y components using the target bearing. For z, we simply use range*elev_std_dev (this ignores any bearing or
```
jmbarr, line 259: use math mode for range*elev_std_dev

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+            groundtruth_paths[truth] = []
+        groundtruth_paths[truth].append(loc)
+
+    for detection in sim.detections:
+        detect_state = detection.measurement_model.inverse_function(detection)
+        loc = (detect_state[0], detect_state[2])
+        detections.append(loc)
+
+# %%
+# Plotting the tracker output we get:
```
jmbarr, line 360: Give us a key or describe what's what in the plot. I can't tell which line is which.

docs/examples/Sensor_Platform_Simulation.py
```diff
@@ -0,0 +1,380 @@
...
+    ax.plot(X, Y, color='r')
+
+for key in kalman_tracks:
+    X = [coord[0] for coord in kalman_tracks[key]]
+    Y = [coord[1] for coord in kalman_tracks[key]]
+    ax.plot(X, Y, color='y')
+
+X = [coord[0] for coord in detections]
+Y = [coord[1] for coord in detections]
+ax.scatter(X, Y, color='k')
```
jmbarr, line 380: Can we add a line or two at the end telling everyone what they've learnt?

docs/examples/Sensor_Platform_Simulation.py
rjgreen-dstl: agreed, will add some text too alongside a dramatic pause

docs/examples/Sensor_Platform_Simulation.py
rjgreen-dstl: done

docs/examples/Sensor_Platform_Simulation.py
rjgreen-dstl: Tried the compact solution and it looks messy so reverted to original structure

Issue #262, opened 2020-06-30
Deep copy of a platform produces errors 


gawebb-dstl:
There is problem with platforms/sensors are copied. I believe it's related to the weak reference between the sensor and platform. It may be related to this [stackoverflow post](https://stackoverflow.com/questions/46203335/creating-a-deepcopy-of-class-instance-with-nested-weakref-to-it). Example code which shows errors is given below:

```python
# Import external dependencies
import datetime
import copy
import numpy as np
import traceback

from stonesoup.types.state import State
from stonesoup.types.array import StateVector, CovarianceMatrix
from stonesoup.models.transition.linear import (
    ConstantVelocity,
    CombinedLinearGaussianTransitionModel
)
from stonesoup.platform.base import MovingPlatform
from stonesoup.sensor.radar.radar import RadarRangeRateBearingElevation

# Set up Radar Model
transition_model = CombinedLinearGaussianTransitionModel(
    (ConstantVelocity(0), ConstantVelocity(0), ConstantVelocity(0)))

pos_mapping = np.array([0, 2, 4])
vel_mapping = np.array([1, 3, 5])
noise_covar = CovarianceMatrix(np.eye(4))

radar1 = RadarRangeRateBearingElevation(ndim_state=6,
                                        position_mapping=pos_mapping,
                                        velocity_mapping=vel_mapping,
                                        noise_covar=noise_covar)

# create a platform with the simple radar mounted
timestamp_init = datetime.datetime.now()
platform_1_prior_state_vector = StateVector([[5], [0], [0], [0.25], [0], [0]])
platform_1_state = State(platform_1_prior_state_vector, timestamp_init)

mounting_offsets = [StateVector([[0], [0], [0]])]
rotation_offsets = [StateVector([[0], [0], [0]])]

transition_model = CombinedLinearGaussianTransitionModel(
    (ConstantVelocity(0), ConstantVelocity(0), ConstantVelocity(0)))

platform_1 = MovingPlatform(states=platform_1_state,
                            position_mapping=pos_mapping,
                            velocity_mapping=vel_mapping,
                            transition_model=transition_model,
                            sensors=[radar1],
                            mounting_offsets=mounting_offsets,
                            rotation_offsets=rotation_offsets)

target_state = State(StateVector([[0], [0], [0], [0], [0], [0]]), timestamp_init)

# Make a measurement with a platform
measurement_pol_1 = platform_1.sensors[0].measure(target_state)

[...]

PR #263, opened 2020-07-03
Fix issues with copying platforms

etfrogers-dstl:
This commit makes two changes to address the issues raised by @gawebb-dstl in #262 :

1. Modify the `sensors` Property of a `Platform` to make it read-only (including making it effectively immutable)
2. Add a `__deepcopy__` method to the `Base` class to allow deep copying of a platform. This needs some thought as it may have unintended consequences. I can't see that it will, but I'd appreciate more eye on this. 

Thanks to @sdhiscocks for the idea of a getter returning a tuple and @gawebb-dstl for the test scenarios!

### DIFF
--- a/stonesoup/base.py
+++ b/stonesoup/base.py
@@ -56,8 +56,10 @@ This is equivalent to the following:
 """
 import inspect
 import sys
+import weakref
 from abc import ABCMeta
 from collections import OrderedDict
+from copy import deepcopy
 from types import MappingProxyType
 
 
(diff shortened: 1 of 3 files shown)


stonesoup/base.py
```diff
@@ -305,3 +307,22 @@ def __repr__(self):
...
+        for name, prop in new.__dict__.items():
+            if isinstance(prop, weakref.ref):
+                original_target = prop()  # call the weakref to get a reference
+                try:
+                    # if we are copying the parent as well, the copy should be in memodict
+                    copy_of_target = memodict[id(original_target)]
+                    new.__setattr__(name, weakref.ref(copy_of_target))
+                except KeyError:
+                    # if we can't find the parent, then leave the original ref in place
+                    pass
```
sdhiscocks, line 327: Minor suggestion.

```suggestion
                try:
                    # if we are copying the parent as well, the copy should be in memodict
                    copy_of_target = memodict[id(original_target)]
                except KeyError:
                    # if we can't find the parent, then leave the original ref in place
                    pass
                else:
                    new.__setattr__(name, weakref.ref(copy_of_target))
```

etfrogers-dstl: Not sure that will work. I don't think `copy_of_target` will be set in the `KeyError` case.

sdhiscocks: `else` block is only executed if there isn't an error. This is useful to avoid unintentionally catching another `KeyError` when calling `__setattr__`

etfrogers-dstl: Sorry, was mistaking for `finally`. Haven't seen that usage. You learn something every day!

P.S. This is why everyone should both have and do code reviews :)

PR #264, opened 2020-07-03
Add Pytube to dependencies

etfrogers-dstl:
@sdhiscocks ~I think this also needs adding to setup.py?~

Does this need to go anywhere else?

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ These dependencies are required for building Stone Soup documentation.
 | [sphinx-gallery](https://sphinx-gallery.github.io/) | BSD |
 | [pillow](https://pillow.readthedocs.io/en/stable/index.html) | [PIL Software License](https://pillow.readthedocs.io/en/stable/about.html#license) |
 | [folium](https://python-visualization.github.io/folium/) | MIT |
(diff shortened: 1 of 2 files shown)


sdhiscocks:
The video example requires a fair amount of computing power, and also not really suitable for CircleCI and Read The Docs to be executing due to this, as well as requiring 500MB of model weights being downloaded every time. So decision was made in #251 to add it as a pre-computed example. Therefore, it's not required to build the documentation.

However, there is an issue on Windows and Mac at the moment due to sphinx-gallery checksum's being platform dependent. This should be solved if/when https://github.com/sphinx-gallery/sphinx-gallery/pull/722 is merged.

etfrogers-dstl:
Oops. OK. Thanks for explaining. As I think you know, it still causes my Sphinx build to fail.

sdhiscocks:
@etfrogers-dstl  Work around for now is to restore any modified files in `docs/auto_demos/` and run the following if your using Git Bash on Windows:

```bash
echo -n $(cat docs/demos/Video_Processing.py | md5sum | cut -d\  -f1) > docs/source/auto_demos/Video_Processing.py.md5
```

etfrogers-dstl:
Works. Thanks Steve!

PR #266, opened 2020-07-06
Fix initiator covariance issue.

DaveKirkland:
Fixed the covariance calculation in the SimpleMeasurementInitiator.  A diag_load parameter was added to the initiator (default =0) which adds a diagonal load value to the resulting covariance matrix. This was due to the Cholesky factorization complaining that the matrix was not positive definite.

The test code was also updated and corrected to test the returned covariance matrix i.e. some additional assert are tested.

I tested this with the UAV tutorial and the Sensor_Platform_Simulation example, which were having problems using the SimpleMeasurementInitiator. They both work now, however the Sensor_Platform_Simulation requires a small diag_load value to be supplied i.e. 1e-16.

### DIFF
--- a/stonesoup/initiator/simple.py
+++ b/stonesoup/initiator/simple.py
@@ -70,13 +70,23 @@ class SimpleMeasurementInitiator(GaussianInitiator):
 
     This then replaces mapped values in the :attr:`prior_state` to form the
     initial :class:`~.GaussianState` of the :class:`~.Track`.
+
+    The diagonal loading value is used to try to ensure that the estimated
+    covariance matrix is positive definite, especially for subsequent Cholesky
+    decompositions.
     """
     prior_state = Property(GaussianState, doc="Prior state information")
     measurement_model = Property(MeasurementModel, doc="Measurement model")
     skip_non_reversible = Property(bool, default=False)
     diag_load = Property(float,
                          default=0.0,
-                         doc="Float value for diagonal loading")
+                         doc="Positive float value for diagonal loading")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.diag_load < 0:
+            raise ValueError(
+                "diag_load value can't be less than 0.0")
 
     def initiate(self, detections, **kwargs):
         tracks = set()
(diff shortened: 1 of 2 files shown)


stonesoup/initiator/simple.py
```diff
@@ -74,6 +74,9 @@ class SimpleMeasurementInitiator(GaussianInitiator):
     prior_state = Property(GaussianState, doc="Prior state information")
     measurement_model = Property(MeasurementModel, doc="Measurement model")
     skip_non_reversible = Property(bool, default=False)
+    diag_load = Property(float,
```
sdhiscocks, line 77: I think a short sentence is main doc string above on why you'd want a diagonal load be useful.

stonesoup/initiator/simple.py
```diff
@@ -110,11 +113,12 @@ def initiate(self, detections, **kwargs):
                 model_matrix.T @ np.ones((model_matrix.shape[0], 1)))
             prior_state_vector[mapped_dimensions, :] = 0
             prior_covar[mapped_dimensions, :] = 0
-
+            C0 = inv_model_matrix @ model_covar @ inv_model_matrix.T
```
sdhiscocks, line 116: The last term has changed from `model_matrix.astype(bool)` to `inv_model_matrix.T`. Is that right? I can see it's changed the result in [the test below](https://github.com/dstl/Stone-Soup/pull/266/files#diff-c0310c8dfe4bac1ad2388ecabd89e3f7R182)

DaveKirkland: Yes. This is the main change that was causing the trackers to fail. If you look at the test, I added in assertions that check that the state covariance matrix is consistent with the measurement noise. 
If we assume that R<sub>y</sub> = H R<sub>x</sub> H<sup>T</sup> then R<sub>x</sub> is approximated by  G R<sub>y</sub> G<sup>T</sup>, where G is the psuedo-inverse of H.

stonesoup/initiator/simple.py
DaveKirkland: Should this be in the docstring of the diag_load Property or the docstring of the class?

sdhiscocks: For the class I think. Let's you be more verbose.

DaveKirkland: Done. Note - I added a check to make sure that diag_load >= 0 in the __init__().

PR #269, opened 2020-07-07
Change 2D sensor/platform orientations to 3D shape

oharrald-Dstl:
Returning a length 2 vector sometimes led to casting errors with platform orientations and sensor rotation offsets.

Note: May be worth allowing an input of length 1 (user inputs single bearing for 2D sensor rotation offset)

### DIFF
--- a/stonesoup/platform/base.py
+++ b/stonesoup/platform/base.py
@@ -353,8 +353,8 @@ class MovingPlatform(Platform):
             _, bearing = cart2pol(*velocity.flat)
             return StateVector([0, 0, bearing])
         else:
-            raise ValueError('Orientation of a moving platform is only implemented for 2 and 3 '
-                             'dimensions')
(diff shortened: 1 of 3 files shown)


stonesoup/platform/base.py
```diff
@@ -342,15 +342,16 @@ def orientation(self) -> StateVector:
...
 
         if self.ndim == 3:
             _, bearing, elevation = cart2sphere(*velocity.flat)
             return StateVector([0, bearing, elevation])
         elif self.ndim == 2:
             _, bearing = cart2pol(*velocity.flat)
-            return StateVector([0, bearing])
+            return StateVector([0, 0, bearing])
         else:
             raise ValueError('Orientation of a moving platform is only implemented for 2 and 3 '
```
sdhiscocks, line 356: I think it would make sense for this to raise an `NotImplementedError` as well.

PR #271, opened 2020-07-08
Multi model

Isaac-JenkinsRA:
Hello there,

Sorry for the long delay on getting this pull request sorted. Here we have the multi-model and Raoblackwellised versions of the particle filter.

Additionally I have an example script which I can email directly if you would like, but it is not appropriate for this repo.

Many thanks,

Isaac - RiskAware


stonesoup/updater/particle.py
```diff
@@ -70,7 +70,236 @@ def predict_measurement(self, state_prediction, measurement_model=None,
...
+                # Input the indices that were removed previously
+                for j in range(len(particle.state_vector)):
+                    if j not in position_mapping[i]:
+                        mean = np.insert(mean, j, 0)
+                mean = mean.reshape((1, -1))[0]
+
+                prob_position_given_model_and_old_position = multivariate_normal.pdf(
+                    np.transpose(particle.state_vector),
+                    mean=mean,
+                    cov=np.diag([0.75 for i in range(len(mean))])
```
sdhiscocks, line 285: Looking at this now @Isaac-JenkinsRA, and wondering about where this covariance comes from? (Sorry, I'm not familiar with paper on this filter).

stonesoup/updater/particle.py
```diff
@@ -70,7 +70,236 @@ def predict_measurement(self, state_prediction, measurement_model=None,
...
+            selected_model_sum = 0
+            # Loop over the previous models m_k-1
+            for i, model in enumerate(model_list):
+                # Looks up p(m_k|m_k-1)
+                # Note: if p(m_k|m_k-1) = 0 then p(m_k|x_1:k) = 0
+                transition_probability = transition_matrix[i][l]
+                if transition_probability == 0:
+                    break
+                # Getting required states to apply the model to that state vector
+                parent_required_state_space = particle.parent.state_vector[
```
sdhiscocks, line 268: Should this be the parent particle? Or should it be the prior?

stonesoup/updater/particle.py
```diff
@@ -70,7 +70,236 @@ def predict_measurement(self, state_prediction, measurement_model=None,
...
+
+        denominator_components = []
+        # Loop over the current models m_k
+        for l in range(len(previous_probabilities)):
+            selected_model_sum = 0
+            # Loop over the previous models m_k-1
+            for i, model in enumerate(model_list):
+                # Looks up p(m_k|m_k-1)
+                # Note: if p(m_k|m_k-1) = 0 then p(m_k|x_1:k) = 0
+                transition_probability = transition_matrix[i][l]
```
sdhiscocks, line 264: Just wondering if this should be `transition_matrix[l][i]`?

stonesoup/updater/particle.py
sdhiscocks: I can see now that this is the prior, as the predictor assigned the parent to this. I think the parent variable should be reserved for resampling really.

stonesoup/updater/particle.py
Isaac-JenkinsRA: I believe this is correct, as it is looking at the probability of the model you have compared to every other model

stonesoup/updater/particle.py
Isaac-JenkinsRA: This is concerning as I too am not sure what this arbitrary value of 0.75 is supposed to be. The covariance here should be the noise applied to a particle once it has been acted on by a transition model. Hopefully this helps.

sdhiscocks:
Replaced by #719

PR #273, opened 2020-07-09
Fix (G)OSPA Metrics

sdhiscocks:
For OSPA, cost matrix should be square, with `c` value for mismatched number of tracks and truths.

This also fixes a bug where when number of tracks or truths is 1, that the index of association would be set to 1, rather than 0 (assume this is simple off-by-one error by original author).

Also includes general code tidy and some documentation fixes.

### DIFF
--- a/stonesoup/metricgenerator/ospametric.py
+++ b/stonesoup/metricgenerator/ospametric.py
@@ -141,11 +141,12 @@ class GOSPAMetric(MetricGenerator):
 
         Returns
         ---------
-        truth_to_measured: Vector of size 1xm, which has indices of the
-                    measured objects or '-1' if unassigned.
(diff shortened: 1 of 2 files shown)


stonesoup/metricgenerator/ospametric.py
```diff
@@ -262,28 +242,39 @@ def compute_assignments(self, cost_matrix, max_iter):
...
         truth_states: list of states
+        complete: bool
+            Cost matrix will be square, with :attr:`~.c` present for where
+            there is a mismatch in cardinality
 
         Returns
         ----------
-        cost_matrix: Matrix of euclidian distance between each element in each
-        list of states
+        cost_matrix: np.ndarry
```
oharrald-Dstl, line 247: *np.ndarray

PR #278, opened 2020-07-16
Orbitalstate

jmbarr:
Introducing OrbitalState 

It inherits from State, is a 6-dimensional Cartesian representation of position and velocity in a so-called *Earth-centred intertial* (ECI)  frame of reference. The class also includes a bunch of methods to return various commonly-used orbital parameters (e.g. semi-major axis, eccentricity, etc, etc). ~~Three further classes are derived from OrbitalState allowing explicit initialisation in popular orbital parameterisation.~~ See docs for more info.


dcherrie-dstl:
Works as advertised. Jacobian matrix output and SpacecraftState would be useful additions, perhaps a future update?

jmbarr:
> Works as advertised. Jacobian matrix output and SpacecraftState would be useful additions, perhaps a future update?

So click the approve button already! Get some momentum going.

dcherrie-dstl (approved):
A really valuable addition to Stone Soup. In my limited testing it appears to work as advertised. Jacobian matrix output for transforming covariance (e.g. equinoctial to cartesian) would be the next most obvious upgrade. More than happy to approve current scope of PR for merge into main branch.

jmbarr:
hey @sdhiscocks what are these new codecov checks?

jmbarr:
A brief notebook to explain how it might work: https://gist.github.com/jmbarr/5c185eb1c106a98568b0b8a81d2f7fc9

jmbarr:
I get a local flake8 error on test_tracktotruthmetrics.py. Doesn't happen here though. ?

jmbarr:
Now I get some circle errors on docs associated with the new Prediction.from_state() function. @sdhiscocks ?

sdhiscocks:
> Now I get some circle errors on docs associated with the new Prediction.from_state() function. @sdhiscocks ?

Thanks. I think I can see what the issue is. I'll take a look at a fix.

jmbarr:
crissake, now the codecov fails.

jmbarr:
Woo. Hoo.

tjbramly-dstl (approved):
Works as expected. May be worth mentioning issues with perfect retrograde orbits (i = 180deg) in the help documentation.

stonesoup/functions.py
```diff
@@ -556,3 +556,36 @@ def mod_elevation(x):
     elif N == 3:
         x = x - 2.0 * np.pi
     return x
+
+
+def dotproduct(a, b):
+    r"""A convenience function that saves having to remember a lot of
+    counterintuitive formatting stuff inflicted on us by numpy.
```
sdhiscocks, line 563: Needs rewording.

stonesoup/orbital_functions.py
```diff
@@ -0,0 +1,415 @@
+# -*- coding: utf-8 -*-
```
sdhiscocks, line 1: May be we should create a functions submodule. Can move content of `functions.py` to `functions/__init__.py` initially, and then this file can be `functions/orbital.py`.

stonesoup/types/orbitalstate.py
```diff
@@ -0,0 +1,840 @@
...
+    2. https://spaceflight.nasa.gov/realdata/sightings/SSapplications/\
+    Post/JavaSSOP/SSOP_Help/tle_def.html
+
+    3. Broucke, R. A. & Cefola, P. J. 1972, Celestial Mechanics, Volume
+    5, Issue 3, pp. 303-310
+
+    """
+
+    coordinates = Property(
+        str, default='Cartesian',
```
sdhiscocks, line 102: Using a [`Enum`](https://docs.python.org/3/library/enum.html) rather than a string would be better for this property.

stonesoup/types/orbitalstate.py
```diff
@@ -0,0 +1,840 @@
...
+    @property
+    def _nodeline(self):
+        """The vector node line (defines the longitude of the ascending node
+        in the Primary-centered inertial frame)
+
+        Returns
+        -------
+        : numpy.array
+            The node line (defines the longitude of the ascending node)
+        """
```
sdhiscocks, line 219: `@property` docstring is wrong format. Example can be found [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)

stonesoup/types/orbitalstate.py
```diff
@@ -0,0 +1,840 @@
...
+    grav_parameter = Property(
+        float, default=3.986004418e14,
+        doc=r"Standard gravitational parameter :math:`\mu = G M`. The default "
+            r"is :math:`3.986004418 \times 10^{14} \,` "
+            r":math:`\mathrm{m}^3 \mathrm{s}^{-2}`.")
+
+    metadata = Property(
+        dict, default={}, doc="Dictionary containing metadata about orbit"
+    )
+
```
sdhiscocks, line 117: It may be worth adding a `__new__` method to return the specific types below, and then that means their `__init__` methods are used to validate arguments (and can also do coordinate conversion?).

Can see something similar in [`stonesoup.types.hypothesis`](https://github.com/dstl/Stone-Soup/blob/7707806d79a00b270934eeeb3ad5d11e83ac1281/stonesoup/types/hypothesis.py#L122-L130)

sdhiscocks (commented):
Haven't reviewed documentation yet, as the new modules also need inclusion into documentation (with `automodule`) in order to build.

jmbarr:
> Haven't reviewed documentation yet, as the new modules also need inclusion into documentation (with `automodule`) in order to build.

You mean I add an instance of automodule: to stonesoup.types.rst covering the OrbitalState class and its children? Or is this done automatically elsewhere?

sdhiscocks:
> > Haven't reviewed documentation yet, as the new modules also need inclusion into documentation (with `automodule`) in order to build.
> 
> You mean I add an instance of automodule: to stonesoup.types.rst covering the OrbitalState class and its children? Or is this done automatically elsewhere?

Yeah, and extra file/links for orbital functions.

stonesoup/orbital_functions.py
jmbarr: Sounds sensible. But also sounds sensible to have `functions/trigonometric.py`, `functions/coordinate_transforms.py`, `etc/etc.py`. Do that here, or in a different PR?

stonesoup/types/orbitalstate.py
jmbarr: I don't quite get it. `Enum` is a (restricted) set; `coordinates` has to be a member of that set. But then I'd need to create the `Enum` somewhere (where?) and run exactly the same logic over its members. 

?

stonesoup/types/orbitalstate.py
jmbarr: But there aren't any getter methods. So... just remove the docstrings? What if you want to know what the methods do?

stonesoup/types/orbitalstate.py
jmbarr: 'May' or 'must'?

Also, don't understand. Do you mean, use `__new__` to return a some general `KeplerianOrbitalState`, `TLEOrbitalState` , etc (below), then specifically `__init__` one of those? How would that work?

stonesoup/orbital_functions.py
sdhiscocks: I agree, but probably do that on another change, as it will effect loads more files.

stonesoup/types/orbitalstate.py
sdhiscocks: Yeah, it makes it more explicit what values are allowed, versus a string. You could create an Enum in this file. You can also have [descriptive strings](https://docs.python.org/3/library/enum.html#using-a-descriptive-string), rather than numbers, which may be better for this use case.

stonesoup/types/orbitalstate.py
sdhiscocks: This method if the getter. I think this should be right (but double check once docs rendered):
```python
    @property
    def _nodeline(self):
        """numpy.ndarray: The vector node line (defines the longitude of the ascending node
        in the Primary-centered inertial frame)
        """
```

stonesoup/types/orbitalstate.py
sdhiscocks: Just a suggestion. You can use `__new__` to decide what type to return, which in turn will then create an instance of that class instead (which will also call it's `__init__` method, etc.)

Would be something like:
```python
def __new__(cls, ...):
    if coordinates == "TLE":
        return super().__new__(TLEOrbitalState)
    elif ...:
        ...
    else:
        raise TypeError("Unknown coordinate system...")
```

stonesoup/orbital_functions.py
jmbarr: > > > Haven't reviewed documentation yet, as the new modules also need inclusion into documentation (with `automodule`) in order to build.
> > 
> > 
> > You mean I add an instance of automodule: to stonesoup.types.rst covering the OrbitalState class and its children? Or is this done automatically elsewhere?
> 
> Yeah, and extra file/links for orbital functions.

ok - see new docs. I've also attempted to place them somewhere clever, so `OrbitalState` is a subsection of `State` and "orbital functions" a subsection of "functions". Check that makes sense.

stonesoup/types/orbitalstate.py
jmbarr: Check docs. That work?

stonesoup/types/orbitalstate.py
jmbarr: Unless there's a strong argument for doing this then I'll defer. It's going to require me learning new stuff, and likely moving significant code around, which will introduce a fair delay.

stonesoup/types/orbitalstate.py
jmbarr: But then I can't initialise via state = OrbitalState(statevector, coordinates="Keplerian") anymore. I now have to have the wherewithal to create some `CoordinateSystem(Enum)`, say `enumerated_coordinates =  CoordinateSystem("Keplerian")` and then pass that as the keyword. Just telling a user which string to use seems easier.

sdhiscocks: But you can do: `OrbitalState(statevector, coordinates=CoordinateSystem.KEPLERIAN)`, which will support autocompletion.
Also `CoordinateSystem("Keplerian") is  CoordinateSystem(CoordinateSystem("Keplerian"))`, so can also use it to support strings still, as well as provide input validation:
```python
    self.coordinates = CoordinateSystem(self.coordinates)
```

jmbarr: What if I want to do coordinates="keplerian", or coordinates="kEplEriAN"?

sdhiscocks: > What if I want to do coordinates="keplerian", or coordinates="kEplEriAN"?

Yes, another good reason why clearly defined constants/enums are preferable vs. strings.

jmbarr: I'd argue the opposite. At present coordinates="Keplerian" or coordinates="keplerian" both work. Your proposal would make one of them fail, => less flexibility.

sdhiscocks: Where a defined supported values are available, enums (or string constants) should really be used. This way they are defined in code, as well as documentation. Whilst we should encourage use of enum value, you could still have case insensitive string support by using Enum `_missing_` class method:
```python
    @classmethod
    def _missing_(cls, value):
        for element in cls:
            if element.value.lower() == value.lower():
                return element
```

stonesoup/types/orbitalstate.py
jmbarr: I can make this work such that `a = OrbitalState(x, coordinates='TLE',...)` initiates a  `TLEOrbitalState`, but now initialisation via `b = TLEOrbitalState(x,..)` chucks an error ("no property _property_state_vector"), though `c = TLEOrbitalState(x, coordinates='TLE',...)` still works, despite the redundant keyword.

However:

I've come to the conclusion that `KeplerianOrbitalState`, `TLEOrbitalState`, etc classes are a bad idea. They are fundamentally the same thing as OrbitalState - the state vector is always the Cartesian position/velocity. It's just the initialisation that proceeds differently. So the names are misleading, and should be reserved for States whose state vectors conform to that characterisation, which might happen in the future.

jmbarr:
What dickery is CircleCI up to now? Tests fully pass on the desktop but fail here?

sdhiscocks:
> What dickery is CircleCI up to now? Tests fully pass on the desktop but fail here?

Looks like the behaviour of Enum `_missing_` method changed  between 3.6 and 3.7+. On [Python 3.6 an error is raised](https://github.com/python/cpython/blob/34df10a9a16b38d54421eeeaf73ec89828563be7/Lib/enum.py#L546-L548) and [on current version `None` is returned](https://github.com/python/cpython/blob/895591c1f0bdec5ad357fe6a5fd0875990061357/Lib/enum.py#L703-L705).

Moving the error you raise in `OrbitalState.__init__` to the `CoordinateSystem._missing_` should fix it.

stonesoup/types/orbitalstate.py
```diff
@@ -24,11 +24,15 @@ class CoordinateSystem(Enum):
...
-                return element
+        try:
+            for element in cls:
+                if element.value.lower() == value.lower():
+                    return element
+                if element.value == "TLE" and value.lower() == "twolineelement":
+                    return element
+
+        except ValueError:
+            raise ValueError("%r is not a valid %s" % (value, cls.__name__))
```
sdhiscocks, line 35: No need for `try`/`except` here. Just have error raised at the end (as means other returns haven't been caught up).

jmbarr:
> Looks like the behaviour of Enum `_missing_` method changed between 3.6 and 3.7+. On [Python 3.6 an error is raised](https://github.com/python/cpython/blob/34df10a9a16b38d54421eeeaf73ec89828563be7/Lib/enum.py#L546-L548) and [on current version `None` is returned](https://github.com/python/cpython/blob/895591c1f0bdec5ad357fe6a5fd0875990061357/Lib/enum.py#L703-L705).
> 
Isn't the error precisely caused by `None` being returned? So it should happen on 3.7+ (it doesn't) and not on 3.6 (but it does) (?!)

PR #288, opened 2020-08-11
Iterated Kalman updater

jmbarr:
A version of the Kalman updater which achieves a better linearisation of h(x) \approx H_x by progressively updating the posterior value of x and recalculating H_x until it doesn't change much.

The stopping criterion is enacted when x_(I+1) - x_i < user_defined_parameter, but there's currently no protection for the potential for an infinite while loop. I could just apply some iteration limit but suggestions on any pythonic ways of coping with this are welcome.


stonesoup/updater/kalman.py
```diff
@@ -562,3 +558,88 @@ def _posterior_covariance(self, hypothesis):
...
+        x_{k,0} &= x_{k|k-1} \ \mathrm{and} \ P_{k,0} = P_{k|k-1}
+
+    It inherits from the ExtendedKalmanUpdater as it uses the same linearisation of the sensor
+    function via the :meth:`_measurement_matrix()` function.
+    """
+
+    tolerance = Property(float, default=1e-6,
+                         doc="The value of the difference in the measure used as a stopping "
+                             "criterion.")
+    measure = Property(Measure, default=Euclidean(), doc="")
```
sdhiscocks, line 592: Doc missing

stonesoup/updater/kalman.py
```diff
@@ -562,3 +558,88 @@ def _posterior_covariance(self, hypothesis):
...
+        prev_state = hypothesis.prediction
+
+        # Get the measurement model
+        measurement_model = self._check_measurement_model(hypothesis.measurement.measurement_model)
+
+        # The first iteration is just the application of the EKF
+        post_state = super().update(hypothesis, **kwargs)
+
+        # Now update the measurement prediction mean and loop
+        while self.measure(prev_state, post_state) > self.tolerance:
```
sdhiscocks, line 626: I think it is worth having a max iteration count just in case.  Something like this should work:
```python
iterations = 0
while self.measure(prev_state, post_state) > self.tolerance \
        and iterations < self.max_iterations:
    # Do stuff here...
    iterations += 1
```

or maybe:
```python
iterations = 0
while self.measure(prev_state, post_state) > self.tolerance:
    # Do stuff here...
    iterations += 1
    if iterations > self.max_iterations:
        warnings.warn("Iterated Kalman update did not converge")
        break
```

stonesoup/updater/kalman.py
```diff
@@ -562,3 +558,88 @@ def _posterior_covariance(self, hypothesis):
...
+            post_state.hypothesis.measurement_prediction.state_vector = \
+                measurement_model.function(post_state, noise=None) + \
+                hh@(hypothesis.prediction.state_vector - post_state.state_vector)
+
+            cross_cov = self._measurement_cross_covariance(hypothesis.prediction, hh)
+            post_state.hypothesis.measurement_prediction.cross_covar = cross_cov
+            post_state.hypothesis.measurement_prediction.covar = \
+                self._innovation_covariance(cross_cov, hh, measurement_model)
+
+            prev_state = post_state  # Does this require a copy?
```
sdhiscocks, line 642: So this would require a copy if you don't want to modify the original measurement_prediction that was on the hypothesis. Not sure if this is really an issue to be honest.

If you did want to copy, suggest you just do it once before the while loop, after the first call to `super`. You shouldn't need a _deepcopy_, other than also copying `measurement_prediction` because the way you are modifying things.
```python
post_state.hypothesis = copy.copy(post_state.hypothesis)
post_state.hypothesis.measurement_prediction = copy.copy(post_state.hypothesis.measurement_prediction)
```

stonesoup/updater/kalman.py
jmbarr: That simple eh? Would have expected something a bit fancier from Python.

jmbarr: Done

jmbarr: The second way

stonesoup/updater/kalman.py
```diff
@@ -562,3 +559,100 @@ def _posterior_covariance(self, hypothesis):
...
+        measurement_model = self._check_measurement_model(hypothesis.measurement.measurement_model)
+
+        # The first iteration is just the application of the EKF
+        post_state = super().update(hypothesis, **kwargs)
+
+        # Now update the measurement prediction mean and loop
+        iterations = 0
+        while self.measure(prev_state, post_state) > self.tolerance:
+
+            if iterations > self.max_iterations:
```
oharrald-Dstl, line 637: Is this 'iterations' value being incremented somewhere? I can't see.

stonesoup/updater/kalman.py
```diff
@@ -562,3 +559,100 @@ def _posterior_covariance(self, hypothesis):
...
+    .. math::
+
+        H_i &= h^{\prime}(x_{k,i}), \ K_i = P_{k|k-1} H_i^T (H_i P_{k|k-1} H_i^T + R)^{-1}
+
+    and
+
+    .. math::
+
+        x_{k,0} &= x_{k|k-1} \ \mathrm{and} \ P_{k,0} = P_{k|k-1}
+
```
oharrald-Dstl, line 585: Formatting errors somewhere here.

jmbarr: Probably me trying to be too clever. Done it simpler.

stonesoup/updater/kalman.py
jmbarr: Oh, heh, no, seems not. My bad. Good spot. Added `iterations += 1`

jswright-dstl (approved):

oharrald-Dstl (approved):

PR #289, opened 2020-08-19
Add smooth transition model maker

oharrald-Dstl:
This branch adds the 'get_smooth_transition_models' method.
This is to provide more realistic transition models/times for the 'MultiTransitionMovingPlatform' class when simulating motion between given 2D cartesian coordinates (and timestamps).

Note:
The mathematics in the model maker are not using meaningful variable names for the sake of not cluttering up the code. Instead
I've added in-line comments where unclear. Alterations to this are welcome.


stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,266 @@
...
+        for each corresponding transition model in transition_models.
+
+    Notes
+    -----
+    x_coords, y_coords and times must be of same length.
+    """
+
+    state = deepcopy(initial_state)  # don't alter platform state with calculations
+
+    if len(x_coords) != len(y_coords) or len(x_coords) != len(times):
```
sdhiscocks, line 43: This might be clearer:
```python
if not len(x_corrds) == len(y_coords) == len(times):
```

stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,266 @@
...
+        raise ValueError('x_coords, y_coords and times must be same length')
+
+    transition_models = []
+    transition_times = []
+
+    for i in range(1, len(times)):  # make models to reach next coordinates from previous.
```
sdhiscocks, line 49:
```python
for x_coord, y_coord, time in zip(x_coords, y_coords, times):
```

stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,266 @@
...
+
+        dx = x_coords[i] - state.state_vector[0]  # distance to next x-coord
+        dy = y_coords[i] - state.state_vector[2]  # distance to next y-coord
+
+        if i == 0 and dx == 0 and dy == 0:
```
sdhiscocks, line 54: `i` is never equal to `1`

stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,266 @@
...
+        if vx != 0 or vy != 0:  # if velocity is 0, keep previous bearing
+            a = np.arctan2(vy, vx)  # initial bearing
+
+        if dx == 0 and dy == 0 and vx == 0 and vy == 0:  # if at destination with 0 speed, stay
+            transition_times.append(times[i])
+            transition_models.append(CombinedLinearGaussianTransitionModel((ConstantVelocity(0),
+                                                                            ConstantVelocity(0))))
+            continue
+
+        d = np.sqrt(dx ** 2 + dy ** 2)  # distance to next coord
```
sdhiscocks, line 69: This would be clear for calculations, grouping by operator order.
```python
np.sqrt(dx**2 + dy**2)
```

Other places in PR could be changed as well.

stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,266 @@
...
+from copy import deepcopy
+
+import numpy as np
+
+from stonesoup.models.transition.linear import ConstantTurn, ConstantVelocity,\
+    CombinedLinearGaussianTransitionModel
+from stonesoup.types.array import StateVector
+
+
+def get_smooth_transition_models(initial_state, x_coords, y_coords, times, turn_rate):
```
sdhiscocks, line 12: I wonder if "generate" or "create" would be better than "get" for the function name.

stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,266 @@
...
+        must be at each corresponding coordinate.
+    turn_rate: Float
+        Angular turn rate (radians/second) measured anti-clockwise from positive x-axis.
+
+    Returns
+    -------
+    transition_models: A list of :class:`~.ConstantTurn` and
+        :class:`~.Point2PointConstantAcceleration` transition models.
+    transition_times: A list of :class:`~.datetime.timedelta` dictating the transition time
+        for each corresponding transition model in transition_models.
```
sdhiscocks, line 34: Format of these isn't quite right, so they don't render properly. e.g. for first parameter
```
    Parameters
    ----------
    initial_state: :class:`~.State`
        The initial state of the platform.
```

stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,275 @@
...
+
+    if not len(x_coords) == len(y_coords) == len(times):
+        raise ValueError('x_coords, y_coords and times must be same length')
+
+    transition_models = []
+    transition_times = []
+
+    for x_coord, y_coord, time in zip(x_coords[1:], y_coords[1:], times[1:]):
+
+        dx = x_coord - state.state_vector[0]  # distance to next x-coord
```
etfrogers-dstl, line 56: It is assumed here (and several other places below) that `len(state.statev_vector)==6`, assuming [x, vx, y, vy, z, vz]. 

Is this safe? Could you get access to a position/velocity mapping to use instead? I'm not thinking that you add one to this class, but `Platform`s have them already.

stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,275 @@
...
+    initial_state: :class:`~.State` The initial state of the platform.
+    x_coords:
+        A list of int/float x-coordinates (cartesian) in the order that the target must follow.
+    y_coords:
+        A list of int/float y-coordinates (cartesian) in the order that the target must follow.
+    times:
+        A list of :class:`~.datetime.datetime` dictating the times at which the target must be at
+        each corresponding coordinate.
+    turn_rate: Float
+        Angular turn rate (radians/second) measured anti-clockwise from positive x-axis.
```
etfrogers-dstl, line 30: Is this a maximum, preferred or fixed turn rate (i.e. will this always be the same rate even if a small turn is needed)?

docs/demos/Smooth_Platform_Coordinate_Transitions.py
```diff
@@ -0,0 +1,214 @@
...
+    print('Duration: ', transition_time.total_seconds(), 's ',
+          'Model: ', type(transition_model), end=' ')
+    if isinstance(transition_model, ConstantTurn):
+        print('turn-rate: ', transition_model.turn_rate)
+    else:
+        print('x-acceleration: ', transition_model.ax, end=', ')
+        print('y-acceleration: ', transition_model.ay)
+
+# %%
+# The last transition model calls a :class:`~.Point2PointStop` transition model to bring the target
```
etfrogers-dstl, line 113: This overshoot: It's not clear to me why a linear transition would lead to an overshoot. Surely it could just accelerate less. Maybe this could be clarified?

docs/demos/Smooth_Platform_Coordinate_Transitions.py
```diff
@@ -0,0 +1,214 @@
...
+
+# %%
+platform_coords = []
+sim_rate = 100  # 'sim_rate'-seconds each time-step.
+for i in range(int((times[-1] - times[0]).total_seconds()/sim_rate)):
+    platform_coords.append((platform.state.state_vector[0], platform.state.state_vector[2]))
+    platform.move(timestamp=start+timedelta(seconds=sim_rate*i))
+platform_coords.append((platform.state.state_vector[0], platform.state.state_vector[2]))
+
+# ax.plot([coord[0] for coord in platform_coords],
```
etfrogers-dstl, line 201: Guess the commented out code could be removed.

stonesoup/simulator/transition.py
oharrald-Dstl: This is a fixed turn rate. Variable turn rate will require a bit more work but is something I'm interested in adding. Perhaps providing a max and min turn rate and having some logic to decide what to choose inbetween?

stonesoup/simulator/transition.py
oharrald-Dstl: It is quite an assumption. Probably not necessary. I'll look in to making it less specific.

stonesoup/simulator/transition.py
```diff
@@ -0,0 +1,275 @@
...
+        vx = ux + self.ax*t  # final x-speed
+        vy = uy + self.ay*t  # final y-speed
+
+        self.over_shot = False
+
+        if np.sign(ux) != np.sign(vx) or np.sign(uy) != np.sign(vy):
+            # if the resultant speeds have changed sign, the platform must have turned-back on
+            # itself hence must have overshot the destination
+            self.over_shot = True
+            self.new_model = Point2PointStop(state, destination, duration)  # use a different model
```
etfrogers-dstl, line 177: This  `new_model` construction seems a little inelegant. Unfortunately I can't think of a way brilliant way around it... Perhaps a "factory function" that generates the required model and returns either one or the other?

etfrogers-dstl (commented):
Looks like a very useful addition to me. Some individual comments as well as the one in the review.

etfrogers-dstl:
@gawebb-dstl Might be interested in this: I think he mentioned a waypoint following model, which might be similar.

stonesoup/simulator/transition.py
etfrogers-dstl: So the platform is always either turning **or** accelerating? Never both?

docs/demos/Smooth_Platform_Coordinate_Transitions.py
oharrald-Dstl: In a situation where linearly decelerating to zero speed exactly at the next coordinate leads to the platform arriving there early, there is an issue with overshooting. If we were to apply a stronger deceleration, the platform would hit zero speed before the next coord. So a weaker deceleration is needed, hence the platform will go past the next coord and reverse back.

stonesoup/simulator/transition.py
oharrald-Dstl: Agreed. Okay as a placeholder, but something that could be cleaned at a later date?

stonesoup/simulator/transition.py
oharrald-Dstl: Yes, quite a basic set of transition models, but allowing speed change during turn seems complicated.

stonesoup/simulator/transition.py
oharrald-Dstl: This can be done in 2D cartesian too, so the model maker does not assume a 'z-direction'. While it is an assumption that the state should be of form (x, vx, y, vy), this form is contained completely within the model maker. It only returns transition models, which can then be used to apply to different state space arrangements according to how one sets up their own platforms (and corresponding position/velocity mappings).

docs/demos/Smooth_Platform_Coordinate_Transitions.py
etfrogers-dstl: Ahh of course.

stonesoup/simulator/transition.py
etfrogers-dstl: Perhaps it would be worth adding this information to the documentation

etfrogers-dstl (approved):

etfrogers-dstl:
@oharrald-Dstl Is there a reason why your transition models don't inherit from `TransitionModel` Is it because `rvs` and `pdf` don't make sense?

stonesoup/simulator/transition.py
etfrogers-dstl: What do you think of the approach in #298? If you like it, I think you can just merge into this PR, before this one is merged.

oharrald-Dstl:
> @oharrald-Dstl Is there a reason why your transition models don't inherit from `TransitionModel` Is it because `rvs` and `pdf` don't make sense?

I cannot think of a reason that they do not inherit from this. Since they are rather niche transition models I did not consider adding them to the inheritance. Since they are used to get a platform exactly from one point to another, there should be no noise generated, but I suppose this could be as simple as returning 0/None (not sure) in the rvs method.

stonesoup/simulator/transition.py
```diff
@@ -171,10 +180,7 @@ def __init__(self, state, destination, duration):
         self.over_shot = False
```
etfrogers-dstl, line 180: Oops, this variable should be removed.
```suggestion
```

etfrogers-dstl:
Looks good to me now. I like the idea and may well need to make use of it!

sdhiscocks (changes requested):
Looks great. Would be interesting in future to extend these to 3D. Could also see a use case where you define a Platform, where it's arguments are same as `create_smooth_transition_models`, just to ease creation and able to store in config.

PR #291, opened 2020-08-19
Add correct error for simple measurement initiator

oharrald-Dstl:
Needed exception for NotImplementedError on initiate method.


jswright-dstl (approved):

sdhiscocks:
I think it's worth adding a test here, covering both covering skip option and handling `NotImplementedError`.

sdhiscocks (approved):

PR #298, opened 2020-09-14
Refactor to avoid nested models

etfrogers-dstl:
Avoids the need for the `new_model` approach in transition models. 

Note this is a PR onto #289 , not a pull to `master`. I suggest that this PR is either pulled onto #289 before merge, or that #289 is updated to include this commit, then this PR is closed.


oharrald-Dstl (commented):
Looks like a much better approach to the 'overshoot' issue.

PR #304, opened 2020-09-24
Correct property name in specification error

oharrald-Dstl:
Error previously used the wrong property name (taken from previous loop)
[ci skip]


sdhiscocks (approved):

etfrogers-dstl (approved):
Well spotted. Looks correct.

PR #307, opened 2020-09-28
Fix issue with properties being reinstated from ancestors

sdhiscocks:
This issue was caused by removed properties in a new classes parent(s), being added into the new class by using properties from further ancestors. This change ensures that only properties of direct parents are used.

New test added which demonstrated this, as prior to fix, `_TestSubSubclassRemoveProperty` would fail to initialise, as it would expect `property_a` to be provided on initialisation, despite being fixed value in parent.

Also tidied up handling of properties, as there was multiple loops, assignments, etc. that weren't needed or could be consolidated.

### DIFF
--- a/stonesoup/base.py
+++ b/stonesoup/base.py
@@ -197,11 +197,14 @@ class BaseMeta(ABCMeta):
 
         cls._subclasses = set()
         cls._properties = OrderedDict()
-        # Update subclass lists, and update properties (in reverse order)
+        # Update subclass lists, and update properties from direct bases (in reverse order as
+        # first defined class must take precedence, and dictionary update overwrites)
         for bcls in reversed(cls.mro()[1:]):
             if type(bcls) is mcls:
                 bcls._subclasses.add(cls)
-                cls._properties.update(bcls._properties)
+                if bcls in bases:
+                    cls._properties.update(bcls._properties)
+
         for key, value in namespace.items():
             if isinstance(value, Property):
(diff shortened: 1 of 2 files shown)


etfrogers-dstl (approved):
Looks cleaner - good to clean up some of the "organic" code. 

Only other thing that could be addressed the same time is the signature for `__new__` and `__prepare__` where PyCharm is complaining that the first argument is conventionally called `mcs`. Not terribly worried if it doesn't  happen, but seems this is a opportunity to do that minor tidy.

sdhiscocks:
> Only other thing that could be addressed the same time is the signature for `__new__` and `__prepare__` where PyCharm is complaining that the first argument is conventionally called `mcs`. Not terribly worried if it doesn't happen, but seems this is a opportunity to do that minor tidy.

Not sure about this, as I've never seen `mcs` used before. Python standard library uses `mcls` in ABC module.

etfrogers-dstl:
> > Only other thing that could be addressed the same time is the signature for `__new__` and `__prepare__` where PyCharm is complaining that the first argument is conventionally called `mcs`. Not terribly worried if it doesn't happen, but seems this is a opportunity to do that minor tidy.
> 
> Not sure about this, as I've never seen `mcs` used before. Python standard library uses `mcls` in ABC module.

Fine, leave it.

gawebb-dstl (approved):

oharrald-Dstl (approved):

PR #310, opened 2020-10-01
Fix nonlinear models to give correct elevation

etfrogers-dstl:
I have been having a problem with the elevation measurements from measurement models, particularly with sensors with non-zero z velocity.

A minimal example would be a sensor on a platform with state `(0, 1, 0, 0, 0, 1)` and a target state `(10, 0, 0, 0, 10, 0)`.
I would expect an elevation of 0 (because the sensor is pointing directly at the target).

The current measurement models give elevation = pi/2, which doesn't seem to make sense.

This PR introduces a set of tests like the above (and a few others), and changes `NonLinearGaussianMeasurement._rotation_matrix()` such that they pass. 

I have to admit that, while I'm pretty convinced of the tests, **I'm not so sure that the fix is correct** (that is - it works, but does it make sense? Should the fix be made in a different place?), and am happy to discuss that.


stonesoup/models/measurement/nonlinear.py
```diff
@@ -138,7 +138,7 @@ def _rotation_matrix(self) -> np.ndarray:
         """
 
         theta_x = -self.rotation_offset[0, 0]
-        theta_y = -self.rotation_offset[1, 0]
+        theta_y = self.rotation_offset[1, 0]
         theta_z = -self.rotation_offset[2, 0]
```
jjosborne-dstl, line 143: I think it would be better if MovingPlatform's orientation was changed instead of the rotation matrix. That way we can keep the rotations as clockwise.

etfrogers-dstl: That wouldn't fix these tests: they only uses models, not platforms.

The velocity->orientation transform in my tests is done (as it is in platform, but it looks sensible to me) using.
``` python
    _, bearing, elevation = cart2sphere(*sensor_velocity)
    orientation = StateVector([0, elevation, bearing])
```

How would you change this?

gawebb-dstl:
It's probably worth adding some more documentation around the rotations etc, so it will be easy easier to understand the next time

stonesoup/models/measurement/nonlinear.py
jjosborne-dstl: I've given it some more thought and I think It will be fine as `theta_y = self.rotation_offset[1, 0]` but it is probably a good idea to specify how the rotations work in the doc string and checking if we need to change any other rotation matrices.

etfrogers-dstl:
To (maybe!) clarify: current behaviour gives 
`sensor position (0, 0, 0), target (1, 0, 1)` => `elevation == -45 degrees`. 

That just looks like a sign convention "problem" and after my change it gives elevation=+45. You can argue whether the fix is correct or whether actually current behaviour is more consistent. 

That's is not the oddest part though: If you have a sensor on a platform with velocity (1, 0, 1) and positions as above then `elevation == +90 degrees`. I would expect it to be 0. That is more than just sign conventions, and (I think) is due to different sign conventions in different places. My "fix" makes `elevation == 0`.

jjosborne-dstl:
oops don't know how I did that.

sdhiscocks:
So as we've discussed offline, I think the issue is mixed use of elevation/theta. One clockwise from x/y plane (e.g. `sphere2cart`), and rotation matrix being based on anti-clockwise from z-axis (e.g. rotation matrices). I suggest we update the code to use ISO standard used as mentioned in [this Wikipedia article](https://en.wikipedia.org/wiki/Spherical_coordinate_system).

Would also be good to document angles better. I think we could do with adding a "conventions" page, which highlights this as well as other things, like fact that our state spaces are usually (x, \dot{x}, y, \dot{y}) rather than (x, y, \dot{x}, \dot{y})

gawebb-dstl:
> Would also be good to document angles better. I think we could do with adding a "conventions" page, which highlights this as well as other things, like fact that our state spaces are usually (x, \dot{x}, y, \dot{y}) rather than (x, y, \dot{x}, \dot{y})

I think a "conventions" page would definitely be useful. Especially for people that are new to tracking/Stone Soup. I remember when I first started using Stonesoup, I was quite frustrated that all the functions required (x, \dot{x}, y, \dot{y}) but didn't state that needed the statevectors in that format

etfrogers-dstl:
> So as we've discussed offline, I think the issue is mixed use of elevation/theta. One clockwise from x/y plane (e.g. `sphere2cart`), and rotation matrix being based on anti-clockwise from z-axis (e.g. rotation matrices). I suggest we update the code to use ISO standard used as mentioned in [this Wikipedia article](https://en.wikipedia.org/wiki/Spherical_coordinate_system).
> 
> Would also be good to document angles better. I think we could do with adding a "conventions" page, which highlights this as well as other things, like fact that our state spaces are usually (x, \dot{x}, y, \dot{y}) rather than (x, y, \dot{x}, \dot{y})

Reading that article (and if I understand it correctly): That would mean that a target directly in front of a sensor should have an elevation of 90 degrees. Do I understand that correctly? 

This behavior is what the code (before this PR) did for a sensor position (0, 0, 0), velocity  (1, 0, 1) and target position (1, 0, 1) => elevation +90degrees. I have to say, my initial reaction to this result was that it was nonsense, but it looks like it may have been correct.

However, sensor position (0, 0, 0), velocity  (1, 0, 0) and target position (1, 0, 0) should, by my reading **also** have an elevation of +90 degrees. And it doesn't. Have I misunderstood the convention? Or just misunderstood everything?!  :)

sdhiscocks:
> > So as we've discussed offline, I think the issue is mixed use of elevation/theta. One clockwise from x/y plane (e.g. `sphere2cart`), and rotation matrix being based on anti-clockwise from z-axis (e.g. rotation matrices). I suggest we update the code to use ISO standard used as mentioned in [this Wikipedia article](https://en.wikipedia.org/wiki/Spherical_coordinate_system).
> > Would also be good to document angles better. I think we could do with adding a "conventions" page, which highlights this as well as other things, like fact that our state spaces are usually (x, \dot{x}, y, \dot{y}) rather than (x, y, \dot{x}, \dot{y})
> 
> Reading that article (and if I understand it correctly): That would mean that a target directly in front of a sensor should have an elevation of 90 degrees. Do I understand that correctly?
> 
Yes
> This behavior is what the code (before this PR) did for a sensor position (0, 0, 0), velocity (1, 0, 1) and target position (1, 0, 1) => elevation +90degrees. I have to say, my initial reaction to this result was that it was nonsense, but it looks like it may have been correct.
>
I think the issue at the moment (before PR) is that the rotation matrix currently rotates assuming the angle goes from z-axis, but `cart2sphere` is from x-y plane. So the sensor was incorrectly rotated down, and then angle calculate positive up from there.

> However, sensor position (0, 0, 0), velocity (1, 0, 0) and target position (1, 0, 0) should, by my reading **also** have an elevation of +90 degrees. And it doesn't. Have I misunderstood the convention? Or just misunderstood everything?! :)
>
This would be 0 degrees, as  angles are currently from the from x-y plane.


I think we should go ahead with the PR as it fixes the bug accounting for incorrect direction on rotation matrix, and can raise another issue to discuss if we want to change elevation angles to go from x-y plane clockwise, or from z axis anti-clockwise.

DaveKirkland:
So just looking at the cart2sphere and sphere2cart in the master. The elevation angle is measured from the x,y plane not the z axis as per the Wiki article. This is stated in the documentation of the functions and the  code has:
`theta = np.arcsin(z/rho)
z = rho * np.sin(theta)`
which corresponds with that as well.

etfrogers-dstl:
> So just looking at the cart2sphere and sphere2cart in the master. The elevation angle is measured from the x,y plane not the z axis as per the Wiki article. This is stated in the documentation of the functions and the code has:
> `theta = np.arcsin(z/rho)`
> `z = rho * np.sin(theta)`
> which corresponds with that as well.

This is my understanding also.

> I haven't checked the models to see if they make any adjustments to these angles.

They don't make adjustments, but the bit of the model code that I changed is in `_rotation_matrix`.

> So I don't know why the target at (1,0,1) was coming up as -45 deg. If we do end up changing this function, we also have to be mindful of how the modulo function works when adding elevations - the azimuth angle covers 0 - 360 degs, but elevation only covers 180 deg e.g. 0-180, or -90 to 90.
> 
> I've not sure how the velocity portion of the state vector is supposed to affect the bearing elevation measurements?

In this case (the sensor is on a moving platform), the velocity sets the orientation of the sensor. That is: the sensor is looking along the direction of the velocity. It is a proxy for setting the orientation. 

So a sensor (or platform) moving with velocity (1,0,1) is "looking" on bearing 0 and at elevation +45. Therefore, it should be looking directly at a target at location (1,0,1) and the reported (that is, relative) bearing and elevation should both be 0. Does that make sense?

DavidFCrouse:
You might want to add an option to select the coordinate system mean for spherical coordinates with the one that you mentioned as the default. The reason is that things like antenna patterns are usually given in a different spherical coordinate system. You can do something like what I did in the Tracker Component Library:

https://github.com/USNavalResearchLaboratory/TrackerComponentLibrary/blob/master/Coordinate%20Systems/spher2Cart.m

 

As another comment, in higher-fidelity simulations, the velocity is not the pointing direction. Due to wind and currents, the direction that the nose of an aircraft or the bow of a ship points is not the direction it is moving. You have to steer slightly into the wind to go straight.

 

-          David Crouse

sdhiscocks (approved):
Happy this fixes rotation issue. Let's move discussion about angles/rotations to #316

DaveKirkland:
> 
> 
> > So just looking at the cart2sphere and sphere2cart in the master. The elevation angle is measured from the x,y plane not the z axis as per the Wiki article. This is stated in the documentation of the functions and the code has:
> > `theta = np.arcsin(z/rho)`
> > `z = rho * np.sin(theta)`
> > which corresponds with that as well.
> 
> This is my understanding also.
> 
> > I haven't checked the models to see if they make any adjustments to these angles.
> 
> They don't make adjustments, but the bit of the model code that I changed is in `_rotation_matrix`.
> 
> > So I don't know why the target at (1,0,1) was coming up as -45 deg. If we do end up changing this function, we also have to be mindful of how the modulo function works when adding elevations - the azimuth angle covers 0 - 360 degs, but elevation only covers 180 deg e.g. 0-180, or -90 to 90.
> > I've not sure how the velocity portion of the state vector is supposed to affect the bearing elevation measurements?
> 
> In this case (the sensor is on a moving platform), the velocity sets the orientation of the sensor. That is: the sensor is looking along the direction of the velocity. It is a proxy for setting the orientation.
> 
> So a sensor (or platform) moving with velocity (1,0,1) is "looking" on bearing 0 and at elevation +45. Therefore, it should be looking directly at a target at location (1,0,1) and the reported (that is, relative) bearing and elevation should both be 0. Does that make sense?

Ok - That helps explain what it is doing, but not necessarily why. I'm not sure I follow all the rotational interactions between, platforms, sensors and states. I'll post more on the Issue #316.

DaveKirkland (approved):
I'm good with these changes. The other issues will be addressed in #316.

PR #314, opened 2020-10-12
Change 'Ambiguity' metric to return 1 when there are no tracks/truths

oharrald-Dstl:
When there are no tracks or truths, there is no ambiguity as to track association. Therefore the target score of 1 makes more sense to return than the 'worst-case' of 0. [ci skip]


sdhiscocks (approved):

rjgreen-dstl (approved):
makes good sense

PR #317, opened 2020-10-15
Modify custom models to take state deep-copies

oharrald-Dstl:
I don't fully understand why this is needed, but it prevents some bugs when rerunning a platform simulation.
Since the model maker is only called once per platform, having a deepcopy shouldn't slow things down much.
[ci skip]


sdhiscocks (approved):
This makes sense as you want to maintain the state that was used to create the model, and the generation function is modifying the state.

sglvladi (approved):

PR #318, opened 2020-10-16
Add closed continuous (real) interval type

oharrald-Dstl:
Add set-like implementation of closed continuous real intervals

### DIFF
--- a/stonesoup/types/interval.py
+++ b/stonesoup/types/interval.py
@@ -3,10 +3,10 @@ import copy
 import operator
 from itertools import combinations
 from numbers import Real
-from typing import Sequence, Union, MutableSequence
+from typing import Sequence, Union, MutableSequence, Tuple
 
-from stonesoup.base import Property
-from stonesoup.types import Type
+from ..base import Property
+from ..types import Type
 
 
 class Interval(Type):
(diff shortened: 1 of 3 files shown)


stonesoup/types/interval.py
```diff
@@ -0,0 +1,336 @@
...
+        while current_interval.left >= self.left:
+            interval_list.append(current_interval)
+            current_interval = ClosedContinuousInterval(current_interval.left - interval_size,
+                                                        current_interval.left)
+        return iter(interval_list)
+
+    def __contains__(self, item):
+        if isinstance(item, Real):
+            return self.left <= item <= self.right
+        else:
```
sdhiscocks, line 106: Could/should this also support intervals?

stonesoup/types/interval.py
```diff
@@ -0,0 +1,336 @@
...
+            return False
+
+    def __str__(self):
+        return '[' + str(self.left) + ', ' + str(self.right) + ']'
+
+    def __add__(self, other):
```
sdhiscocks, line 112: Wonder if this should be `union`/`|`/`__or__` like set type instead of `+`/`__add__`?

Could also include `intersection`/`&`/`__and__`?

stonesoup/types/interval.py
```diff
@@ -0,0 +1,336 @@
...
+        lb = min(self.left, other.left)
+        ub = max(self.right, other.right)
+
+        if ub - lb < a_diff + b_diff:
+            return lb, ub
+        else:
+            return False
+
+
+class DisjointIntervals:
```
sdhiscocks, line 187: This may be able to have a more `set` like interface as well?

stonesoup/types/interval.py
oharrald-Dstl: Yes, that makes more sense.

sdhiscocks (approved):

jswright-dstl (approved):

PR #320, opened 2020-10-22
Add GaussianState initialization to WeightedGaussianState.

DaveKirkland:
This change allows a WeightedGaussianState to be created from a GaussianState, rather than specifying, state, cover and timestamp individually. It throws an exception if you over specify i.e. give it a state and a GaussianState. 

There are a couple ways of doing this - I choose not to have the gaussian_state as a Property, because this could lead to an inconsistency in the Timestamps. Note - changes to the state_vector or covar would be linked i.e. changing one would change the other. 

Please check the generated documentation to see if the appropriate.

Note - I also fixed an issue in the unit test. It wasn't properly testing the weight Property.

### DIFF
--- a/stonesoup/types/state.py
+++ b/stonesoup/types/state.py
@@ -130,7 +130,9 @@ class GaussianState(State):
     covar: CovarianceMatrix = Property(doc='Covariance matrix of state.')
 
     def __init__(self, state_vector, covar, *args, **kwargs):
-        covar = CovarianceMatrix(covar)
+        # Don't cast away subtype of covar if not necessary
+        if not isinstance(covar, CovarianceMatrix):
+            covar = CovarianceMatrix(covar)
         super().__init__(state_vector, covar, *args, **kwargs)
         if self.state_vector.shape[0] != self.covar.shape[0]:
             raise ValueError(
@@ -184,6 +186,48 @@ class WeightedGaussianState(GaussianState):
     for a GaussianMixtureState.
     """
     weight: Probability = Property(default=0, doc="Weight of the Gaussian State.")
+
+    @property
+    def gaussian_state(self):
+        """The Gaussian state."""
+        return GaussianState(self.state_vector,
+                             self.covar,
+                             timestamp=self.timestamp)
+
+    @classmethod
+    def from_gaussian_state(cls, gaussian_state, *args, copy=True, **kwargs):
+        r"""
+        Returns a WeightedGaussianState instance based on the gaussian_state.
+
+        Parameters
+        ----------
+        gaussian_state : :class:`~.GaussianState`
+            The guassian_state used to create the new WeightedGaussianState.
+        \*args : See main :class:`~.WeightedGaussianState`
+            args are passed to :class:`~.WeightedGaussianState` __init__()
+        copy : Boolean, optional
+            If True, the WeightedGaussianState is created with copies of the elements
+            of gaussian_state. The default is True.
+        \*\*kwargs : See main :class:`~.WeightedGaussianState`
+            kwargs are passed to :class:`~.WeightedGaussianState` __init__()
+
+        Returns
+        -------
+        :class:`~.WeightedGaussianState`
+            Instance of WeightedGaussianState.
+        """
+        state_vector = gaussian_state.state_vector
+        covar = gaussian_state.covar
+        timestamp = gaussian_state.timestamp
+        if copy:
+            state_vector = state_vector.copy()
+            covar = covar.copy()
+        return cls(
+            state_vector=state_vector,
+            covar=covar,
+            timestamp=timestamp,
+            *args, **kwargs
+        )
 
 
 class TaggedWeightedGaussianState(WeightedGaussianState):
--- a/stonesoup/types/tests/test_state.py
+++ b/stonesoup/types/tests/test_state.py
@@ -99,9 +99,33 @@ def test_weighted_gaussian_state():
     mean = StateVector([[1], [2], [3], [4]])  # 4D
     covar = CovarianceMatrix(np.diag([1, 2, 3]))  # 3D
     weight = 0.3
+    timestamp = datetime.datetime.now()
     with pytest.raises(ValueError):
-        a = WeightedGaussianState(mean, covar, weight)
-        assert a.weight == weight
+        WeightedGaussianState(mean, covar, timestamp, weight)
+
+    # Test initialization using a GuassianState
+    mean = StateVector([[1], [2], [3], [4]])  # 4D
+    covar = CovarianceMatrix(np.diag([1, 2, 3, 4]))
+    weight = 0.3
+    gs = GaussianState(mean, covar, timestamp=timestamp)
+    wgs = WeightedGaussianState.from_gaussian_state(gaussian_state=gs, weight=weight)
+    assert np.array_equal(gs.state_vector, wgs.state_vector)
+    assert np.array_equal(gs.covar, wgs.covar)
+    assert gs.timestamp == wgs.timestamp
+    assert weight == wgs.weight
+    assert wgs.state_vector is not gs.state_vector
+    assert wgs.covar is not gs.covar
+
+    # Test copy flag
+    wgs = WeightedGaussianState.from_gaussian_state(gaussian_state=gs, copy=False)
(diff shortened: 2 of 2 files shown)


sdhiscocks:
I'm not sure that changing the init method and adding a Property is ideal approach. I think adding a specific class method would be cleaner (suggested code modified below, diffed from _master_ for clarity):
```diff
diff --git a/stonesoup/types/state.py b/stonesoup/types/state.py
index cfa5f72f..1fd3eb76 100644
--- a/stonesoup/types/state.py
+++ b/stonesoup/types/state.py
@@ -182,9 +182,37 @@ class WeightedGaussianState(GaussianState):
 
     Gaussian State object with an associated weight.  Used as components
     for a GaussianMixtureState.
+
+    Example
+    -------
+    >>> gs = GaussianState(state, covar, timestamp)
+    >>> wgs = WeightedGaussianState.from_gaussian_state(gs, weight=0.5)
     """
     weight: Probability = Property(default=0, doc="Weight of the Gaussian State.")
 
+    @property
+    def gaussian_state(self):
+        """The Gaussian state."""
+        return GaussianState(self.state_vector,
+                             self.covar,
+                             timestamp=self.timestamp)
+
+    @classmethod
+    def from_gaussian_state(cls, gaussian_state, *args, copy=True, **kwargs):
+        """TODO: Doc string"""
+        state_vector = gaussian_state.state_vector
+        covar = gaussian_state.covar
+        timestamp = gaussian_state.timestamp
+        if copy:
+            state_vector = state_vector.copy()
+            covar = covar.copy()
+        return cls(
+            state_vector=state_vector,
+            covar=covar,
+            timestamp=timestamp,
+            *args, **kwargs
+        )
+
 
 class TaggedWeightedGaussianState(WeightedGaussianState):
     """Tagged Weighted Gaussian State Type
```

Also added a `copy` option, which can be used to copy the arrays, which probably should default to `True`

DaveKirkland:
> 
> 
> I'm not sure that changing the init method and adding a Property is ideal approach. I think adding a specific class method would be cleaner (suggested code modified below, diffed from _master_ for clarity):
>
> Also added a `copy` option, which can be used to copy the arrays, which probably should default to `True`

Yes, I think that is a better approach. The use of the Property is likely to break if other things change and from the __init__() code it is hard to see what it is doing. I'll code it up and add the tests.

DaveKirkland:
> 
> 
> > I'm not sure that changing the init method and adding a Property is ideal approach. I think adding a specific class method would be cleaner (suggested code modified below, diffed from _master_ for clarity):
> > Also added a `copy` option, which can be used to copy the arrays, which probably should default to `True`
> 
> Yes, I think that is a better approach. The use of the Property is likely to break if other things change and from the **init**() code it is hard to see what it is doing. I'll code it up and add the tests.

@sdhiscocks I've updated the file and squashed the commits.  There is one strange thing I found when trying to code the unit test. I was trying to test the copy=False option. When I run the following:
```
gs = GaussianState(state_vector=[3, 3], covar=np.array(np.zeros((2,2))))
wgs = WeightedGaussianState.from_gaussian_state(gaussian_state=gs, copy=False)
gs.state_vector is wgs.state_vector
gs.covar is wgs.covar
```

The first one is True while the covar is False. I couldn't figure out why. Both the state_vector and covar are still linked i.e. if you change one then the other is changed as well. Curious.

sdhiscocks:
> There is one strange thing I found when trying to code the unit test. I was trying to test the copy=False option. When I run the following:
> ```
> gs = GaussianState(state_vector=[3, 3], covar=np.array(np.zeros((2,2))))
> wgs = WeightedGaussianState.from_gaussian_state(gaussian_state=gs, copy=False)
> gs.state_vector is wgs.state_vector
> gs.covar is wgs.covar
> ```
> 
> The first one is True while the covar is False. I couldn't figure out why. Both the state_vector and covar are still linked i.e. if you change one then the other is changed as well. Curious.

Looks like in the `State` type it checks if `state_vector` is already a `StateVector` but `GaussianState` doesn't do the same for `covar` to check if it's already a `CovarianceMatrix`. Would be good to add similar logic to avoid needlessly creating a new instance and fix copying.

DaveKirkland:
> 
> 
> > There is one strange thing I found when trying to code the unit test. I was trying to test the copy=False option. When I run the following:
> > ```
> > gs = GaussianState(state_vector=[3, 3], covar=np.array(np.zeros((2,2))))
> > wgs = WeightedGaussianState.from_gaussian_state(gaussian_state=gs, copy=False)
> > gs.state_vector is wgs.state_vector
> > gs.covar is wgs.covar
> > ```
> > 
> > 
> > The first one is True while the covar is False. I couldn't figure out why. Both the state_vector and covar are still linked i.e. if you change one then the other is changed as well. Curious.
> 
> Looks like in the `State` type it checks if `state_vector` is already a `StateVector` but `GaussianState` doesn't do the same for `covar` to check if it's already a `CovarianceMatrix`. Would be good to add similar logic to avoid needlessly creating a new instance and fix copying.

OK - I added the check on the covar in GaussianState. I updated the test in WeightedGaussianState to test for this.

stonesoup/types/state.py
```diff
@@ -185,6 +187,48 @@ class WeightedGaussianState(GaussianState):
...
+    @classmethod
+    def from_gaussian_state(cls, gaussian_state, *args, copy=True, **kwargs):
+        """
+        Returns a WeightedGaussianState instance based on the gaussian_state.
+
+        Parameters
+        ----------
+        gaussian_state : TYPE
+            DESCRIPTION.
+        *args : See main :class:`~.WeightedGaussianState`
```
sdhiscocks, line 206: `*` needs to be escaped `\*`

DaveKirkland: For the documentation would you prefer the parameters to be listed as just "args" and "kwargs" rather than "*args" and "**kwargs", or would you prefer them with the "*"s and then escaped?

sdhiscocks: I think it would be good to have `*args` and `**kwargs`.

DaveKirkland: So, if I escape them then the Sphinx documentation works out, but pyflake complains that there is an invalid escape sequence '\\*'  - Oh the joy :) Do we just want to live with the lint error?

sdhiscocks: Ah, that'll be because you need to escape the escape `\\*`. You need to escape the Python string, so it can get to the RST parser.

DaveKirkland: Ok. I used an r string instead of escaping the escape. It's a bit cleaner. I pushed the changes. Let me know if there are any further issues.

stonesoup/types/state.py
```diff
@@ -185,6 +187,48 @@ class WeightedGaussianState(GaussianState):
...
+
+    @classmethod
+    def from_gaussian_state(cls, gaussian_state, *args, copy=True, **kwargs):
+        r"""
+        Returns a WeightedGaussianState instance based on the gaussian_state.
+
+        Parameters
+        ----------
+        gaussian_state : :class:`~.GaussianState`
+            The guassian_state use to created the new WeightedGaussianState.
```
oharrald-Dstl, line 205:
```suggestion
            The guassian_state used to create the new WeightedGaussianState.
```

stonesoup/types/tests/test_state.py
```diff
@@ -99,9 +99,27 @@ def test_weighted_gaussian_state():
...
+    assert np.array_equal(gs.covar, wgs.covar)
+    assert gs.timestamp == wgs.timestamp
+    assert weight == wgs.weight
+    assert wgs.state_vector is not gs.state_vector
+    assert wgs.covar is not gs.covar
+
+    # Test copy flag
+    wgs = WeightedGaussianState.from_gaussian_state(gaussian_state=gs, copy=False)
+    assert wgs.state_vector is gs.state_vector
+    assert wgs.covar is gs.covar
```
oharrald-Dstl, line 123: Adding a test to call on the `gaussian_state` property will increase code coverage

DaveKirkland: Done.

oharrald-Dstl (approved):

jswright-dstl (approved):

PR #323, opened 2020-11-05
WIP: Add handling of various prior types predictors/updater

sglvladi:
This PR aims to add the ability for predictors/updaters to produce the appropriate Prediction/Update, depending on the type of the provided prior/prediction.

The code in this PR has stemmed from discussions with @jswright-dstl and @sgmranso. The idea is that if a predictor is provided with a prior that is not of a "base" state (e.g. GaussianState), but rather a subclass (e.g. WeightedGaussianState), the generated prediction should be of a prediction type that matched the type of the prior. The predictor will only modify the properties that it needs to, and will preserve all other properties that are members of the identified prediction type.

So far, only the code for the KalmanPredictor (and its subclasses has been coded), but the same concept can be applied to other predictors and updaters.

### DIFF
--- a/stonesoup/predictor/kalman.py
+++ b/stonesoup/predictor/kalman.py
@@ -44,9 +44,6 @@ class KalmanPredictor(Predictor):
         doc="The control model to be used. Default `None` where the predictor "
             "will create a zero-effect linear :class:`~.ControlModel`.")
 
-    # This attribute tells the :meth:`predict()` method what type of prediction to return
-    _prediction_class = GaussianStatePrediction
-
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


stonesoup/predictor/kalman.py
```diff
@@ -413,8 +442,20 @@ class SqrtKalmanPredictor(KalmanPredictor):
...
+
+        # Get prediction type
+        pred_class = self._prediction_class(prior)
+        kwargs = {prop: getattr(prior, prop)
+                  for prop in pred_class.__dict__['_properties'].keys()}
+        kwargs['state_vector'] = state_vector
+        kwargs['sqrt_covar'] = covar
+        kwargs['timestamp'] = timestamp
+
+        return pred_class(**kwargs)
```
jmbarr, line 458: This looks like repeated code. Could just exist in parent class?

jswright-dstl: It was written like this because the `SqrtKalmanPredictor` requires the keyword argument _sqrt_covar_ over the `KalmanPredictor` needing the argument _covar_.
If we used _sqrt_covar_ in `KalmanPredictor` we would get an unexpected keyword argument type error.

sdhiscocks:
I've create 1a9e5f1e796bfa9e29b23281f1cc168b28addafe, as a slightly different approach which instead maintains a mapping of state to prediction type, to avoid using import lib, and also means it's more broadly compatible with things like Particle states. Also covers measurement predictions if Weighted/Tagged `MeasurementPredictions` variants are created. What are your thoughts on this approach?

sglvladi:
> I've create [1a9e5f1](https://github.com/dstl/Stone-Soup/commit/1a9e5f1e796bfa9e29b23281f1cc168b28addafe), as a slightly different approach which instead maintains a mapping of state to prediction type, to avoid using import lib, and also means it's more broadly compatible with things like Particle states. Also covers measurement predictions if Weighted/Tagged `MeasurementPredictions` variants are created. What are your thoughts on this approach?

Looks neat 👍

stonesoup/predictor/kalman.py
jmbarr: Could just overload this method to replace 'covar' with 'sqrt_covar' via some string variable, then call the method via super.

sdhiscocks:
Closed by alternative PR #329

Issue #138, opened 2019-08-27
Track metadata should handle removal of states


sdhiscocks:
Reference #121

sglvladi:
I think we should proceed with caution on this one, because it can turn out to be really inefficient. Only way I can think of to update the metadata, following a state removal, is to first clear the metadata, and then reconstruct the dict by considering all states, but the one to be removed, incrementally. An example way of doing this would be the following:

``` python
def __delitem__(self, index):
    # Update metadata
    # 1) Reset the metadata
    self.metadata = dict({})
    # 2) Reconstruct the metadata, without considering the state to be removed
    for i in range(len(self)):
        if i == index:
            continue
        self._update_metadata_from_state(self, self.states[i])
    return super().__delitem__(index)
```

Obviously, the above can end up being very costly. 

Therefore, unless we can find an efficient alternative, we may want to consider the following questions:
- Why do we actually need to remove states? Can we come up with some use-cases?
- Assuming there are use-cases, do we also want to allow for the removal of states from the middle of list? If so, what would be an example use-case?
- Assuming we decide that we only allow removal from the ends, should we allow both ends? Use-case?
- If the only use-case is to limit the history length, then should we actually update the metadata once a state is removed? It might be more sensible to just keep the metadata.

PR #328, opened 2020-11-13
Modify tracks to store metadata history

oharrald-Dstl:
Storing a track's metadata history will allow for metadata comparisons in metric generation.
Previously, calling `track.metadata` would return the track's metadata value at the end of its life.


sdhiscocks (approved):

sdhiscocks:
I think this could be modified to cover #138 without too much effort. Noting @sglvladi [comments](https://github.com/dstl/Stone-Soup/issues/138#issuecomment-608990875), being able to modify the track is questionable, but certainly the metadata should reflect modifications whilst this is an option.

nperree-dstl (approved):

sdhiscocks:
On merge, an ID SIAP test fails. Could you rebase and resolve this @oharrald-Dstl?

PR #329, opened 2020-11-13
Add handling of various prior types predictors/updater

sdhiscocks:
Alternative approach to #323

### DIFF
--- a/stonesoup/types/prediction.py
+++ b/stonesoup/types/prediction.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from typing import Any, Union
+
 from ..base import Property
 from .array import CovarianceMatrix
 from .base import Type
@@ -6,7 +8,22 @@ from .state import (State, GaussianState, ParticleState, SqrtGaussianState,
                     TaggedWeightedGaussianState)
 
 
-def _from_state(cls, state, prediction_type=None, **kwargs):
+def _from_state(
+        cls,
+        state: State,
+        prediction_type: Union['Prediction', 'MeasurementPrediction', None] = None,
+        **kwargs: Any) -> Union['Prediction', 'MeasurementPrediction']:
+    """Return new (Measurement)Prediction instance of suitable type using existing properties
+
+    Parameters
+    ----------
+    state: State
+        :class:`~.State` to use existing properties from, and identify prediction type from
+    prediction_type: :class:`~.Update`, optional
+        Type to use for prediction, overriding one from :attr:`class_mapping`.
+    \\*\\*kwargs: Mapping
+        New property names and associate value for use in newly created prediction.
+    """
     try:
         state_type = next(type_ for type_ in type(state).mro() if type_ in cls.class_mapping)
     except StopIteration:
--- a/stonesoup/types/prediction.py
+++ b/stonesoup/types/prediction.py
@@ -6,26 +6,62 @@ from .state import (State, GaussianState, ParticleState, SqrtGaussianState,
                     TaggedWeightedGaussianState)
 
 
+def _from_state(cls, state, prediction_type=None, **kwargs):
+    try:
+        state_type = next(type_ for type_ in type(state).mro() if type_ in cls.class_mapping)
+    except StopIteration:
+        raise TypeError(f'{cls.__name__} type not defined for {type(state).__name__}')
+    if prediction_type is None:
+        prediction_type = cls.class_mapping[state_type]
+
+    # Use current state kwargs that also properties of prediction type
+    new_kwargs = {
+        name: getattr(state, name)
+        for name in state_type.properties.keys() & prediction_type.properties.keys()}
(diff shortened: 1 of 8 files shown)


jmbarr:
> I need to update the doc string. If this is None, it'll use the inbound type (e.g. Weighted/Tagged).

What I meant to ask was what is it about the SqrtKalmanXXXX classes mean you can't use the same logic, i.e. preserve SqrtGaussianXXXX in the same way as you preserve GaussianXXXX or WeightedGaussianXXXX or etc.

stonesoup/types/prediction.py
```diff
@@ -1,31 +1,84 @@
...
+        state: State,
+        prediction_type: Union['Prediction', 'MeasurementPrediction', None] = None,
+        **kwargs: Any) -> Union['Prediction', 'MeasurementPrediction']:
+    """Return new (Measurement)Prediction instance of suitable type using existing properties
+
+    Parameters
+    ----------
+    state: State
+        :class:`~.State` to use existing properties from, and identify prediction type from
+    prediction_type: :class:`~.Update`, optional
```
jmbarr, line 22: Shouldn't this be :class:`~.Prediction`?

sdhiscocks:
> > I need to update the doc string. If this is None, it'll use the inbound type (e.g. Weighted/Tagged).
> 
> What I meant to ask was what is it about the SqrtKalmanXXXX classes mean you can't use the same logic, i.e. preserve SqrtGaussianXXXX in the same way as you preserve GaussianXXXX or WeightedGaussianXXXX or etc.

So in `SqrtGaussian` types, this has a `sqrt_covar` rather than a `covar` being provided. I thought you could provide a non-sqrt prior into the SqrtKalmanPredictor as well, and it would handle it, but looking at the code this isn't the case, which then makes explicitly defining the prediction class type a little redundant.

It'd be nice not to have the special handling for `sqrt_covar` as well. I'll give that some thought.

stonesoup/updater/kalman.py
```diff
@@ -226,9 +227,8 @@ def predict_measurement(self, predicted_state, measurement_model=None,
         meas_cross_cov = self._measurement_cross_covariance(predicted_state, hh)
         innov_cov = self._innovation_covariance(meas_cross_cov, hh, measurement_model)
 
-        return GaussianMeasurementPrediction(pred_meas, innov_cov,
-                                             predicted_state.timestamp,
-                                             cross_covar=meas_cross_cov)
+        return MeasurementPrediction.from_state(
+            predicted_state, state_vector=pred_meas, covar=innov_cov, cross_covar=meas_cross_cov)
```
jmbarr, line 231: The fact that the timestamp isn't passed explicitly fooled me a bit. But the tests pass, and if I follow the logic the timestamp is passed via the predicted state kwargs. Right?

stonesoup/updater/kalman.py
```diff
@@ -280,8 +280,10 @@ def update(self, hypothesis, **kwargs):
             posterior_covariance = \
                 (posterior_covariance + posterior_covariance.T)/2
 
-        return self._update_class(posterior_mean, posterior_covariance, hypothesis,
-                                  timestamp=hypothesis.measurement.timestamp)
+        return Update.from_state(
+            hypothesis.prediction, self._update_class,
+            state_vector=posterior_mean, covar=posterior_covariance,
+            timestamp=hypothesis.measurement.timestamp, hypothesis=hypothesis)
```
jmbarr, line 286: So given previous comment, the timestamp could be taken from the prediction...

stonesoup/updater/kalman.py
sdhiscocks: Yes it is, but could explicit include it to make it clearer.

stonesoup/updater/kalman.py
sdhiscocks: So I tried this without, but some initiators have a prior without a defined timestamp, and which causes an issue. I wonder if this could effect the `predicted_measurement` as well?

sdhiscocks:
> > > I need to update the doc string. If this is None, it'll use the inbound type (e.g. Weighted/Tagged).
> > 
> > 
> > What I meant to ask was what is it about the SqrtKalmanXXXX classes mean you can't use the same logic, i.e. preserve SqrtGaussianXXXX in the same way as you preserve GaussianXXXX or WeightedGaussianXXXX or etc.
> 
> So in `SqrtGaussian` types, this has a `sqrt_covar` rather than a `covar` being provided. I thought you could provide a non-sqrt prior into the SqrtKalmanPredictor as well, and it would handle it, but looking at the code this isn't the case, which then makes explicitly defining the prediction class type a little redundant.
> 
> It'd be nice not to have the special handling for `sqrt_covar` as well. I'll give that some thought.

I've updated the PR with 0535ef2 which allows positional arguments. This solves the issues around `SqrtKalman` states.

jswright-dstl (approved):

jmbarr (approved):

PR #337, opened 2020-11-25
Fix Prediction/Update from_state being called with StateMutableSequence

sdhiscocks:
This fixes the issue per title and some underlying causes relating to a `StateMutableSequence` type check being too specific.


jswright-dstl (approved):

jmbarr (approved):
Fixes sphinx issues, but why wasn't this caught when it was merged first off?

oharrald-Dstl (approved):

sdhiscocks:
> Fixes sphinx issues, but why wasn't this caught when it was merged first off?

Each PR passed tests separately, but merged combination hit edge case. In hindsight given nature of change in #329, a local merge and test would be sensible next time.

PR #344, opened 2020-12-01
Multi-Frame Assignment

sglvladi:
This PR adds a Multi-Frame Assignment hypothesiser and Data Associator. Since it makes use of the MatlabWrapper, the PR depends on #105, which should be merged first.

An example script that runs the algorithm can be found [here](https://gist.github.com/sglvladi/2aeebd831f1264541266f777d00c241b).


sdhiscocks:
Closed in preference of Python version in #568.

PR #346, opened 2020-12-02
Modify gaussian models to raise errors with None covariances

### LINKED ISSUE #343: GuassianModel rvs method generates noise from identity if covariance is None-type
Would it be appropriate to handle `covar = None` as 'noiseless'?
Currently, using the scipy.stats.multivariate_normal.rvs method will default a covariance to the identity matrix. Therefore, when passing in None, noise is generated.

oharrald-Dstl:
Rather than attempting to handle covariance arguments being None on instantiation, since some models have their own covariance methods, it seemed more appropriate to simply raise an error when rvs and pdf methods are called.
Closes #343 
Note: Only models with unmodified covariance methods are affected/tested

### DIFF
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -38,6 +38,8 @@ class CombinedReversibleGaussianMeasurementModel(ReversibleModel, GaussianModel,
         for model in self.model_list:
             if model.ndim_state != self.ndim_state:
                 raise ValueError("Models must all have the same `ndim_state`")
+            if model.covar(**kwargs) is None:
+                raise ValueError("Gaussian models must have defined covariances")
 
     @property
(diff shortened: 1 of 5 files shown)


stonesoup/models/measurement/nonlinear.py
```diff
@@ -38,6 +38,8 @@ def __init__(self, *args, **kwargs):
         for model in self.model_list:
             if model.ndim_state != self.ndim_state:
                 raise ValueError("Models must all have the same `ndim_state`")
+            if model.covar(**kwargs) is None:
```
sdhiscocks, line 41: Not sure this should be called here, certainly not with `kwargs`. I think check in `rvs` and `pdf` is fine.

sdhiscocks (approved):

jswright-dstl (approved):

PR #352, opened 2021-01-04
Modify ID SIAPs to consider track metadata at timestamps

oharrald-Dstl:
Previously, ID SIAP calculations were done by considering `track.metadata` which calls on the metadata value of the track at the end of its life. This value may not reflect the value of the track at a particular timestamp in its lifetime. These changes introduce a new method which will index the track at the required timestamp, attain the relevant metadata, and make ID SIAP calculations using this (along with tests).

### DIFF
--- a/stonesoup/metricgenerator/tracktotruthmetrics.py
+++ b/stonesoup/metricgenerator/tracktotruthmetrics.py
@@ -1335,10 +1335,7 @@ class SIAPMetrics(MetricGenerator):
         state = track[timestamp]
         index = track.index(state)
         metadata = track.metadatas[index]
-        if self.track_id in metadata:
-            return metadata[self.track_id]
-        else:
-            return None
+        return metadata.get(self.track_id)
 
     def _ju_check(self, track, truth, timestamp):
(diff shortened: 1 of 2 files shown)


stonesoup/metricgenerator/tracktotruthmetrics.py
```diff
@@ -1327,12 +1327,21 @@ def _check_j_t(self, manager, timestamp, check_function):
...
-    def _ju_check(self, track, *args):
-        return track.metadata.get(self.track_id) is None
+    def _get_track_id(self, track, timestamp):
+        state = track[timestamp]
+        index = track.index(state)
+        metadata = track.metadatas[index]
+        if self.track_id in metadata:
+            return metadata[self.track_id]
+        else:
+            return None
```
sdhiscocks, line 1341:
```suggestion
        return metadata.get(self.track_id)
```

sdhiscocks (approved):

jswright-dstl (approved):
Agree with Steve's suggested change.

PR #355, opened 2021-01-08
Add Uncertainty Metric

nperree-dstl:
Add an uncertainty metric which calculates the sum of track covariance matrix norms at each time step.


apiszcz:
my 2 cents, 
the uncertainty estimates are helpful, i also use min/max/std/avg for a 'track' metric (aggregated from the track points)

stonesoup/metricgenerator/tests/test_uncertaintymetric.py
```diff
@@ -0,0 +1,92 @@
...
+                                                       timestamp=time_start)])
+              for i in range(5)}
+
+    det_states = generator.extract_states(detections)
+    assert det_states.states == detections
+    track_states = generator.extract_states(tracks)
+    assert set(track_states) == set(t.states[0] for t in tracks)
+    truth_states = generator.extract_states(truths)
+    assert set(truth_states) == set(t.states[0] for t in truths)
+
```
oharrald-Dstl, line 34: Adding a test to make sure error is raised when you extract states of something non-`state` and non-`statemutablesequence` will increase code coverage.

stonesoup/metricgenerator/uncertaintymetric.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+
+        return self.compute_over_time(self.extract_states(manager.tracks))
+
+    @staticmethod
+    def extract_states(object_with_states):
+        """
+        Extracts a list of :class:`~states` from a list of (or single) objects
+        containing states. This method is defined to handle :class:`~track`,
+        :class:`~groundtruthpath` and :class:`~detection` objects
```
oharrald-Dstl, line 41: Rather this is defined to handle sets of `StateMutableSequence` and `State` types.

stonesoup/metricgenerator/tests/test_uncertaintymetric.py
```diff
@@ -0,0 +1,92 @@
...
+                                        covar=np.diag([i, i, i, i]))
+                          for i in range(5)])
+
+    metric = generator.compute_uncertainty(track.states)
+
+    assert metric.title == "Uncertainty Sum"
+    assert metric.value == 20
+    assert metric.timestamp == time
+    assert metric.generator == generator
+
```
oharrald-Dstl, line 52: Adding a test to check whether an error is thrown when the states list passed in to the `compute_uncertainty` method contains more than one timestamp will increase code coverage.

oharrald-Dstl (approved):
A good addition. Once code coverage is increased, all checks should pass.

stonesoup/metricgenerator/tests/test_uncertaintymetric.py
```diff
@@ -0,0 +1,103 @@
...
+                                        covar=np.diag([i, i, i, i]))
+                          for i in range(5)])
+
+    metric = generator.compute_uncertainty(track.states)
+
+    assert metric.title == "Uncertainty Sum"
+    assert metric.value == 20
+    assert metric.timestamp == time
+    assert metric.generator == generator
+    with pytest.raises(ValueError):
```
oharrald-Dstl, line 55:
```suggestion
    with pytest.raises(ValueError, match="All states must be from the same time to compute total uncertainty"):
```

stonesoup/metricgenerator/uncertaintymetric.py
```diff
@@ -0,0 +1,124 @@
...
+
+from .base import MetricGenerator
+from ..types.state import State, StateMutableSequence
+from ..types.metric import SingleTimeMetric, TimeRangeMetric
+from ..types.time import TimeRange
+
+
+class UncertaintyMetric(MetricGenerator):
+    """
+    Computes the covariance matrix of the state for each track as a measure of the
```
jmbarr, line 13: Be clear the metric is computed at a single instance for a collection of states from different tracks. It's the sum of the matrix norms of each state at that instance. The metric generator will return this number for each instance (e.g. each timestep) in the track(s).

nperree-dstl: Would this be better - "Computes the sum of the covariance matrix norms of each state at a time step. The metric generator will return this value at each time step in the track(s) as a measure of the uncertainty." ?

jmbarr (commented):
I know it may seem like a minor quibble, and I should have brought it up earlier, but I think the name isn't descriptive enough. There are all sorts of metrics based on uncertainty estimates, of which this is but one. So we shouldn't hog that broad description. I'd suggest `SumofCovarianceNormsMetric` as close to fitting the bill.

While we're at it we should make it clear that the we use the _Frobenius_ norm. Other matrix norms are available but that's the one that `np.linalg.norm()` returns by default. (We might consider allowing the passing of arguments to `np.linlg.norm()` to access the other options?)

sdhiscocks:
I'd be tempted to revert/drop commit 05ed0de (except names of test functions), as these files will then be ready to also hold other uncertainty based metrics.

jswright-dstl (approved):

PR #365, opened 2021-02-03
Vectorise particle filter

idorrington-dstl:
Vectorised and optimised the particle filter. 

Did this by creating a new particles object where the state_vector, weight, and parent attributes are vectors/arrays corresponding to the same attributes of the individual particles. 

The most significant of the other optimisations made is that the pdf function now optionally outputs the indexes to sort the particles by size order. This saves us running np.argsort on the particle weights in the resample step leading to a  ~35% decrease in run time (though it does make the code slightly harder to follow as the order of the particles is calculated in a different function to where it is used).

### DIFF
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -264,8 +264,11 @@ class CartesianToElevationBearingRange(NonLinearGaussianMeasurement, ReversibleM
 
         # Convert to Spherical
         rho, phi, theta = cart2sphere(*xyz_rot)
+        elevations = [Elevation(i) for i in np.atleast_1d(theta)]
+        bearings = [Bearing(i) for i in np.atleast_1d(phi)]
+        rhos = [i for i in np.atleast_1d(rho)]
 
-        return StateVector([[Elevation(theta)], [Bearing(phi)], [rho]]) + noise
+        return StateVectors([elevations, bearings, rhos]) + noise
 
     def inverse_function(self, detection, **kwargs) -> StateVector:
 
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -414,10 +414,10 @@ class CartesianToBearingRange(NonLinearGaussianMeasurement, ReversibleModel):
                 noise = 0
 
         # Account for origin offset
-        xyz = [list(state.state_vector[self.mapping[0], :] - self.translation_offset[0, 0]),
-               list(state.state_vector[self.mapping[1], :] - self.translation_offset[1, 0]),
-               [0] * state.state_vector.shape[1]
-               ]
+        xyz = np.array([state.state_vector[self.mapping[0], :] - self.translation_offset[0, 0],
+                        state.state_vector[self.mapping[1], :] - self.translation_offset[1, 0],
+                        [0] * state.state_vector.shape[1]
+                        ])
 
         # Rotate coordinates
         xyz_rot = self._rotation_matrix @ xyz
@@ -635,10 +635,10 @@ class Cartesian2DToBearing(NonLinearGaussianMeasurement):
                 noise = 0
 
         # Account for origin offset
-        xyz = [list(state.state_vector[self.mapping[0], :] - self.translation_offset[0, 0]),
-               list(state.state_vector[self.mapping[1], :] - self.translation_offset[1, 0]),
-               [0] * state.state_vector.shape[1]
-               ]
+        xyz = np.array([state.state_vector[self.mapping[0], :] - self.translation_offset[0, 0],
+                        state.state_vector[self.mapping[1], :] - self.translation_offset[1, 0],
+                        [0] * state.state_vector.shape[1]
+                        ])
 
         # Rotate coordinates
         xyz_rot = self._rotation_matrix @ xyz
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -546,7 +546,9 @@ class CartesianToElevationBearing(NonLinearGaussianMeasurement):
         # Convert to Angles
         phi, theta = cart2angles(*xyz_rot)
 
-        return StateVector([[Elevation(theta)], [Bearing(phi)]]) + noise
+        bearings = [Bearing(i) for i in np.atleast_1d(phi)]
+        elevations = [Elevation(i) for i in np.atleast_1d(theta)]
+        return StateVectors([elevations, bearings]) + noise
 
     def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
         out = super().rvs(num_samples, **kwargs)
@@ -929,11 +931,14 @@ class CartesianToElevationBearingRangeRate(NonLinearGaussianMeasurement, Reversi
         xyz_vel = state.state_vector[self.velocity_mapping, :] - self.velocity
 
         # Use polar to calculate range rate
-        rr = np.dot(xyz_pos[:, 0], xyz_vel[:, 0]) / np.linalg.norm(xyz_pos)
-        return StateVector([[Elevation(theta)],
-                            [Bearing(phi)],
-                            [rho],
-                            [rr]]) + noise
+        rr = np.einsum('ij,ij->j', xyz_pos, xyz_vel) / np.linalg.norm(xyz_pos, axis=0)
+
+        bearings = [Bearing(i) for i in np.atleast_1d(phi)]
+        elevations = [Elevation(i) for i in np.atleast_1d(theta)]
+        return StateVectors([elevations,
+                             bearings,
+                             np.atleast_1d(rho),
+                             rr]) + noise
 
     def inverse_function(self, detection, **kwargs) -> StateVector:
         theta, phi, rho, rho_rate = detection.state_vector
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -412,11 +412,10 @@ class CartesianToBearingRange(NonLinearGaussianMeasurement, ReversibleModel):
 
         # Account for origin offset
         n = state.state_vector.shape[1]
-        xyz = [list(state.state_vector[self.mapping[0], :n]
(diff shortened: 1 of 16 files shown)


sdhiscocks:
> The most significant of the other optimisations made is that the pdf function now optionally outputs the indexes to sort the particles by size order. This saves us running np.argsort on the particle weights in the resample step leading to a ~35% decrease in run time (though it does make the code slightly harder to follow as the order of the particles is calculated in a different function to where it is used).

I took a look at this and agree that it's harder to read, and also would be better not too pass weights orders around. Out of interest, I tried using the log values to sort, to avoid overhead of the custom type, and then pure NumPy C optimised code could do the sort. I found applying the following to f4a1966 yielded similar gains (which was a little unexpected TBH).

```diff
diff --git a/stonesoup/resampler/particle.py b/stonesoup/resampler/particle.py
index bb275766..77c11eb1 100644
--- a/stonesoup/resampler/particle.py
+++ b/stonesoup/resampler/particle.py
@@ -26,7 +26,8 @@ class SystematicResampler(Resampler):
             particles = Particles(particle_list=particles)
         n_particles = len(particles)
         weight = Probability(1/n_particles)
-        weight_order = np.argsort(particles.weight, kind='stable')
+        weights = np.array([weight.log_value for weight in particles.weight])
+        weight_order = np.argsort(weights, kind='stable')
         cdf = [v.log_value for v in np.cumsum([particles.weight[i] for i in weight_order])]

         # Pick random starting point
```

stonesoup/resampler/particle.py
```diff
@@ -23,25 +22,24 @@ def resample(self, particles):
...
-                Particle(particle.state_vector,
-                         weight=weight,
-                         parent=particle))
-
+        u_j = u_i + (1 / n_particles) * np.arange(n_particles)
+        index = weight_order[np.searchsorted(cdf, np.log(u_j))]
+        new_particles = Particles(state_vector=particles.state_vector[:, index],
+                                  weight=[weight]*n_particles,
+                                  parent=Particles(state_vector=particles.state_vector[:, index],
+                                                   weight=[weight]*n_particles))
```
sdhiscocks, line 44: This should be the parents weights.
```suggestion
                                                   weight=particles.weight[index]))
```

stonesoup/types/particle.py
```diff
@@ -25,3 +29,71 @@ def __init__(self, state_vector, weight, parent=None, *args, **kwargs):
...
+class Particles(Type):
+    """
+    Particle type
+
+    A collection of particles. Contains a state and weight for each particle
+    """
+    state_vector: StateVectors = Property(default=None, doc="State vectors of particles")
+    weight: MutableSequence[Probability] = Property(default=None, doc='Weights of particles')
+    parent: 'Particles' = Property(default=None, doc='Parent particles')
+    particle_list: MutableSequence[Particle] = Property(default=None,
```
jmbarr, line 43: Why do you need a collection of `StateVector`s and a list of `Particle`s? Could we do away with `Particle` altogether?

idorrington-dstl: The list of particles was for backwards compatibility. In particular many of the unit tests loop over lists of particles so this was a good sanity check. I guess now I have finished vectorising and am confident I've not changed the output we could get rid of Particle and edit all the unit tests to use the new class?

sdhiscocks: So little on the fence either way with this one. The only thought I had, is there may be cases where you can't easily vectorise a function or operation, and therefore being able to loop over particles (as `Particle` states) to run a function/calculation, and use that list to repopulate the `Particles`, could be useful.

We probably should encourage people not to do this however, so should update our code/tests.

stonesoup/types/particle.py
```diff
@@ -25,3 +29,71 @@ def __init__(self, state_vector, weight, parent=None, *args, **kwargs):
...
+        particle = Particle(state_vector=self.state_vector[:, item],
+                            weight=self.weight[item],
+                            parent=p)
+        return particle
+
+    def __len__(self):
+        return self.state_vector.shape[1]
+
+    def __iter__(self):
+        return self
```
sdhiscocks, line 87: Not sure we should make this class an iterator, but instead return a generator (so make the class iterable still). Making this an iterator, means that you can't iterate over this in parallel, due to having to use an `_index` variable. 

Something like:
```python
def __iter__(self):
    yield from (self[i] for i in range(len(particles))
```
and drop the `__next__` method and `_index`.

stonesoup/updater/particle.py
```diff
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+import numpy as np
```
sdhiscocks, line 2: Import should be below standard library.

sdhiscocks (changes requested):
Great contribution, with significant performance improvement. A few suggestions and think one model missing being vectorised.

Would also be good to squash commits down as there a number of changes that were done and then undone, and rebase onto `master`.

stonesoup/types/particle.py
idorrington-dstl: This was my first attempt at making it also work just like a list of particles (for testing). In the end I added getitem, so I'll delete the iter and next methods.

sdhiscocks (approved):

stonesoup/models/measurement/nonlinear.py
```diff
@@ -264,8 +264,11 @@ def function(self, state, noise=False, **kwargs) -> StateVector:
 
         # Convert to Spherical
         rho, phi, theta = cart2sphere(*xyz_rot)
+        elevations = [Elevation(i) for i in np.atleast_1d(theta)]
+        bearings = [Bearing(i) for i in np.atleast_1d(phi)]
+        rhos = np.atleast_1d(rho)
 
-        return StateVector([[Elevation(theta)], [Bearing(phi)], [rho]]) + noise
+        return StateVectors([elevations, bearings, rhos]) + noise
```
oharrald-Dstl, line 271: Just incase I'm confusing myself here, the StateVectors type will hstack these to have elevation, bearing, rho rows?

idorrington-dstl: That's right. So the 0 index gives another StateVectors object with all the elevations, 1 gives bearings, and 2 gives rhos.

stonesoup/types/numeric.py
```diff
@@ -231,16 +233,15 @@ def log(self):
     @classmethod
     def sum(cls, values):
         """Carry out LogSumExp"""
-        log_values = [cls._log(value) for value in values]
-        if not log_values:
+        log_values = np.array([cls._log(value) for value in values])
+        if not len(log_values):
```
oharrald-Dstl, line 237: Very minor comment here: Could this be changed to `if len(log_values) == 0` for the sake of readability?

stonesoup/types/particle.py
```diff
@@ -25,3 +30,58 @@ def __init__(self, state_vector, weight, parent=None, *args, **kwargs):
...
+    state_vector: StateVectors = Property(default=None, doc="State vectors of particles")
+    weight: MutableSequence[Probability] = Property(default=None, doc='Weights of particles')
+    parent: 'Particles' = Property(default=None, doc='Parent particles')
+    particle_list: MutableSequence[Particle] = Property(default=None,
+                                                        doc='List of Particle objects')
+
+    def __init__(self, state_vector=None, weight=None, parent=None, particle_list=None,
+                 *args, **kwargs):
+        if state_vector is not None:
+            assert particle_list is None,\
```
oharrald-Dstl, line 50: This should probably raise an error instead of attempting to assert first.

stonesoup/types/particle.py
```diff
@@ -25,3 +30,58 @@ def __init__(self, state_vector, weight, parent=None, *args, **kwargs):
...
+                "Particles object cannot use both state_vector and particle_list"
+
+        if particle_list and isinstance(particle_list, list):
+            state_vector = StateVectors([particle.state_vector for particle in particle_list])
+            weight = np.array([Probability(particle.weight) for particle in particle_list])
+            parent_list = [particle.parent for particle in particle_list
+                           if particle.parent is not None]
```
oharrald-Dstl, line 57: would we not want to keep `None` where a particle has no parent? Or is a misalignment between parents and children not an issue?

idorrington-dstl: Good point. I don't think that situation should ever come up, but the code can definitely be clearer. 

@sdhiscocks  Is there ever a case when some particles would have parents and others wouldn't? My code doesn't like being given None as a particle, so I'll need to handle that if it can happen. If not then I'll change this section of the code to check that either all particles have parents or none do and handle appropriately.

sdhiscocks: I can't think of a scenario, although we don't have any practical use of parent yet anyway.  As you suggest, something like checking the parent list is either empty (all `None`) or correct size (all not `None`) and raising a `ValueError` otherwise is a good approach.

oharrald-Dstl (approved):

stonesoup/resampler/particle.py
```diff
@@ -28,7 +28,7 @@ def resample(self, particles):
         weight = Probability(1/n_particles)
         log_weights = np.array([weight.log_value for weight in particles.weight])
         weight_order = np.argsort(log_weights, kind='stable')
-        cdf = [v.log_value for v in np.cumsum(particles.weight[weight_order])]
+        cdf = np.log(np.cumsum(np.exp(log_weights[weight_order])))
```
idorrington-dstl, line 31: @sdhiscocks @oharrald-Dstl  I changed this line. I don't think it should cause any problems, as we take the log value from the probability object to begin with, but it makes it about 10% faster.

stonesoup/models/measurement/tests/test_models.py
```diff
@@ -626,6 +627,225 @@ def fun(x):
...
+    meas_pred_wo_noise = model.function(state)
+    ndim_meas = np.shape(meas_pred_wo_noise)[0]
+    eval_m = h(single_state_vec,
+               model.mapping,
+               model.velocity_mapping,
+               model.translation_offset,
+               model.rotation_offset,
+               model.velocity)
+    print(meas_pred_wo_noise)
+    print(eval_m)
```
sdhiscocks, line 740: Just spotted these, that need removal.

PR #366, opened 2021-02-03
Add ID based associator

nperree-dstl:
Create an ID based associator which associates tracks to truths with the same ID.

### DIFF
--- a/stonesoup/dataassociator/tracktotrack.py
+++ b/stonesoup/dataassociator/tracktotrack.py
@@ -344,10 +344,8 @@ class TrackIDbased(TrackToTrackAssociator):
                                                                truth[0].timestamp),
                                                            min(track[-1].timestamp,
                                                                truth[-1].timestamp))))
-                    except TypeError:
-                        associations.add(Association((track, truth)))
-
-                    except ValueError:
+                    except (TypeError, ValueError):
+                        # A timestamp is None, or non-overlapping timestamps (start > end)
                         associations.add(Association((track, truth)))
 
         return AssociationSet(associations)
--- a/stonesoup/dataassociator/tracktotrack.py
+++ b/stonesoup/dataassociator/tracktotrack.py
@@ -4,7 +4,7 @@ from operator import attrgetter
 from .base import TrackToTrackAssociator
 from ..base import Property
 from ..measures import Measure, Euclidean
-from ..types.association import AssociationSet, TimeRangeAssociation
+from ..types.association import AssociationSet, TimeRangeAssociation, Association
 from ..types.time import TimeRange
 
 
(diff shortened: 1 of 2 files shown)


stonesoup/dataassociator/tracktotrack.py
```diff
@@ -302,3 +302,45 @@ def associate_tracks(self, tracks_set, truth_set):
...
+        associations = set()
+
+        for track in tracks_set:
+            for truth in truths_set:
+                if track.id == truth.id:
+                    associations.add(TimeRangeAssociation((track, truth),
+                                                          TimeRange(truth[0].timestamp,
+                                                                    truth[-1].timestamp)))
+                else:
+                    continue
```
sdhiscocks, line 344: This isn't required.

jswright-dstl (approved):

stonesoup/dataassociator/tracktotrack.py
```diff
@@ -302,3 +302,43 @@ def associate_tracks(self, tracks_set, truth_set):
...
+                   Contains a set of :class:`~.Association` objects
+
+               """
+
+        associations = set()
+
+        for track in tracks_set:
+            for truth in truths_set:
+                if track.id == truth.id:
+                    associations.add(TimeRangeAssociation((track, truth),
```
jmbarr, line 340: Is this a `TimeRangeAssociation`, or is it being used because we don't have any other type? In principle, one might have tracks with no timestamps in which case "TimeRange" seems meaningless (though I don't think it'll error).

Elsewhere 'TimeRangeAssociation` seems to be for tracks that are only associated at particular times, and not elsewhere. This associator actually ignores timesteps but does attach the complete truth time range, which might not be correct for the track. Suggest we find, or create a simpler `Association` object.

jmbarr: Could make this an `AssociationPair`, assuming unique IDs.

jmbarr:
Should mention that, although it would be assumed, there's no checking for uniqueness in truth or track IDs. If there are repeated IDs then there'll be multiple associations.

stonesoup/dataassociator/tracktotrack.py
jmbarr: But maybe not, while there may be unique truth, the tracks could be discontinuous (like with occlusion), you'd still like to associate the fragments with the same ID to the correct truth. So recommend `Association` then I suppose.

nperree-dstl: I think I used `TimeRangeAssociation` as I thought that was what you need for the SIAP metrics - on a second look I don't think that's true so I'll look at using `Association` instead.

sdhiscocks: It may be okay to use the time range association, but just use the track times, rather than the truth times? Or even the overlapping time (`max(track[0].timestamp, truth.timestamp[0])` and `min(track[-1].timestamp, truth[-1].timestamp`)

oharrald-Dstl (approved):

stonesoup/dataassociator/tracktotrack.py
```diff
@@ -302,3 +302,52 @@ def associate_tracks(self, tracks_set, truth_set):
...
+                            TimeRangeAssociation((track, truth),
+                                                 TimeRange(max(track[0].timestamp,
+                                                               truth[0].timestamp),
+                                                           min(track[-1].timestamp,
+                                                               truth[-1].timestamp))))
+                    except TypeError:
+                        associations.add(Association((track, truth)))
+
+                    except ValueError:
+                        associations.add(Association((track, truth)))
```
sdhiscocks, line 351:
```suggestion
                    except (TypeError, ValueError):
                        # A timestamp is None, or non-overlapping timestamps (end > start)
                        associations.add(Association((track, truth)))
```

oharrald-Dstl: If timestamps are non-overlapping, isn't that end < start?

sdhiscocks: Doh! 🤦

sdhiscocks: > If timestamps are non-overlapping, isn't that end < start?

Fixed

PR #367, opened 2021-02-03
Move functions and orbital functions to sub-module

sdhiscocks:
This will enable future functions to be grouped logically, and may also consider breaking up existing functions in future.

### DIFF
--- a/stonesoup/functions/__init__.py
+++ b/stonesoup/functions/__init__.py
@@ -4,8 +4,8 @@
 import numpy as np
 from copy import copy
 
-from .types.numeric import Probability
-from .types.array import StateVector, StateVectors, CovarianceMatrix
+from ..types.numeric import Probability
+from ..types.array import StateVector, StateVectors, CovarianceMatrix
 
 
 def tria(matrix):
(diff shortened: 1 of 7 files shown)


jswright-dstl (approved):

stonesoup/functions/__init__.py
```diff
@@ -4,8 +4,8 @@
 import numpy as np
 from copy import copy
 
-from .types.numeric import Probability
-from .types.array import StateVector, StateVectors, CovarianceMatrix
+from ..types.numeric import Probability
+from ..types.array import StateVector, StateVectors, CovarianceMatrix
 
 
 def tria(matrix):
```
jmbarr, line 11: Is the intent to move these functions out to separate files too?

jmbarr (approved):
Worth thinking about whether it makes sense to group some of the existing functions into files, rather than having them in __init__.py.

sdhiscocks:
> Worth thinking about whether it makes sense to group some of the existing functions into files, rather than having them in __init__.py. 

Yeah. Thought it worth doing this change now, as shouldn't break anything existing, an orbital states was just merged. Then we can spend some time thinking about logical grouping for other functions.

PR #375, opened 2021-02-09
Add support for plugins

svidal-dstl:
Add ability to load plugins using package metadata in package's setup.py [entry_points](https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata). This will enable third party Stone Soup components to be easily discoverable.

### DIFF
--- a/stonesoup/plugins.py
+++ b/stonesoup/plugins.py
@@ -35,16 +35,8 @@ import importlib.util
 for entry_point in pkg_resources.iter_entry_points('stonesoup.plugins'):
     try:
         name = entry_point.name
-        module_name = entry_point.module_name
-
-        module = importlib.util.find_spec(module_name)
-        module_spec = importlib.util.spec_from_loader(name=module.name,
-                                                      loader=module.loader,
-                                                      origin=module.origin)
-
         plugin_module = f'{__name__}.{name}'
-
-        sys.modules[plugin_module] = importlib.util.module_from_spec(module_spec)
+        sys.modules[plugin_module] = entry_point.load()
 
     except (ImportError, ModuleNotFoundError) as e:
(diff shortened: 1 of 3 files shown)


sdhiscocks (approved):
Looks good, and I've tested with a dummy plugin.

One issue is this is lacking tests, but I'm not sure how we'd approach that, as the entry points would need to be present; maybe mock with monkeypatching. Suggest we raise this as separate issue for future change.

stonesoup/plugins.py
```diff
@@ -0,0 +1,50 @@
...
+        module_name = entry_point.module_name
+
+        module = importlib.util.find_spec(module_name)
+        module_spec = importlib.util.spec_from_loader(name=module.name,
+                                                      loader=module.loader,
+                                                      origin=module.origin)
+
+        plugin_module = f'{__name__}.{name}'
+
+        sys.modules[plugin_module] = importlib.util.module_from_spec(module_spec)
```
sdhiscocks, line 47: Could you simplify this to the below?
```suggestion
        plugin_module = f'{__name__}.{name}'
        sys.modules[plugin_module] = entry_point.load()
```

stonesoup/plugins.py
```diff
@@ -0,0 +1,50 @@
...
+    from stonesoup.plugins.my_plugin import MyClass
+
+.. note::
+    When developing plugins for Stone Soup, :attr:`entry_points` must be associated with
+    the :attr:`stonesoup.plugins` key in the :attr:`entry_points` dictionary.
+"""
+import sys
+import pkg_resources
+import warnings
+import importlib.util
```
sdhiscocks, line 33: Would also remove this if taking my other suggestion.
```suggestion
```

oharrald-Dstl (approved):

PR #377, opened 2021-02-10
Add example of using a Hidden Markov Model to track classification

oharrald-Dstl:
Details are available on the actual sphinx gallery example.

Note:
- Aim is to demonstrate using a hidden Markov model to classify a target using Stone Soup components
- Further sphinx gallery examples are to be added


docs/examples/Naive_Classification.py
```diff
@@ -0,0 +1,188 @@
...
+# --------
+# We plot the track classification as a stacked bar graph. Green indicates bicycle, white car, and
+# red bus. The larger a particular bar, the greater the probability that the track at that time
+# is of the corresponding classification.
+# Observations are plotted underneath this graph. Light blue indicates an observation that the
+# target is small, blue that it is medium, and dark blue that it is large.
+
+import matplotlib.pyplot as plt
+
+t_a, t_n, t_e = np.array([list(state.state_vector) for state in track]).T
```
nperree-dstl, line 161: Just wondering - what do the a, n and e in  `t_a, t_n, t_e` stand for?

docs/examples/Naive_Classification.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python
+# coding: utf-8
+
```
jmbarr, line 3: The notebook needs to have a broader, more informative title. Why not use the one above, or something similar: "using a Hidden Markov Model for classification", perhaps?

Forward algorithm isn't specific enough (there are a number of forward algorithms). A bit of explanatory text, below the title, saying what we're up to and what this forward algorithm is being used for.

docs/examples/Naive_Classification.py
```diff
@@ -0,0 +1,188 @@
...
+
+# %%
+# Modelling state transition
+# --------------------------
+# We require a model of the state transition in order to track the target's classification.
+# This will be a simple linear map, represented by a :math:`3\times 3` matrix with no noise added.
+# We might not know that the target does not switch classification, so the matrix should not simply
+# be the identity matrix.
+#
```
jmbarr, line 41: This contradicts what's stated above ("ie. if it is a bus, it stays a bus"). 

It would be sensible to construct a model where the transition is the identity matrix, but noise can be added to account for the spontaneous transformation of a bus into a bike (presumably a very small amount in this instance. Of course the noise can't be merely additive, but you can use the `noise=True` flag to sample from a Categorical distribution with specified probabilities.

You could then use `noise=False` which would ensure that

docs/examples/Naive_Classification.py
```diff
@@ -0,0 +1,188 @@
...
+              [0, 0.1, 0.9]])
+
+
+def transit(state):
+    return F.T @ state.state_vector
+
+
+# %%
+# Modelling observations
+# ----------------------
```
jmbarr, line 64: I think it's important to state that we don't observe the state directly (hence 'hidden'). We observe 'emissions' governed by a matrix of probabilities `E`. (We could add noise to this but again, bit more complicated than additive.)

docs/examples/Naive_Classification.py
```diff
@@ -0,0 +1,188 @@
...
+# 	    \end{bmatrix}
+#
+
+import scipy
+
+E = np.array([[0.89, 0.1, 0.01],
+              [0.3, 0.3, 0.4],
+              [0.1, 0.3, 0.6]])
+
+# ie. 89% of bicycles are small, 10% are medium, 1% are large.
```
jmbarr, line 89: It would be more accurate to say that bicycles return measurements = small 89% of the time, medium 10% and large 1%. That might be because 1% of bicycles are large or (more likely) because sensor isn't perfect.

docs/examples/Naive_Classification.py
```diff
@@ -0,0 +1,188 @@
...
+from stonesoup.types.detection import Detection
+
+observations = []
+for i in range(0, nsteps):
+    observations.append(Detection(observe(gt[i])))
+
+
+# %%
+# Tracking classification
+# -----------------------
```
jmbarr, line 120: This merits a bit more explanation. (See https://en.wikipedia.org/wiki/Forward_algorithm for a start.) In particular it would be good to make connection between the predict-update recursion in other Stone Soup filters and the transition-emission probabilities here.

docs/examples/Naive_Classification.py
oharrald-Dstl: Just some very bad variable naming from me. Since Id changed the original class names, they don't mean anything now

oharrald-Dstl:
I've realised the emission matrix is used incorrectly in generating measurements. Will correct this.

oharrald-Dstl:
Closed due to #460 having a more complete tutorial/example.

PR #382, opened 2021-02-19
Fix plotter metric so it uses Stone Soup Plotter class

nperree-dstl:
Change how plotter metric works so it uses the Stone Soup plotting functions from the Plotter class

### DIFF
--- a/stonesoup/metricgenerator/plotter.py
+++ b/stonesoup/metricgenerator/plotter.py
@@ -42,17 +42,21 @@ class TwoDPlotter(PlotGenerator):
         return metric
 
     def plot_tracks_truth_detections(self, tracks, groundtruth_paths,
-                                     detections):
+                                     detections, uncertainty=False, particle=False):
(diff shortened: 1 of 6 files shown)


svidal-dstl (approved):

jswright-dstl (approved):

stonesoup/metricgenerator/plotter.py
```diff
@@ -46,92 +42,60 @@ def compute_metric(self, manager, *args, **kwargs):
         return metric
 
     def plot_tracks_truth_detections(self, tracks, groundtruth_paths,
-                                     detections):
+                                     detections, uncertainty=False, particle=False):
```
sdhiscocks, line 45: So `uncertainty` and `particle` should be `Property`s on the class as well, passed when calling this method from `TwoDPlotter.compute_metric()`.

sdhiscocks (changes requested):

sdhiscocks (approved):

PR #386, opened 2021-02-23
Update StateMutableSequence error handling

etfrogers-dstl:
This PR tweaks error handling in `StateMutableSequence`

Before this change, any class that inherited from `StateMutableSequence` (e.g. `Platform` would give the following behaviour:

```python
platform = Platform(some_parameters)
platform.missing_attribute
```
```
AttributeError: 'State' object has no attribute 'missing_attribute'
```

This is odd behaviour, as `Platform` is not a `State` and caused me to spend an annoying amount of time debugging the problem.

After the PR, the error message would be:
```
AttributeError: 'Platform' object has no attribute 'missing_attribute'
```
This is much more helpful and seems like a sensible change at the expense of only a little more code.

### DIFF
--- a/stonesoup/types/state.py
+++ b/stonesoup/types/state.py
@@ -111,7 +111,15 @@ class StateMutableSequence(Type, abc.MutableSequence):
                 "{!r} object has no attribute {!r}".format(
                     type(self).__name__, item))
         else:
-            return getattr(self.state, item)
+            try:
+                return getattr(self.state, item)
+            except AttributeError as err:
+                if str(err).startswith("'State' object has no attribute"):
+                    raise AttributeError(
+                        "{!r} object has no attribute {!r}".format(
+                            type(self).__name__, item))
+                else:
+                    raise err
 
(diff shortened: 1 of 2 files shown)


etfrogers-dstl:
I should note that this change isn't perfect, as it still swallows any other `AttributeError`s generated by sub-classes (example available on request), but it improves things significantly.

I can't find a way of fixing the whole problem (and ideas welcome!), and this is a good start.

sdhiscocks (approved):

oharrald-Dstl (approved):
Looks good. May be worth adding a test now that you're raising an error with custom message.

etfrogers-dstl:
> Looks good. May be worth adding a test now that you're raising an error with custom message.

Good idea to add a test, and I now have. This test would have failed before the change. 

I also improved the commenting of ``__getattr__`` to explain that this isn't actually a custom error message (or rather, it's _trying_ not to be). It's "restoring" the original error message as much as possible.

PR #394, opened 2021-03-01
Plotter with track labels

jmbarr:
Modified plotter to accept track labels as keyword arguments. Adjusted labels to preserve line colour in the legend.

### DIFF
--- a/stonesoup/plotter.py
+++ b/stonesoup/plotter.py
@@ -42,7 +42,7 @@ class Plotter:
         self.handles_list = []
         self.labels_list = []
 
-    def plot_ground_truths(self, truths, mapping, **kwargs):
+    def plot_ground_truths(self, truths, mapping, truths_label="Ground Truth", **kwargs):
         """Plots ground truth(s)
 
         Plots each ground truth path passed in to :attr:`truths` and generates a legend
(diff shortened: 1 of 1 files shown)


jswright-dstl (approved):

stonesoup/plotter.py
```diff
@@ -202,14 +202,17 @@ def plot_tracks(self, tracks, mapping, uncertainty=False, particle=False, **kwar
...
-            self.ax.plot([state.state_vector[mapping[0]] for state in track],
-                         [state.state_vector[mapping[1]] for state in track],
-                         **tracks_kwargs)
+            line = self.ax.plot([state.state_vector[mapping[0]] for state in track],
+                                [state.state_vector[mapping[1]] for state in track],
+                                **tracks_kwargs)
+
+        # Assuming a single track or all plotted as the same colour then the following will work.
+        # Otherwise will just render the final track colour.
+        tracks_kwargs['color'] = plt.getp(line[0], 'color')
```
nperree-dstl, line 211: Not sure if this is an issue worth fixing but this means uncertainty ellipses are also plotted as the same colour as the final track. Could be fixed by making all ellipses a default colour?

nperree-dstl (approved):

stonesoup/plotter.py
jmbarr: It's a good thing that the ellipses match the track colour, no?

nperree-dstl: Yes, sorry I meant when there are multiple tracks with different colours all the ellipses will have the colour of the final track, not the track they are attached to.

jmbarr: Oh, I see. That might look a bit odd. 

I imagine there must be a way of finding out, ahead of time or separately, the order in which colours appear, and then passing that to the legend/ellipse plotter.

sdhiscocks: You could use the same `plt.getp(line[0], 'color')` in the loop above, creating a dictionary of colours keyed by the track, and then use that for the ellipse plotting below.

https://github.com/dstl/Stone-Soup/pull/394/files#diff-40bed2921b0b0211da770554c467db7cc29b9e0f0fa0c436191a7f1b50f95458R232

jmbarr: That link's not taking me anywhere...

sdhiscocks: Yeah, not working for me now either. Was just linking to the line below where the color is passed for the ellipse.

https://github.com/dstl/Stone-Soup/blob/0fc6be4054da5b63c388517b51cb224cfa5186a5/stonesoup/plotter.py#L232

PR #398, opened 2021-03-02
Alpha-Beta updater

jmbarr:
The Alpha-Beta filter

Example: https://gist.github.com/jmbarr/4e632ea56e9bb1980935a9d79a933b52

Designed to close off the work started by Jack Harris. Provides a single class, `AlphaBetaUpdater`, derived from `Updater`. See documentation for full details.


stonesoup/updater/alphabeta.py
```diff
@@ -0,0 +1,115 @@
...
+
+        Returns
+        -------
+         : :class:`~.StateUpdate`
+            The updated state
+        """
+        out_statevector = hypothesis.prediction.state_vector.copy()
+        pred_meas = self.predict_measurement(hypothesis.prediction)
+
+        pmap = np.array(self.measurement_model.mapping)
```
sdhiscocks, line 103: I'm not sure this is always true, as some models may map velocity.  I think you could add `pmap` as a `Property` similar to `vmap`, defaulting to `None` and use this as the default.

stonesoup/updater/tests/test_alphabeta.py
```diff
@@ -0,0 +1,71 @@
...
+
+    eval_posterior_position = prediction.state_vector[[0, 2]] + \
+        alpha * (measurement.state_vector - eval_measurement_prediction.state_vector)
+    eval_posterior_velocity = prediction.state_vector[[1, 3]] + \
+        beta/timediff.total_seconds() * (measurement.state_vector -
+                                         eval_measurement_prediction.state_vector)
+
+    eval_state_vect = np.concatenate((eval_posterior_position, eval_posterior_velocity))
+
+    print(eval_state_vect)
```
sdhiscocks:
```suggestion
```

stonesoup/updater/alphabeta.py
```diff
@@ -0,0 +1,115 @@
+# -*- coding: utf-8 -*-
+
+
+import numpy as np
+
+from stonesoup.base import Property
+from stonesoup.updater import Updater
+from stonesoup.types.prediction import MeasurementPrediction
+from stonesoup.types.update import Update
```
sdhiscocks, line 9: For consistency with rest of Stone Soup, these should be relative imports.

stonesoup/updater/alphabeta.py
```diff
@@ -0,0 +1,115 @@
...
+        time_interval : :class:`~.timedelta`
+            The time interval over which the prediction has been made.
+
+        Returns
+        -------
+         : :class:`~.StateUpdate`
+            The updated state
+        """
+        out_statevector = hypothesis.prediction.state_vector.copy()
+        pred_meas = self.predict_measurement(hypothesis.prediction)
```
sdhiscocks, line 101: Should check for existing measurement prediction, and measurement's measurement model on hypothesis.

sdhiscocks (changes requested):

stonesoup/updater/alphabeta.py
jmbarr: It should always be true in the alpha-beta filter, as you'd only choose a linear measurement model which returned position. I suppose there might be possible to construct a linear measurement model which doesn't have mapping, but it wouldn't inherit from `MeasurementModel`.

But could still do your suggestion.- it would just introduce an unused default.

stonesoup/updater/alphabeta.py
jmbarr: Ok. Moved check for measurement model into abstract `Updater` class because I think that's something that all updaters might use.

jmbarr:
Why did tests fail on python-3.7 last time and pass now? Do we have a shonky test?

sdhiscocks:
> Why did tests fail on python-3.7 last time and pass now? Do we have a shonky test?

Looks like a random tolerance failure. I'll take a look.

sdhiscocks (approved):

stonesoup/updater/alphabeta.py
```diff
@@ -0,0 +1,128 @@
...
+    returns :math:`p`. This isn't checked.
+
+    """
+
+    alpha: float = Property(doc="The alpha parameter. Controls the weight given to the "
+                                "measurements over the transition model.")
+    beta: float = Property(doc="The beta parameter. Controls the amount of variation allowed in "
+                               "the velocity component.")
+
+    vmap: np.ndarray = Property(default=None, doc="Binary map of the velocity elements in the"
```
nperree-dstl, line 64:
```suggestion
    vmap: np.ndarray = Property(default=None, doc="Binary map of the velocity elements in the "
```

stonesoup/updater/alphabeta.py
```diff
@@ -0,0 +1,128 @@
...
+                                                  "state vector. If left default, the class will"
```
nperree-dstl, line 65:
```suggestion
                                                  "state vector. If left default, the class will "
```

nperree-dstl (approved):
Only some missing spaces - otherwise looks good!

Issue #327, opened 2020-11-13
Caching Metrics


oharrald-Dstl:
Many metric component calculations are repeated (see SIAPMetric generator).
Caching component values would make metric generation quicker.

mharris-dstl:
Addressed in [505](https://github.com/dstl/Stone-Soup/pull/505).

PR #403, opened 2021-03-09
Add optional caches to SIAP metric generators

oharrald-Dstl:
This PR introduces functools' 'lru cache' decorators to the SIAP metric generators, with the 'simple metric manager' clearing these caches when new data is added or old data overwritten.
Other metric generators have not had their methods decorated (open to discussion).
A new decorator has been created allowing for the user to choose whether the SIAP components are cached or not (either all or none), and provide additional cache arguments ('maxsize', 'typed').
A user may want to cache metrics in an alternate way, however this PR could be considered an intermediate step to creating a more versatile method of allowing user-defined caching methods.

### DIFF
--- a/stonesoup/metricgenerator/_utils.py
+++ b/stonesoup/metricgenerator/_utils.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from functools import lru_cache, wraps
+
+
+def clearable_lru_cache():
+    """Cache decorator that allows keeping track of which methods are decorated.
+    A new cache is created for each instance of :class:`SIAPMetrics`
+    """
+
+    def cache_decorator(func):
+        @wraps(func)
+        def cache_factory(self, *args, **kwargs):
+            if self.cache:
+                instance_cache = lru_cache(*self.cache_args, **self.cache_kwargs)(func)
+                instance_cache = instance_cache.__get__(self, self.__class__)
+                setattr(self, func.__name__, instance_cache)
+
+                self.cached_functions.append(instance_cache)
+
+                return instance_cache(*args, **kwargs)
+            else:
+                return func(self, *args, **kwargs)
+
+        return cache_factory
+
+    return cache_decorator
--- a/stonesoup/metricgenerator/_utils.py
+++ b/stonesoup/metricgenerator/_utils.py
@@ -4,7 +4,8 @@ from functools import lru_cache, wraps
 
 def clearable_lru_cache():
     """Cache decorator that allows keeping track of which methods are decorated.
-    A new cache is created for each instance of :class:`SIAPMetrics`
+    Requires a class utilising this to have a `cache` boolean property, determining whether the
+    decorated method is cached or not.
     """
 
     def cache_decorator(func):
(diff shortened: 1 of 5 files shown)


sdhiscocks:
Would resolve #327

stonesoup/metricgenerator/tracktotruthmetrics.py
```diff
@@ -980,13 +1008,14 @@ def _assoc_distances_sum_t(self, manager, timestamp, mapping, weighting):
             Sum of Euclidean distances (of position or velocity) of each truth to its associated
             tracks in manager at timestamp
         """
-        measure = EuclideanWeighted(mapping=mapping, weighting=weighting)
+        measure = EuclideanWeighted(mapping=list(mapping), weighting=weighting)
```
sdhiscocks, line 1011: This was fixed in #391

stonesoup/metricgenerator/manager.py
```diff
@@ -46,8 +60,15 @@ def add_data(self, groundtruth_paths: Iterable[Union[GroundTruthPath, Platform]]
...
-        self._add(overwrite, groundtruth_paths=groundtruth_paths,
-                  tracks=tracks, detections=detections)
+        # Clear all generator caches
+        for generator in self.generators:
+            try:
+                generator.clear_caches()
+            except AttributeError:
+                pass
+
+        self._add(overwrite, _groundtruth_paths=groundtruth_paths,
```
sdhiscocks, line 70: When `overwrite` is `True`, the `_add` method should probably copy the set, otherwise you are storing mutable version which may be modified elsewhere.

stonesoup/metricgenerator/_utils.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from functools import lru_cache, wraps
+
+
+def clearable_lru_cache():
+    """Cache decorator that allows keeping track of which methods are decorated.
+    A new cache is created for each instance of :class:`SIAPMetrics`
```
sdhiscocks, line 7: Probably don't need to reference SIAP here, as could be used for other metrics too.

sdhiscocks (changes requested):

stonesoup/metricgenerator/_utils.py
oharrald-Dstl: Oops, must be left over from when Id had the cache as a decorator in the SIAP file

stonesoup/metricgenerator/manager.py
oharrald-Dstl: Does this need to be a deep copy? Individual ground truth paths / tracks / detections could be modified?

sdhiscocks: I think a shallow copy would be okay.

stonesoup/metricgenerator/manager.py
```diff
@@ -46,11 +61,23 @@ def add_data(self, groundtruth_paths: Iterable[Union[GroundTruthPath, Platform]]
...
+        if overwrite:
+            self._add(overwrite, _groundtruth_paths=copy.copy(groundtruth_paths),
+                      _tracks=copy.copy(tracks), _detections=copy.copy(detections))
+        else:
+            self._add(overwrite, _groundtruth_paths=groundtruth_paths,
+                      _tracks=tracks, _detections=detections)
 
     def _add(self, overwrite, **kwargs):
         for key, value in kwargs.items():
+            print(key)
```
sdhiscocks, line 80: `print` left in here.

PR #415, opened 2021-03-18
Trackers are Iterables instead of Generators 

gawebb-dstl:
The trackers have been turned into iterables instead of generators. This allows trackers to be copied/ manipulated while producing tracks if needed.

Instead of `for time, tracks in tracker.tracks_gen():`
you now write
`for time, tracks in tracker:`

There _should_ be no other changes to how the trackers function

### DIFF
--- a/stonesoup/tracker/base.py
+++ b/stonesoup/tracker/base.py
@@ -2,23 +2,24 @@
 from abc import abstractmethod
 
 from ..base import Base
-from ..buffered_generator import BufferedGenerator
 
 
-class Tracker(Base, BufferedGenerator):
+class Tracker(Base):
     """Tracker base class"""
 
     @property
+    @abstractmethod
     def tracks(self):
-        return self.current[1]
+        raise NotImplementedError
+
+    def __iter__(self):
+        return self
 
     @abstractmethod
-    @BufferedGenerator.generator_method
-    def tracks_gen(self):
-        """Returns a generator of tracks for each time step.
-
-        Yields
-        ------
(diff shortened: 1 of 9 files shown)


gawebb-dstl:
@sdhiscocks I've added a generator method (with a deprecation warning) into the base tracker so that the tracker will function as normal. This is so people's code doesn't (hopefully) break straight away

sdhiscocks:
If we are going to make a change, we should probably extend this to all other classes that use the `BufferedGenerator`: `Reader`, `Feeder`, `Detector`, `Simulator` and `Writer`.

I think migrating to making these classes *iterators* (and thinking *iterator* rather than *iterable* to be clear on distinction) is good idea. This is more standard approach than using `BufferedGenerator`, which was kinda working around the fact the class instances were not iterable in the first place.

On key element of iterators, is subsequent calls to `__iter__` shouldn't reset or modify the calls to `__next__`. I don't think this should be an issue, but may need to see when code changes. Note that one of readers that I made a change to in this PR breaks that, but should be easy to resolve.

stonesoup/tracker/base.py
```diff
@@ -9,19 +10,29 @@ class Tracker(Base, BufferedGenerator):
     """Tracker base class"""
```
sdhiscocks, line 10: This probably shouldn't inherit form `BufferedGenerator` anymore.

oharrald-Dstl (approved):

jswright-dstl (approved):

PR #416, opened 2021-03-19
Add mapping to CovarianceBasedDeleter

oharrald-Dstl:
Sometimes it is useful to consider deleting a track only when particular components of its state vector become very uncertain. This change introduces a mapping to the `CovarianceBasedDeleter`, which allows the user to define which components of the track state should be taken in to consideration when evaluating a covariance trace.
For example, with a track state vector (x, vx, y, vy), the user might only want to delete a track if its positional uncertainties grow too large, in which case a mapping of [0, 2] could be passed in, in which case the deleter will sum the variance in x and y only.

### DIFF
--- a/stonesoup/deleter/error.py
+++ b/stonesoup/deleter/error.py
@@ -17,9 +17,10 @@ class CovarianceBasedDeleter(Deleter):
 
     covar_trace_thresh: float = Property(doc="Covariance matrix trace threshold")
     mapping: Sequence[int] = Property(default=None,
-                                 doc="Track state vector indices whose corresponding covariances' "
-                                     "sum is to be considered. Defaults to None, whereby the "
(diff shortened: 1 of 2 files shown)


stonesoup/deleter/error.py
```diff
@@ -14,6 +16,10 @@ class CovarianceBasedDeleter(Deleter):
     """
 
     covar_trace_thresh: float = Property(doc="Covariance matrix trace threshold")
+    mapping: Sequence = Property(default=None,
```
sdhiscocks, line 19:
```suggestion
    mapping: Sequence[int] = Property(default=None,
```

sdhiscocks (approved):
:+1:

jswright-dstl (approved):

PR #427, opened 2021-03-26
Refactor platforms and sensors to remove weakrefs

etfrogers-dstl:
This reafactor splits `Platform` functionality into moving (`Movable`) and and sensor-holding (`Platform`) functionality. Removes the need for sensors to hold weakrefs back to their platform.

`Platform`s now have a `movement_controller: Movable` Property to control movement, and attribute access on the Platform is proxied to its movement_controller, if not directly provided by the Platform. Also, the Platform constructor will try to construct a `movement_controller`, if one is not provided, using its input arguments. When a `Sensor` is added to a `Platform`, the `Sensor`'s `movement_controller` is set equal to the `Platform`'s `movement_controller`.

The external interface to these classes is very similar to before: The major change being that sensors now hold their own mounting/rotation offsets, rather than the platform holding them. This massively reduces the amount of book-keeping needed.

Also, sensors no long have a `platform`  Property from which to get their position: they now have a `movement_controller` instead.


etfrogers-dstl:
This change means the we have no weakrefs in the codebase, and so the `__deepcopy__` implmentation that I build in `base.py` is no longer tested (or, in fact, needed).

Should it be deleted? I think it probably should, as I would argue we shouldn't be re-introducing weakrefs later.

stonesoup/platform/base.py
```diff
@@ -1,122 +1,89 @@
...
-        doc="Mapping between platform velocity and state dims. If not "
-            "set, it will default to ``[m+1 for m in position_mapping]``")
 
-    # TODO: Determine where a platform coordinate frame should be maintained
+    _movable_mapping = {'FixedPlatform': FixedMovable,
+                        'MovingPlatform': MovingMovable,
+                        'MultiTransitionMovingPlatform': MultiTransitionMovable,
+                        'FixedRadarTarget': FixedMovable,
+                        'MovingRadarTarget': MovingMovable,
+                        'ManoeuvringRadarTarget': MultiTransitionMovable}
```
sdhiscocks, line 46: This would be better as a class attribute in the sub class, than having a mapping here.

etfrogers-dstl: Much better! Have fixed.

stonesoup/platform/base.py
```diff
@@ -229,428 +108,58 @@ def remove_sensor(self, sensor: "BaseSensor") -> None:
...
-                        noise=True,
-                        time_interval=self.current_interval,
-                        **kwargs),
-                    timestamp=timestamp
-                )
-                time_interval -= self.current_interval
-                self.transition_index = (self.transition_index + 1) % len(self.transition_models)
-                self.current_interval = self.transition_times[self.transition_index]
+    def __iter__(self):
+        return self.movement_controller.__iter__()
```
sdhiscocks, line 150: How about changing this for `__getitem__`? This would enable access to states, and make the class iterable still.

sdhiscocks (commented):
Looks good. :+1:

Was wondering if `moveable` would be better under `platform` i.e. `stonesoup.platform.moveable`?

etfrogers-dstl:
> Looks good. 👍
> 
> Was wondering if `moveable` would be better under `platform` i.e. `stonesoup.platform.moveable`?

It was there originally. I moved it to its own folder because of the import of `Platform` in `stonesoup/platform/__init__.py`. That import leads to a circular import if `Moveable` is within `stonesoup/platform`. I will have another think if there is a better way, though.

sdhiscocks:
> > Looks good. 👍
> > Was wondering if `moveable` would be better under `platform` i.e. `stonesoup.platform.moveable`?
> 
> It was there originally. I moved it to its own folder because of the import of `Platform` in `stonesoup/platform/__init__.py`. That import leads to a circular import if `Moveable` is within `stonesoup/platform`. I will have another think if there is a better way, though.

I hadn't realised that would cause an issue. Fine to leave it where it is then.

stonesoup/platform/base.py
etfrogers-dstl: Agreed. Done

jjosborne-dstl (approved):

stonesoup/moveable/movable.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """
+        Ensure that the platform location and the sensor locations are
+        consistent at initialisation.
+        """
+        super().__init__(*args, **kwargs)
+        # Set values to defaults if not provided
+
+        if self.velocity_mapping is None:
```
jjosborne-dstl, line 45: Should moveable insist on having at least one state in states when initiated? Does FixedMoveable need to have a velocity_mapping?

etfrogers-dstl: `states` is already a required input. So this would be a check that `states` is not empty. Not sure whether this is a good idea or not. It might be.

A `FixedMovable` doesn't need a `velocity_mapping` (and in fact it will be ignored)

jjosborne-dstl: Seems like it inherits states from StateMutableSequence even though it shouldn't. When it gets to StateMutableSequence.__init__ states = None and then it is set to an empty list. This could just be a problem on my side though.

sdhiscocks (approved):

stonesoup/moveable/movable.py
```diff
@@ -0,0 +1,532 @@
...
+
+    @property
+    def is_moving(self) -> bool:
+        return False
+
+    def move(self, timestamp: datetime.datetime, **kwargs) -> None:
+        """For a fixed platform this method has no effect other than to update the timestamp."""
+        # TODO Is this a sensible implementation?
+        # Return without moving static platforms
+        self.state.timestamp = timestamp
```
oharrald-Dstl, line 219: While the state vector remains the same, as the timestamp changes, should this 'new' state be appended to :attr:`states`?

stonesoup/moveable/movable.py
```diff
@@ -0,0 +1,532 @@
...
+        roll always set to zero (as this is the angle the platform is rotated about the velocity
+        axis, which is not defined in this approximation).
+
+        Notes
+        -----
+        A non-moving platform (``self.is_moving == False``) does not have a defined orientation in
+        this approximations and so raises an :class:`AttributeError`
+        """
+        if not self.is_moving:
+            raise AttributeError('Orientation of a zero-velocity moving platform is not defined')
```
oharrald-Dstl, line 265: A moving platform that has slowed to 0 speed in every direction could still have an orientation equal to its orientation just before coming to a complete stop?

stonesoup/moveable/movable.py
```diff
@@ -0,0 +1,532 @@
...
+            (the default is ``None``)
+
+        Notes
+        -----
+        This methods updates the value of :attr:`position`.
+
+        Any provided ``kwargs`` are forwarded to the :attr:`transition_model`.
+
+        If :attr:`transition_model` or ``timestamp`` is ``None``, the method has
+        no effect, but will return successfully.
```
oharrald-Dstl, line 316: This is false in the instance there is no transition model. An `AttributeError` is raised.

stonesoup/moveable/movable.py
```diff
@@ -0,0 +1,532 @@
...
+
+        if self.state.timestamp is None:
+            self.state.timestamp = timestamp
+            return
+
+        # Compute time_interval
+        try:
+            time_interval = timestamp - self.state.timestamp
+        except TypeError:
+            # TypeError: (timestamp or prior.timestamp) is None
```
oharrald-Dstl, line 328: Could this comment be removed?

stonesoup/moveable/movable.py
```diff
@@ -0,0 +1,532 @@
...
+
+def _get_rotation_matrix(vel: StateVector) -> np.ndarray:
+    """ Generates a rotation matrix which can be used to determine the
+    corrected sensor offsets.
+
+    In the 2d case this returns the following rotation matrix
+    [cos[theta] -sin[theta]]
+    [cos[theta]  sin[theta]]
+
+    In the 2d case this will be a 3x3 matrix which rotates around the Z axis
```
oharrald-Dstl, line 439: In the 3D case?

stonesoup/platform/base.py
```diff
@@ -1,122 +1,83 @@
...
-        doc="Mapping between platform velocity and state dims. If not "
-            "set, it will default to ``[m+1 for m in position_mapping]``")
 
-    # TODO: Determine where a platform coordinate frame should be maintained
+    _default_movable_class = None  # Will be overridden by subclasses
+
+    def __getattribute__(self, name):
+        # This method is called if we try to access an attribute of self. First we try to get the
+        # attribute directly, but if that fails, we want to try getting the same attribute from
+        # self.state instead. If that, in turn,  fails we want to return the error message that
```
oharrald-Dstl, line 46: Attempt to get attribute from the `controller` attribute, not the state

stonesoup/platform/base.py
```diff
@@ -1,122 +1,83 @@
...
-        deformation (e.g. flex).
+    A simple platform is considered to always be aligned with its principle
+    velocity. It does not take into account issues such as bank angle or body
+    deformation (e.g. flex).
 
+    Movement is controlled by the Platform's :attr:`Platform.movement_controller`, and access to
+    attributes of the Platform is proxied to the movement controller, to allow the Platform to
+    report it's own position, orientation etc.
 
-        .. note:: This class abstract and not intended to be instantiated. To get the behaviour of
```
oharrald-Dstl, line 34: Probably want to still mention that this is not meant to be instantiated (due to default movement controller being None)

stonesoup/sensor/base.py
```diff
@@ -1,101 +1,103 @@
...
+
+        if self.rotation_offset is None:
+            self.rotation_offset = StateVector([0] * 3)
+
+    @movement_controller.setter
+    def movement_controller(self, value):
+        if self._has_internal_controller:
+            raise AttributeError('The movement controller cannot be set on a Sensor with a private'
+                                 'internal controller')
+        self._property_movement_controller = value
```
oharrald-Dstl, line 75: Should this not be the attribute that is checked against later in the code, rather than `_internal_movement_controller`?
I'm confused over the use of having both.

oharrald-Dstl (commented):
A few comments. Many places where I might just be confused.

stonesoup/moveable/movable.py
etfrogers-dstl: I have made two changes in a new commit (to be pushed later after the other comments are addressed):
1) `states` is now checked for emptyness.
2) `FixedMovable` checks that velocity mapping is not set.

stonesoup/moveable/movable.py
etfrogers-dstl: Well spotted. Fixed.

stonesoup/moveable/movable.py
etfrogers-dstl: Documentation is correct to match the behaviour of the code.

stonesoup/moveable/movable.py
etfrogers-dstl: Documentation corrected

stonesoup/platform/base.py
etfrogers-dstl: Documentation corrected

stonesoup/platform/base.py
etfrogers-dstl: This is still in the documentation (current line 26). It is marked as deleted, but then also marked as re-added later (as the text changed, along with the code around it).

stonesoup/platform/base.py
```diff
@@ -229,428 +108,58 @@ def remove_sensor(self, sensor: "BaseSensor") -> None:
...
-    """
-    transition_model: TransitionModel = Property(doc="Transition model")
-
+    # The position, orientation and velocity getters are not required, as __getattribute__ will do
+    # the job, but the setters are required, and this seems the cleanest way to implement them
     @property
-    def velocity(self) -> StateVector:
-        """Return the velocity of the platform.
+    def position(self):
+        return self.movement_controller.position
```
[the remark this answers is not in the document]
etfrogers-dstl: See comment on (current) line 124

````
    # The position, orientation and velocity getters are not required, as __getattribute__ will do
    # the job, but the setters are required, and this seems the cleanest way to implement them
    @property
    def position(self):
        return self.movement_controller.position
````

File: stonesoup/platform/base.py [hunk 13]
```diff
@@ -229,428 +108,58 @@ def remove_sensor(self, sensor: "BaseSensor") -> None:
...
-    Returns
-    -------
-    np.ndarray
-        3x3 rotation matrix
-    """
-    return _rot3d_tuple(tuple(vec.flat))
 
+class FixedPlatform(Platform):
+    _default_movable_class = FixedMovable
+    pass
```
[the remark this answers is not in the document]
etfrogers-dstl: No, not needed. Removed.

stonesoup/moveable/movable.py
```diff
@@ -0,0 +1,532 @@
...
+        if not self.is_moving:
+            raise AttributeError('Orientation of a zero-velocity moving platform is not defined')
+        velocity = self.velocity
+
+        if self.ndim == 3:
+            _, bearing, elevation = cart2sphere(*velocity.flat)
+            return StateVector([0, elevation, bearing])
+        elif self.ndim == 2:
+            _, bearing = cart2pol(*velocity.flat)
+            return StateVector([0, 0, bearing])
```
jjosborne-dstl, line 273: For the 2d case should the orientation be just the bearing and the rotation matrix is 2d?

stonesoup/sensor/base.py
```diff
@@ -1,101 +1,103 @@
...
+from ..base import Base, Property
 from ..types.array import StateVector
-from ..types.detection import TrueDetection
-from ..types.groundtruth import GroundTruthState
 
 
-class BaseSensor(Base, ABC):
-    """Sensor base class
+class PlatformMountable(Base, ABC):
+    """Base class for any objects that is part of a platform.
```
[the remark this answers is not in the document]
etfrogers-dstl: Update documentation

stonesoup/sensor/base.py
```diff
@@ -1,101 +1,103 @@
...
+    position and orientation.
 
     """
 
+    rotation_offset: StateVector = Property(
+        default=None,
+        doc="A StateVector containing the sensor rotation "
+            "offsets from the platform's primary axis (defined as the "
+            "direction of motion). Defaults to a zero vector with the "
+            "same length as the Platform's :attr:`position_mapping`")
```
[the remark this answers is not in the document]
etfrogers-dstl: Updated both documentation and code to this approach

stonesoup/sensor/base.py
etfrogers-dstl: You did notice an oddity here. I typed up the explanation below, before looking at the code and realizing it didn't actually do that! I have reworked the code and I'll try and explain the new logic here:

If no moveable is supplied to a sensor at construction, but it is given a position and/or orientation, it constructs an _internal_ `FixedMovable` to handle it's position. This is stored in the `movement_controller` Property and cannot be changed during the life of the object. It allows the sensor position and orientation to be set.

The `_has_internal_controller` is a (readonly) internal boolean flag that is set to indicate this state. It cannot be stored in the `movement_controller` Property as this used for either a "normal" controller or an internal one.

oharrald-Dstl (approved):
Looks good. There are a few lines of code that could be tested quite easily to get coverage up to ~100%. These are mainly instances where errors get raised.

stonesoup/moveable/movable.py
etfrogers-dstl: The convention throughout is that rotation is handled in 3D. I am unkeen on changing this here, as it's a bit of a big job, and not trivial to get right. I think this is probably out of scope of this PR.

etfrogers-dstl:
> Looks good. There are a few lines of code that could be tested quite easily to get coverage up to ~100%. These are mainly instances where errors get raised.

I have added the tests that were obvious to me... I think this gets coverage of files I touched up to 100%

stonesoup/moveable/movable.py
etfrogers-dstl: This is correct, but without lots of bookkeeping, I'm not sure how you find that out. You could check the last state in the history with a velocity, but in the case of a platform that is moved in large steps and/or is turning rapidly, the last recorded state could be very misleading.

If you can think of a way of finding out "its orientation just before coming to a complete stop", then I'm happy to implement, but right now I can't see how to do it.

oharrald-Dstl: Good point. I guess I was hoping there would be a magical solution to this issue, but what you've got here makes sense and is what we already use so no problem to me.

PR #431, opened 2021-04-08
Speed up track to truth associator by avoiding searching of all times

sdhiscocks:
This speeds up track to truth association by:
 - avoiding looking at truths which start/end are after/before the
   current timestep
 - and maintaining an iterator for each truth to pick up from the
   last timestep and search forward for a matching timestamp.


sdhiscocks:
I just realised this doesn't work correctly when there are multiple states at the and timestamp, as it should return the last one at that time, not the first.

idorrington-dstl (approved):

nperree-dstl (approved):

Issue #396, opened 2021-03-01
Random number generation managment


sdhiscocks:
Currently, the only way to influence random number generation in Stone Soup is via global random seed. This means, for example, that changing a simulation to having 2 targets, to having 3 targets, means that the behaviour of the first 2 targets will also change. This would also effect things like simulated measurements. This behaviour may be undesired.

This could be solved by a combination of:

- Setting a seed value per model.
- Allowing seed to be passed to RVS method of the models.
- Setting a seed to simulators or other code where random numbers are called.

apiszcz:
Desired feature, optional, allow per model. Third option might work similar to other initializers.

sdhiscocks:
Suggested modification that I tested for `GaussianModels`.

```diff
diff --git a/stonesoup/models/base.py b/stonesoup/models/base.py
index 3c2b394b..ac406f8e 100644
--- a/stonesoup/models/base.py
+++ b/stonesoup/models/base.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 from abc import abstractmethod
+from typing import Optional
 
 import numpy as np
 from scipy.stats import multivariate_normal
 
-from ..base import Base
+from ..base import Base, Property
 from ..functions import jacobian as compute_jac
 from ..types.array import StateVector, StateVectors
 from ..types.numeric import Probability
@@ -163,7 +164,13 @@ class GaussianModel(Model):
 
     Base/Abstract class for all Gaussian models"""
 
-    def rvs(self, num_samples=1, **kwargs):
+    seed: Optional[int] = Property(default=None)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.random_state = np.random.RandomState(self.seed) if self.seed is not None else None
+
+    def rvs(self, num_samples=1, random_state=None, **kwargs):
         r"""Model noise/sample generation function
 
         Generates noise samples from the model.
@@ -194,8 +201,9 @@ class GaussianModel(Model):
         if covar is None or None in covar:
             raise ValueError("Cannot generate rvs from None-type covariance")
 
+        random_state = random_state if random_state is not None else self.random_state
         noise = multivariate_normal.rvs(
-            np.zeros(self.ndim), covar, num_samples)
+            np.zeros(self.ndim), covar, num_samples, random_state=random_state)
 
         noise = np.atleast_2d(noise)
 
```

PR #444, opened 2021-05-17
User random seeds

idorrington-dstl:
Addresses the issue raised here
https://github.com/dstl/Stone-Soup/issues/396

Now all models that use random number generation have an optional seed that allows the user to, for example, always produce the same randomly generated truth data but allow other components to change between runs. 

Also needed to change the groundtruth paths to be an ordered set rather than a set in the groundtruth paths generator, otherwise the order that items were iterated through in the set would cause different results between runs. 

I added the seeds to the metrics example as well, as this was my test case and it makes sense that it should produce consistent results.

It would be useful if reviewers could make sure I haven't missed any models. I'd also appreciate a second opinion on the tests I wrote to check they don't miss anything important.

### DIFF
--- a/stonesoup/simulator/simple.py
+++ b/stonesoup/simulator/simple.py
@@ -62,15 +62,18 @@ class SwitchOneTargetGroundTruthSimulator(SingleTargetGroundTruthSimulator):
     The element in the ith row and the jth column is the probability of\
      switching from the ith transition model in :attr:`transition_models`\
      to the jth")
+    seed: Optional[int] = Property(default=None, doc="Seed for random number generation."
+                                                     " Default None")
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.index = 0
+        self.random_state = np.random.RandomState(self.seed)
 
     @property
     def transition_model(self):
-        self.index = np.random.choice(range(0, len(self.transition_models)),
-                                      p=self.model_probs[self.index])
+        self.index = self.random_state.choice(range(0, len(self.transition_models)),
+                                              p=self.model_probs[self.index])
         return self.transition_models[self.index]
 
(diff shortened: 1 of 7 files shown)


stonesoup/simulator/simple.py
```diff
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
+from typing import Optional
 import datetime
 from typing import Sequence
```
sdhiscocks, line 4:
```suggestion
import datetime
from typing import Sequence, Optional
```

stonesoup/simulator/simple.py
```diff
@@ -60,15 +62,19 @@ class SwitchOneTargetGroundTruthSimulator(SingleTargetGroundTruthSimulator):
...
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.index = 0
+        self.random_state = np.random.RandomState(self.seed)
 
     @property
-    def transition_model(self):
-        self.index = np.random.choice(range(0, len(self.transition_models)),
-                                      p=self.model_probs[self.index])
+    def transition_model(self, random_state=None):
```
sdhiscocks, line 74: Can't have argument on _property_ like this, as no way to pass arguments. May be you could change this to `_transition_model` and add `transition_model = property(_transition_model)`, and then call it with `self._transition_model(random_state)` inside the simulation.

Maybe just remove `transition_model` altogether 🤷

idorrington-dstl: Now changed it to just use self.random_state. I think that should fix it.

sdhiscocks (approved):

jswright-dstl (approved):

jswright-dstl (approved):

PR #458, opened 2021-07-02
Add 2d density plot to Plotter

gawebb-dstl:
…f StateMutableSequence

### DIFF
--- a/docs/examples/density_plot_example.py
+++ b/docs/examples/density_plot_example.py
@@ -1,0 +1,53 @@
+from datetime import datetime, timedelta
+from matplotlib import pyplot as plt
+
+
+from stonesoup.types.groundtruth import GroundTruthPath, GroundTruthState
+from stonesoup.models.transition.linear import CombinedLinearGaussianTransitionModel, \
+                                               ConstantVelocity
+from stonesoup.plotter import Plotter
+
+start_time = datetime.now()
+
+
+def generate_ground_truth_path(initial_state, num_steps=20, q=0.01):
+
+    transition_model = CombinedLinearGaussianTransitionModel([ConstantVelocity(q), ConstantVelocity(q)])
+
+    ground_truth = GroundTruthPath([GroundTruthState(initial_state, timestamp=start_time)])
+
+    for k in range(0, num_steps):
+        ground_truth.append(GroundTruthState(
+            transition_model.function(ground_truth[k], noise=True, time_interval=timedelta(seconds=1)),
+            timestamp=start_time+timedelta(seconds=k+1)))
+
+    return ground_truth
+
+
+n_time_steps = 20
+truth = generate_ground_truth_path(initial_state=[0, 0, 0, 1], num_steps=n_time_steps)
+
+Plotter().plot_ground_truths(truth, [0, 2])
+plt.show()
+
+truths = [generate_ground_truth_path(initial_state=[0, 0, 0, 1], num_steps=n_time_steps, q=0.1) for _ in range(100)]
+
+# This is quite messy
+Plotter().plot_ground_truths(set(truths), [0, 2])
+plt.show()
+
+# This is clearer
+Plotter().plot_density(truths, index=None)
+plt.show()
+
+# Only show the last state
+Plotter().plot_density(truths, index=-1)
+plt.show()
+
+plotter = Plotter()
+for i in range(1, n_time_steps):
+    plotter.plot_density(truths, index=i)
+    plt.show(block=False)
+    plt.pause(0.1)
+
+plt.show()
--- a/docs/examples/density_plot_example.py
+++ b/docs/examples/density_plot_example.py
@@ -1,3 +1,17 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+Density Plot Example
+===============================================
+This example looks at how to plot 2d density plots. The `Plotter().plot_density` function can be
+used to plot any number of :class:`~.StateMutableSequence` objects. StateMutableSequences are just a
+container for a number of states, examples include tracks and ground truth paths. In the examples
+below show how to plot ground truth paths (as they're easy to generate). The function can be used to
+analyse large data sets
+"""
+# %%
+# Set the imports and set the start time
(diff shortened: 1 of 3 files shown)


stonesoup/plotter.py
```diff
@@ -271,6 +275,32 @@ def plot_tracks(self, tracks, mapping, uncertainty=False, particle=False, track_
         else:
             self.ax.legend(handles=self.handles_list, labels=self.labels_list)
 
+    def plot_density(self, state_sequences: Sequence[StateMutableSequence], index=-1, mapping=(0, 2), n_bins=300,
+                     **kwargs):
+
+        if index is None:
```
oharrald-Dstl, line 281: Don't need this check with default value of -1. Unless you are expecting users to manually input 'index=-1'.

oharrald-Dstl (commented):
The sphinx-gallery example will not generate due to there being no docstring at the start of the file (see Tutorials/other examples).
Calls of 'fig.show()' might need to be changed to just 'fig' in order for output to be shown correctly, along with separators '# %%' between each section of code (perhaps a new section for each plot).
The plots look good!

docs/examples/density_plot_example.py
```diff
@@ -0,0 +1,53 @@
...
+from stonesoup.types.groundtruth import GroundTruthPath, GroundTruthState
+from stonesoup.models.transition.linear import CombinedLinearGaussianTransitionModel, \
+                                               ConstantVelocity
+from stonesoup.plotter import Plotter
+
+start_time = datetime.now()
+
+
+def generate_ground_truth_path(initial_state, num_steps=20, q=0.01):
+
```
oharrald-Dstl, line 14: Could you add a docstring here?

stonesoup/plotter.py
```diff
@@ -271,6 +275,32 @@ def plot_tracks(self, tracks, mapping, uncertainty=False, particle=False, track_
         else:
             self.ax.legend(handles=self.handles_list, labels=self.labels_list)
 
+    def plot_density(self, state_sequences: Sequence[StateMutableSequence], index=-1, mapping=(0, 2), n_bins=300,
```
oharrald-Dstl, line 278: Could a docstring be added here?

stonesoup/plotter.py
gawebb-dstl: This should have been documented more clearly. If you manually set `index=None` the code will plot all the states in the sequences

stonesoup/plotter.py
gawebb-dstl: Will do.
My IDE had set the limit to 120 for some reason. Does the 100 character limit apply to the example as well?

oharrald-Dstl (approved):

stonesoup/plotter.py
sdhiscocks: > Does the 100 character limit apply to the example as well?

It does, but we're a bit more relaxed about  examples, as long as they look okay when rendered.

docs/examples/density_plot_example.py
```diff
@@ -0,0 +1,95 @@
...
+# Plot each state
+# -------------------------------------------------
+# Plot the density at each time-step and see how the density plot evolves
+plotter = Plotter()
+for i in range(1, n_time_steps):
+    plotter.plot_density(truths, index=i)
+    plt.show(block=False)
+    plt.pause(0.1)
+
+plt.show()
```
sdhiscocks, line 95: I think this would be better, as then it'll animate in the gallery.

```suggestion
plotter = Plotter()


def update(i):
    plotter.ax.clear()
    plotter.plot_density(truths, index=i)
    return plotter.ax,


animation.FuncAnimation(plotter.fig, update, frames=range(1, n_time_steps))
```

docs/examples/density_plot_example.py
```diff
@@ -0,0 +1,95 @@
...
+
+# %%
+# Plot of the Last State of the Ground Truths
+# -------------------------------------------------
+# The function allows you to pick an index of the state sequence (ground truth in this example) to
+# plot. In this example we're only interested in the final state of the sequences. An index of '-1'
+# is the last state in the sequence.
+# The resultant plot in much more spread out
+Plotter().plot_density(truths, index=-1)
+plt.show()
```
sdhiscocks, line 83: These should be removed.

stonesoup/plotter.py
```diff
@@ -271,6 +274,51 @@ def plot_tracks(self, tracks, mapping, uncertainty=False, particle=False, track_
...
+                          for a_state_sequence in state_sequences
+                          for a_state in a_state_sequence])
+        else:  # Only plot one state out of the sequences
+            x = np.array([a_state_sequence.states[index].state_vector[mapping[0]]
+                          for a_state_sequence in state_sequences])
+            y = np.array([a_state_sequence.states[index].state_vector[mapping[1]]
+                          for a_state_sequence in state_sequences])
+
+        # Evaluate a gaussian kde on a regular grid of n_bins x n_bins over data extents
+        k = kde.gaussian_kde([x, y])
```
sdhiscocks, line 313: I think this raises an error when all values of x and y are the same. May want to catch that, as well as skip plotting if `state_sequences` is empty.

stonesoup/plotter.py
```diff
@@ -271,6 +274,51 @@ def plot_tracks(self, tracks, mapping, uncertainty=False, particle=False, track_
         else:
             self.ax.legend(handles=self.handles_list, labels=self.labels_list)
 
+    def plot_density(self, state_sequences: Iterable[StateMutableSequence],
```
sdhiscocks, line 277: It'd be cool to extend this (maybe raise as a separate issue) to support Gaussian States and Gaussian Mixture States

sdhiscocks (changes requested):

jswright-dstl (approved):
I have tidied up the example to include additional comments.

stonesoup/plotter.py
jswright-dstl: Added raise errors for empty `state_sequences`  and when all x, y values are the same.

PR #461, opened 2021-07-08
Fix deterministic use of random seed

joldf:
Deterministic use of random seeds was originally implemented in #444 in response to #396. However, when a model was not explicitly passed a random seed it would use a random value, which meant the run was not deterministic overall. This fixes that, using the change originally suggested in #396.

CC: @idorrington-dstl @sdhiscocks

### DIFF
--- a/stonesoup/models/base.py
+++ b/stonesoup/models/base.py
@@ -196,7 +196,7 @@ class GaussianModel(Model):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.random_state = np.random.RandomState(self.seed)
+        self.random_state = np.random.RandomState(self.seed) if self.seed is not None else None
 
     def rvs(self, num_samples: int = 1, random_state=None, **kwargs) ->\
             Union[StateVector, StateVectors]:


idorrington-dstl:
Is it not typical behaviour that when a model is not explicitly passed a random seed then it is not deterministic? 

If we do want deterministic behaviour by default, then I think a safer way to do it is to use a default seed value rather than set random state to None. Otherwise we will have problems if people try to use methods of the random state (which is not a random state anymore) and we'll have to put "if random_state is not None" everywhere we use random states. To be honest, I'm surprised this passed the tests...

sdhiscocks (approved):

joldf:
@idorrington-dstl 

I don't have enough experience with Stone Soup to properly comment your question, but I can tell you how it affected me: I just ran the tutorials a few times and found they weren't detministic. They currently set the seed by just calling `np.random.seed(1991)` at the start. What would your suggestion be instead? Would they have to pass the seed to every model they instantiate? Aside from being more typing, that would very error prone: it would be easy for the developer to miss one place they need to pass in a seed, and that would ruin the determinism. The problem would pass silently, so someone could waste a lot of time generating data that turns out not to be useful for their needs.

On the other hand, I haven't thought about this from the perspective of someone that *doens't* want determinism.

joldf:
@idorrington-dstl Sorry, I think I had misread your comment a bit. It sounds like your suggestion would not look any different from the user perspective. But it does sound like changing `np.seed(...)` wouldn't then affect the models because they have some other default seed embedded? Or do you just mean a default parameter which is an instance of the `Seed` class that uses the `np.seed` global state? (If so, to make sure it's instantiated at the time of the call, rather than during function parsing, you'd need to use a constant token instead and instantiate inside the function body... if you use `None` as the token then you arrive back at the currently suggested code.)

sdhiscocks:
@idorrington-dstl This issue with the current implementation, is that when a seed isn't provided, rather than using the global random pool (behaviour prior to #444, and I'd say expected behaviour), it creates a RandomState that reads data from system pool (or clock).

From [`RandomState` documentation](https://numpy.org/doc/stable/reference/random/legacy.html?highlight=randomstate#numpy.random.RandomState):
"If seed is None, then the MT19937 BitGenerator is initialized by reading data from /dev/urandom (or the Windows analogue) if available or seed from the clock otherwise."

stonesoup/models/base.py
```diff
@@ -196,7 +196,7 @@ class GaussianModel(Model):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.random_state = np.random.RandomState(self.seed)
+        self.random_state = np.random.RandomState(self.seed) if self.seed is not None else None
```
idorrington-dstl, line 199:
```suggestion
        self.random_state = np.random.RandomState(self.seed) if self.seed else\
            np.random.RandomState(np.random.get_state()[1])
```
Ah, ok, I understand the problem now.  I didn't realise the default random state used a new seed. 

The problem we have is that we should really be doing this stuff with default_rng, as numpy random seed is now legacy code.
https://albertcthomas.github.io/good-practices-random-number-generators/

Unfortunately, this is not in the numpy version we have at dstl (maybe this will change soon?) but for now we need to use seeds/random states. 

Numpy has the get_state method for returning the global state. For now we can pass this as the seed and I think this fixes the reproducibility problem and keeps self.random_state as a random_state. 

The documentation for get_state explicitly says this should not be necessary, but it seems that best practice is to pass the RNG/random state around if you want reproducible results, so any method to use a default global state seems to be a bit of a hack (see [here](https://github.com/numpy/numpy/issues/15322)). I agree this is kind of a pain, but I do get the reasons behind it.

Anyhow, I think this is the least bad way to do things for now, until we get the latest numpy as dstl. At that point we should probably completely move away from random seeds, use default_rng, and get into the habit of passing it around. 

That or I've misunderstood something and really overcomplicated this (which is always a possibility).

idorrington-dstl (changes requested):

stonesoup/models/base.py
idorrington-dstl: I've open a new issue #462 to implement a more permanent solution. We should be getting a newer version of Numpy at Dstl soon which will allow us to use the current best practice. In the mean time, either solution works for the time being, and as it is only a temporary fix I'm not that fussed. 

Thank you @joldf for bringing this to our attention. I didn't realise my new way of producing reproducible random states had meant global random states stopped working.

idorrington-dstl (approved):

joldf:
@idorrington-dstl Thanks 👍

PR #468, opened 2021-07-27
Added equal operator in Base

gawebb-dstl:
If a Stone-Soup object is only defined by it's properties then this should be a way of comparing two objects? 

I haven't tested this yet, but wanted some thoughts before I continued


sdhiscocks:
The only way to be able to compare items as equal, is also if their hashes are also equal, which is more difficult to implement, as then every property you use must then be hashable. Probably not practical to implement this.

svidal-dstl (approved):

sdhiscocks:
Closing this, as incomplete, and I'm not sure it's practical.

PR #475, opened 2021-07-28
set dwell_center.timestamp equal to incoming timestamp if dwell_cente…

alexiad1:
…r.timestamp is None

### DIFF
--- a/stonesoup/sensor/radar/radar.py
+++ b/stonesoup/sensor/radar/radar.py
@@ -163,6 +163,9 @@ class RadarRotatingBearingRange(RadarBearingRange):
             A timestamp signifying when the rotation completes
         """
 
+        if self.dwell_center.timestamp == None:
+            self.dwell_center.timestamp = timestamp
+            
         # Compute duration since last rotation
         duration = timestamp - self.dwell_center.timestamp
 
--- a/stonesoup/sensor/radar/radar.py
+++ b/stonesoup/sensor/radar/radar.py
@@ -163,6 +163,7 @@ class RadarRotatingBearingRange(RadarBearingRange):
             A timestamp signifying when the rotation completes
         """
 
+        # Check if dwell_center has a timestamp instantiated if not sets it to incoming timestamp
         if self.dwell_center.timestamp is None:
             self.dwell_center.timestamp = timestamp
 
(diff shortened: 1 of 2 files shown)


stonesoup/sensor/radar/radar.py
```diff
@@ -163,6 +163,9 @@ def rotate(self, timestamp):
             A timestamp signifying when the rotation completes
         """
 
+        if self.dwell_center.timestamp == None:
```
sdhiscocks, line 166: Changing `==` to `is` should work.
```suggestion
        if self.dwell_center.timestamp is None:
```

sdhiscocks:
Thanks for the pull request @alexiad1. I also see in tests there is a blank line whitespace error.
You can run tests locally as well, with `pytest --flake8 stonesoup` (see our [development dependencies](https://github.com/dstl/Stone-Soup/blob/4e8b02e5610c5544a72a47215cd11786e72ffffc/setup.py#L33-L35)), which will highlight this.

May also be worth modifying the existing test, such that timestamp isn't set initially, to cover the change.
https://github.com/dstl/Stone-Soup/blob/4e8b02e5610c5544a72a47215cd11786e72ffffc/stonesoup/sensor/radar/tests/test_radar.py#L273-L276

alexiad1:
I made the suggested change and ran the test and found no failures. I also attempted to find out how to create a test with a uninstantiated timestamp so that the line would be covered but I was confused on how to do that.

idorrington-dstl:
> I made the suggested change and ran the test and found no failures. I also attempted to find out how to create a test with a uninstantiated timestamp so that the line would be covered but I was confused on how to do that.

In the code snippet @sdhiscocks linked to, you can just delete the timestamp argument to dwell center. The test will then use you code to set the timestamp, meaning it is covered by the tests. 

If you feel like a bigger task, you can look into parameterising that test, as is done for some of the other tests. Then we can have some tests where it is set and others where it isn't. If you don't know how to do that and don't feel like learning then I think you can just delete the timestamp argument like I said above.

sdhiscocks (approved):

idorrington-dstl (approved):
Thanks for parameterising the test as well :)

Issue #397, opened 2021-03-02
Out of Sequence Tracker - Two Tracker (One current, one delayed)


gawebb-dstl:
The idea is that you have two trackers. One running at the current time, one running ‘x’ seconds behind. If you receive an ‘old’ measurement, you run the last ‘x’ seconds of measurements through the old tracker in order. This method is computational expensive but I think it offers the optimal tracking performance for out of sequence measurements.  I think this method could be useful to have in stone-soup’s framework because it gives an upper bound on the ability of any other out of sequence measurements methods.

I an Out of Sequence measurement wrapper for a SingleTargetTracker, I kept running into problems that caused me deviate away from the stone-soup framework. I’d like to make it agnostic of the tracker

```
detection_buffer = []  # contains detections between ‘now’ and ‘now+ ‘x’ seconds
tracker_current = Tracker(x)
tracker_late = Tracker(x)

# new detection comes in
detection_buffer.append(new_detection)
sort(detection_buffer)

if new_detection.time >= tracker_now.state.time  # detection is in sequence
  tracker_now.add_detection(new_detection)
elif  new_detection.time < tracker_now.state.time  # detection out of sequence
  tracker_now = copy(tracker_late)
  tracker_now.add_detection(detection_buffer)
```

One of the major problems with this implementation is on the second to last line of the code. You can’t copy a tracker, with the ‘state’ of the generator. It restarts the generator, which breaks this implementation as we need to copy the state of the generator when we copy it

sdhiscocks:
I think this is resolved now by OOS examples in documentation.

PR #479, opened 2021-08-05
Basic Out of Sequence Measurement Tracker

gawebb-dstl:
A basic out of sequence measurement tracker that is agnostic of the type of tracker used. 

It runs two identical trackers; one current and one delayed by _x_ time. Detections within _x_ time are kept in a buffer. If an out of sequence measurement arrives, it's added to buffer, the detections in the buffer are reordered in time order, the detections are fed into the delayed tracker to get it up to date. See #397 for more detail of the idea.

This is very much just a marker to say I'm working on it. You can run the stone-soup\docs\examples\oosexample.py to get an understanding of the code.

This branch requires the iterable_trackers #415 branch as currently trackers can't be copied and keep their current state.
This uses the trackerTemplate #401 branch for the example.


gawebb-dstl:
This is very old. A new PR will be opened if a Out of Sequence Measurement Tracker is added to stone-soup

PR #484, opened 2021-08-16
Refactor YAML serialiser as ruamel.yaml plugin

sdhiscocks:
This shouldn't be a breaking change, as original class simply proxied existing methods anyway.

Also enable entry point to allow additional extensions to serialisation via entry point. This is to work over limitation of only being able to use one plugin at a time with ruamel.yaml, as well as allow easier integration with Stone Soup plugins.


sdhiscocks:
Not sure how to go about testing entry points part of this PR, but I can confirm I've tested locally, to ensure plugin aspect functions.

oharrald-Dstl (approved):

PR #490, opened 2021-08-25
Add sonar beamformers

PhilClemson:
This PR adds two new detectors for application to sonar tracking problems. Both take hydrophone data and convert these signals to a set of detections using a windowing method.

`capon` is a standard Capon beamformer. `rjmcmc` is a novel beamforming algorithm based on reversible-jump Markov chain Monte Carlo. The latter algorithm is designed for multiple target tracking, but the Metropolis-Hastings algorithm currently used does not perform well in these situations. I've therefore disabled the multiple target detection for now by forcing the update type in `proposal_func()`. The current algorithm has since been replaced with a more effective no-u-turn sampler (NUTS) but this is still undergoing testing (I'll include this in another PR at a later date).

The usage is demonstrated in the corresponding test file `test_beamformers.py`, where I've included a small simulated dataset `fixed_target_example.csv`.

### DIFF
--- a/stonesoup/detector/beamformers.py
+++ b/stonesoup/detector/beamformers.py
@@ -14,13 +14,13 @@ lists the Cartesian coordinates for each time window.
 
 import copy
 import csv
-import math
 from collections.abc import Sequence
 from datetime import datetime, timedelta
 from itertools import islice, zip_longest
 from pathlib import Path
 
 import numpy as np
+from scipy.special import factorial
 from scipy.stats import norm, uniform
 
 from ..base import Property
@@ -181,7 +181,7 @@ class RJMCMCBeamformer(DetectionReader):
             windows = grouper(reader, self.window_size)
             sensor_loc = grouper(reader_loc, self.num_sensors)
 
-            bin_steps = [math.pi/(2*self.nbins), 2*math.pi/self.nbins]
+            bin_steps = [np.pi/(2*self.nbins), 2*np.pi/self.nbins]
 
             for (window, sensor_pos) in zip(windows, sensor_loc):
                 # Grab the next `window_size` lines from the reader and read it into y (also
@@ -212,8 +212,8 @@ class RJMCMCBeamformer(DetectionReader):
 
                 for k in range(0, self.num_sensors):
                     for t in range(0, self.window_size):
-                        sinTy[k] = sinTy[k] + math.sin(2*math.pi*t*self.omega/self.fs)*y[t, k]
-                        cosTy[k] = cosTy[k] + math.cos(2*math.pi*t*self.omega/self.fs)*y[t, k]
+                        sinTy[k] = sinTy[k] + np.sin(2*np.pi*t*self.omega/self.fs)*y[t, k]
+                        cosTy[k] = cosTy[k] + np.cos(2*np.pi*t*self.omega/self.fs)*y[t, k]
                         yTy = yTy + y[t, k]*y[t, k]
 
                 sumsinsq = 0
@@ -222,14 +222,14 @@ class RJMCMCBeamformer(DetectionReader):
 
                 for t in range(0, self.window_size):
                     sumsinsq = sumsinsq \
-                        + math.sin(2*math.pi*t*self.omega/self.fs) \
-                        * math.sin(2*math.pi*t*self.omega/self.fs)
+                        + np.sin(2*np.pi*t*self.omega/self.fs) \
+                        * np.sin(2*np.pi*t*self.omega/self.fs)
                     sumcossq = sumcossq \
-                        + math.cos(2*math.pi*t*self.omega/self.fs) \
-                        * math.cos(2*math.pi*t*self.omega/self.fs)
+                        + np.cos(2*np.pi*t*self.omega/self.fs) \
+                        * np.cos(2*np.pi*t*self.omega/self.fs)
                     sumsincos = sumsincos \
-                        + math.sin(2*math.pi*t*self.omega/self.fs) \
-                        * math.cos(2*math.pi*t*self.omega/self.fs)
+                        + np.sin(2*np.pi*t*self.omega/self.fs) \
+                        * np.cos(2*np.pi*t*self.omega/self.fs)
                 sumsincos = 0
                 old_logp = self.log_prob(params, K, sinTy, cosTy, yTy,
                                          sumsinsq, sumcossq, sumsincos, N)
@@ -250,18 +250,18 @@ class RJMCMCBeamformer(DetectionReader):
                             K = copy.deepcopy(p_K)
                             for k in range(0, K):
                                 # correct for mirrored DOAs in elevation
-                                if ((params[k][0] > math.pi/2) & (params[k][0] <= math.pi)):
-                                    params[k][0] = math.pi - params[k][0]
-                                elif ((params[k][0] > math.pi) & (params[k][0] <= 3*math.pi/2)):
-                                    params[k][0] = params[k][0] - math.pi
-                                    params[k][1] = params[k][1] - math.pi
-                                elif ((params[k][0] > 3*math.pi/2) & (params[k][0] <= 2*math.pi)):
-                                    params[k][0] = 2*math.pi - params[k][0]
-                                    params[k][1] = params[k][1] - math.pi
+                                if ((params[k][0] > np.pi/2) & (params[k][0] <= np.pi)):
+                                    params[k][0] = np.pi - params[k][0]
+                                elif ((params[k][0] > np.pi) & (params[k][0] <= 3*np.pi/2)):
+                                    params[k][0] = params[k][0] - np.pi
+                                    params[k][1] = params[k][1] - np.pi
+                                elif ((params[k][0] > 3*np.pi/2) & (params[k][0] <= 2*np.pi)):
+                                    params[k][0] = 2*np.pi - params[k][0]
+                                    params[k][1] = params[k][1] - np.pi
                                 if (params[k][1] < 0):
-                                    params[k][1] += 2*math.pi
-                                elif (params[k][1] > 2*math.pi):
-                                    params[k][1] -= 2*math.pi
+                                    params[k][1] += 2*np.pi
+                                elif (params[k][1] > 2*np.pi):
+                                    params[k][1] -= 2*np.pi
                         for k in range(0, K):
                             bin_ind = [0, 0]
                             for ind in range(0, 2):
@@ -412,20 +412,18 @@ class RJMCMCBeamformer(DetectionReader):
         for k in range(0, p_K):
             # calculate phase offsets relative to first sensor in the array
             for sensor_ind in range(0, self.num_sensors):
-                alpha = 2*math.pi*self.omega*((self.sensor_pos[sensor_ind, 1]
-                                               - self.sensor_pos[0, 1]) * math.sin(p_params[k][1])
-                                              * math.sin(p_params[k][0])
-                                              + (self.sensor_pos[sensor_ind, 0]
-                                                 - self.sensor_pos[0, 0])
-                                              * math.cos(p_params[k][1])
-                                              * math.sin(p_params[k][0])
-                                              + (self.sensor_pos[sensor_ind, 2]
-                                                 - self.sensor_pos[0, 2])
-                                              * math.sin(p_params[k][0])) / self.wave_speed
-                DTy[k] = DTy[k] + math.cos(alpha) * sinTy[sensor_ind] \
-                    + math.sin(alpha) * cosTy[sensor_ind]
-                sinalpha[k, sensor_ind] = math.sin(alpha)
-                cosalpha[k, sensor_ind] = math.cos(alpha)
+                alpha = 2*np.pi*self.omega*(
+                    (self.sensor_pos[sensor_ind, 1] - self.sensor_pos[0, 1])
+                    * np.sin(p_params[k][1])
+                    * np.sin(p_params[k][0])
+                    + (self.sensor_pos[sensor_ind, 0] - self.sensor_pos[0, 0])
+                    * np.cos(p_params[k][1])
+                    * np.sin(p_params[k][0])
+                    + (self.sensor_pos[sensor_ind, 2] - self.sensor_pos[0, 2])
+                    * np.sin(p_params[k][0])) / self.wave_speed
+                DTy[k] = DTy[k] + np.cos(alpha)*sinTy[sensor_ind] + np.sin(alpha)*cosTy[sensor_ind]
+                sinalpha[k, sensor_ind] = np.sin(alpha)
+                cosalpha[k, sensor_ind] = np.cos(alpha)
 
         for k1 in range(0, p_K):
             DTD[k1, k1] = N/2
@@ -443,9 +441,9 @@ class RJMCMCBeamformer(DetectionReader):
 
         Dterm = np.matmul(np.linalg.solve(1001*DTD, DTy), np.transpose(DTy))
         log_posterior = - (p_K * np.log(1.001) / 2) - (N / 2) * np.log((yTy - Dterm) / 2) \
-            + p_K * np.log(self.Lambda) - np.log(np.math.factorial(p_K)) \
-            - p_K*np.log(math.pi * math.pi)
-        # note: math.pi*math.pi comes from area of parameter space in one dimension (i.e. range of
+            + p_K * np.log(self.Lambda) - np.log(factorial(p_K)) \
+            - p_K*np.log(np.pi * np.pi)
+        # note: np.pi*np.pi comes from area of parameter space in one dimension (i.e. range of
         # azimuth * range of elevation)
 
         return log_posterior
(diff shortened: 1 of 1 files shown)


stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,483 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+import math
+import random
+import numba
```
sdhiscocks, line 5: I'm not sure we want to include `numba` dependency for the one function below. Suggest removing this for now.

Separately I'd be interested if we could use `numba` in certain places however.

PhilClemson: Good spot @sdhiscocks, I'd mistakenly left that function in from an earlier version which was doing some complex number operations. I've removed both `numba` and the function now.

stonesoup/detector/tests/test_beamformers.py
```diff
@@ -0,0 +1,64 @@
...
+from stonesoup.models.measurement.linear import LinearGaussian
+from stonesoup.types.array import CovarianceMatrix
+from stonesoup.predictor.kalman import KalmanPredictor
+from stonesoup.updater.kalman import KalmanUpdater
+from stonesoup.types.state import GaussianState
+from stonesoup.types.hypothesis import SingleHypothesis
+from stonesoup.types.track import Track
+
+
+def fixed_target_beamformer_test():
```
sdhiscocks, line 18: Looks like this didn't run. Test functions must start with `test_`
```suggestion
def test_fixed_target_beamformer():
```

PhilClemson: I've now corrected this (sorry, not familiar with PyTest).

I also noticed that I'd left in the hardcoded value of the signal frequency for testing, which I've now removed.

PhilClemson:
I have now added some descriptions / references above the beamformer classes. I also added a user-set random seed so that the RJMCMC algorithm will give reproducible output (for testing).

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,505 @@
...
+
+        y = np.loadtxt(self.csv_path, delimiter=',')
+
+        L = len(y)
+
+        self.num_sensors = int(np.matrix(self.sensor_loc).size/3)
+
+        N = self.num_sensors*window
+
+        max_targets = 5
```
sdhiscocks, line 273: Should this be configurable?

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,505 @@
...
+        L = len(y)
+
+        window = 1000  # size of sliding window in samples
+        windowm1 = window-1
+
+        thetavals = np.linspace(0, 2*math.pi, num=400)
+        phivals = np.linspace(0, math.pi/2, num=100)
+
+        # spatial locations of hydrophones
+        z = np.matrix(self.sensor_loc)
```
sdhiscocks, line 169: `np.matrix` is deprecated. Should use `np.array` instead.

stonesoup/detector/tests/test_beamformers.py
```diff
@@ -0,0 +1,64 @@
...
+from stonesoup.models.measurement.linear import LinearGaussian
+from stonesoup.types.array import CovarianceMatrix
+from stonesoup.predictor.kalman import KalmanPredictor
+from stonesoup.updater.kalman import KalmanUpdater
+from stonesoup.types.state import GaussianState
+from stonesoup.types.hypothesis import SingleHypothesis
+from stonesoup.types.track import Track
+
+
+def test_fixed_target_beamformer():
```
sdhiscocks, line 18: Could parameterise this test:

```patch
diff --git a/stonesoup/detector/tests/test_beamformers.py b/stonesoup/detector/tests/test_beamformers.py
index fb990f2b..312d31ab 100644
--- a/stonesoup/detector/tests/test_beamformers.py
+++ b/stonesoup/detector/tests/test_beamformers.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
-import numpy as np
 from os import path
 from datetime import datetime
 
+import numpy as np
+import pytest
+
 from stonesoup.detector import beamformers
 from stonesoup.models.transition.linear import CombinedLinearGaussianTransitionModel, \
                                                ConstantVelocity
@@ -15,9 +17,16 @@ from stonesoup.types.hypothesis import SingleHypothesis
 
 
-def test_fixed_target_beamformer():
-
+@pytest.fixture(params=[beamformers.capon, beamformers.rjmcmc])
+def detector(request):
     data_file = path.join(path.dirname(__file__), "fixed_target_example.csv")
+    extra_args = {'seed': 12345} if request.param is beamformers.rjmcmc else {}
+    return request.param(data_file, sensor_loc="0 0 0; 0 10 0; 0 20 0; 10 0 0; 10 10 0; \
+        10 20 0; 20 0 0; 20 10 0; 20 20 0", fs=2000, omega=50, wave_speed=1481, **extra_args)
+
+
+def test_fixed_target_beamformer(detector):
+
     truth = [0.8, 0.2]
 
     transition_model = CombinedLinearGaussianTransitionModel([ConstantVelocity(0.01),
@@ -32,27 +41,9 @@ def test_fixed_target_beamformer():
 
     prior = GaussianState([[0.5], [0], [0.5], [0]], np.diag([1, 0, 1, 0]),
                           timestamp=datetime.now())
+    track = Track()
 
-    detector1 = beamformers.capon(data_file, sensor_loc="0 0 0; 0 10 0; 0 20 0; 10 0 0; 10 10 0; \
-        10 20 0; 20 0 0; 20 10 0; 20 20 0", fs=2000, omega=50, wave_speed=1481)
-    detector2 = beamformers.rjmcmc(data_file, sensor_loc="0 0 0; 0 10 0; 0 20 0; 10 0 0; 10 10 0; \
-        10 20 0; 20 0 0; 20 10 0; 20 20 0", fs=2000, omega=50, wave_speed=1481, seed=12345)
-
-    track1 = Track()
-    track2 = Track()
-
-    for timestep, detections in detector2:
-        for detection in detections:
-            # check beamformer accuracy
-            assert(np.abs(detection.state_vector[0] - truth[0]) < 0.4)
-            assert(np.abs(detection.state_vector[1] - truth[1]) < 0.1)
-            prediction = predictor.predict(prior, timestamp=detection.timestamp)
-            hypothesis = SingleHypothesis(prediction, detection)  # Group prediction + measurement
-            post = updater.update(hypothesis)
-            track2.append(post)
-            prior = track2[-1]
-
-    for timestep, detections in detector1:
+    for timestep, detections in detector:
         for detection in detections:
             # check beamformer accuracy
             assert(np.abs(detection.state_vector[0] - truth[0]) < 0.4)
@@ -60,5 +51,5 @@ def test_fixed_target_beamformer():
             prediction = predictor.predict(prior, timestamp=detection.timestamp)
             hypothesis = SingleHypothesis(prediction, detection)  # Group prediction + measurement
             post = updater.update(hypothesis)
-            track1.append(post)
-            prior = track1[-1]
+            track.append(post)
+            prior = track[-1]
```

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,505 @@
...
+    [2] C. Andrieu and A. Doucet, Joint Bayesian Model Selection and Estimation of
+    Noisy Sinusoids via Reversible Jump MCMC, IEEE Trans. Signal Process. 47(10):2667-2676 (1999)
+    [3] C. Andrieu, N. de Freitas and A. Doucet, Robust Full Bayesian Learning for Radial Basis
+    Networks, Neural Computation 13:2359–2407 (2001)
+
+    """
+    csv_path: str = Property(doc='The path to the csv file, containing the raw data')
+    fs: float = Property(doc='Sampling frequency (Hz)')
+    omega: float = Property(doc='Signal frequency (Hz)')
+    sensor_loc: str = Property(doc='Cartesian coordinates of the sensors in the format\
```
sdhiscocks, line 245: I think a `Sequence[StateVector]` would be better for both beamformers.

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,505 @@
...
+import numpy as np
+import math
+import random
+import copy
+from datetime import datetime, timedelta
+from stonesoup.base import Property, Base
+from stonesoup.buffered_generator import BufferedGenerator
+from stonesoup.models.measurement.linear import LinearGaussian
+from stonesoup.types.array import StateVector, CovarianceMatrix
+from stonesoup.types.detection import Detection
```
sdhiscocks, line 11:
```suggestion
import copy
import math
import random
from datetime import datetime, timedelta

import numpy as np

from ..base import Property, Base
from ..buffered_generator import BufferedGenerator
from ..models.measurement.linear import LinearGaussian
from ..types.array import StateVector, CovarianceMatrix
from ..types.detection import Detection
```

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,505 @@
...
+
+
+def proposal(params, K, p_params):
```
sdhiscocks, line 14: Are these standalone functions only used by `rjmcmc`? If so, I'd add them as class/static methods to the `rjmcmc` class.

sdhiscocks (changes requested):
I'm not familiar with beamforming, but overall looks good.

Code coverage could be improved with some additional tests if possible. Parameterising the tests may make this easier. But may be limited by test data you have available?

This is required to build documentation:
```patch
diff --git a/docs/source/stonesoup.detector.rst b/docs/source/stonesoup.detector.rst
index f0e86d4b..5a3f3e74 100644
--- a/docs/source/stonesoup.detector.rst
+++ b/docs/source/stonesoup.detector.rst
@@ -10,4 +10,8 @@ Detectors
 .. automodule:: stonesoup.detector.tensorflow
     :show-inheritance:
 
+Beamformers
+-----------
 
+.. automodule:: stonesoup.detector.beamformers
+    :show-inheritance:

```

stonesoup/detector/beamformers.py
PhilClemson: Just trying to understand this - is the issue that I'm passing the csv file / data directly into the detector rather than using a reader and passing everything as a `Sequence[StateVector]`?

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,544 @@
...
+    [2] C. Andrieu and A. Doucet, Joint Bayesian Model Selection and Estimation of
+    Noisy Sinusoids via Reversible Jump MCMC, IEEE Trans. Signal Process. 47(10):2667-2676 (1999)
+    [3] C. Andrieu, N. de Freitas and A. Doucet, Robust Full Bayesian Learning for Radial Basis
+    Networks, Neural Computation 13:2359–2407 (2001)
+
+    """
+    path: str = Property(doc='The path to the csv file, containing the raw data')
+    fs: float = Property(doc='Sampling frequency (Hz)')
+    omega: float = Property(doc='Signal frequency (Hz)')
+    sensor_loc: list = Property(doc='Cartesian coordinates of the sensors in the format\
```
oharrald-Dstl, line 158: For `sensor_loc`, having a list of state vectors might be better `Sequence[StateVector]` (as mentioned by SH... I think)

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,544 @@
...
+                covar = CovarianceMatrix(np.array([[1, 0], [0, 1]]))
+                measurement_model = LinearGaussian(ndim_state=4, mapping=[0, 2],
+                                                   noise_covar=covar)
+                current_time = current_time + timedelta(milliseconds=1000*L/self.fs)
+                detection = Detection(state_vector, timestamp=current_time,
+                                      measurement_model=measurement_model)
+
+                yield current_time, {detection}
+
+    def log_prob(self, p_noise, p_params, p_K, y, T, sinTy, cosTy, yTy, sumsinsq, sumcossq,
```
oharrald-Dstl, line 379: Some documentation of this method would be good (and others in this class).

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,544 @@
...
+
+            # wrap angles
+            if p_params[k, 0] < 0:
+                p_params[k, 0] = p_params[k, 0] + 2*math.pi
+            elif p_params[k, 0] > 2*math.pi:
+                p_params[k, 0] = p_params[k, 0] - 2*math.pi
+            if p_params[k, 1] < 0:
+                p_params[k, 1] = p_params[k, 1] + 2*math.pi
+            elif p_params[k, 1] > 2*math.pi:
+                p_params[k, 1] = p_params[k, 1] - 2*math.pi
```
oharrald-Dstl, line 466: There are `Bearing` and `Elevation` types that could be used for this.

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,544 @@
...
+
+        return p_params
+
+    def noise_proposal(self, noise):
+        epsilon = random.gauss(0, 0.1)
+        rand_val = abs(noise+epsilon)
+        p_noise = copy.deepcopy(rand_val)
```
oharrald-Dstl, line 473: Why is a deep copy necessary here?

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,544 @@
...
+    seed: int = Property(doc='Random number generator seed for reproducible output. Set to 0 for\
+                         non-reproducible output')
+    window_size: int = Property(doc='Wndows size', default=750)
+
+    def __init__(self, path, *args, **kwargs):
+        if not isinstance(path, Path):
+            path = Path(path)
+        super().__init__(path, *args, **kwargs)
+        if self.seed != 0:
+            random.seed(a=self.seed, version=2)
```
oharrald-Dstl, line 171: This will set the global random seed, fixing output for entire simulation.
Consider method used in #444 and #461 for locally fixing random seed.

oharrald-Dstl (changes requested):

stonesoup/detector/beamformers.py
sdhiscocks: It's the fact that your passing in sensor locations as a string of numbers, rather than just passing the numbers as an array, or list of arrays for example.

PhilClemson: Got it, thanks! This has been updated in the latest version.

stonesoup/detector/beamformers.py
PhilClemson: Thanks for this, I wasn't aware these existed. I've updated the parameter arrays to be StateVector[Elevation, Bearing]. However, I don't think I can use the `mod_angle` functions (at least for the elevation) as it's tied in with the MCMC algorithm, which works on the range [0, 2pi] for all angles due to the requirements of the method.

PhilClemson:
Okay, I think I've made all the requested changes but let me know if I missed anything!

sdhiscocks (changes requested):

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,560 @@
...
+                        max_ind = ind
+
+                # look for largest N peaks, where N corresponds to peak in the order histogram
+                # use divide-and-conquer quadrant-based approach
+                if max_ind == 0:
+                    # only one target
+                    [unique_peak_inds1, unique_peak_inds2] = np.unravel_index(
+                                                             param_hist[0, :, :].argmax(),
+                                                             param_hist[0, :, :].shape)
+                    num_peaks = 1
```
sdhiscocks, line 292: Variable isn't used.

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,560 @@
...
+                covar = CovarianceMatrix(np.array([[1, 0], [0, 1]]))
+                measurement_model = LinearGaussian(ndim_state=4, mapping=[0, 2],
+                                                   noise_covar=covar)
+                current_time = current_time + timedelta(milliseconds=1000*L/self.fs)
+                detection = Detection(state_vector, timestamp=current_time,
+                                      measurement_model=measurement_model)
+
+                yield current_time, {detection}
+
+    def log_prob(self, p_noise, p_params, p_K, y, T, sinTy, cosTy, yTy, sumsinsq, sumcossq,
```
sdhiscocks, line 373: `p_noise`, `y` and `T` aren't used

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,560 @@
...
+        """The proposal function for the RJMCMC moves. A coin toss is used to decide whether an
+        update move or revesible-jump move is proposed. If a reversible-jump move is chosen, a
+        further coin toss decides whether the proposed move is a birth or death move (adding or
+        removing a target).
+
+        """
+        update_type = uniform.rvs(random_state=self.random_state)
+        p_K = 0
+        Qratio = 1  # ratio of proposal probabilities for forwards and backwards moves
+        update_type = 1  # forced temporarily (for single-target examples)
```
sdhiscocks, line 522: `update_type` is defined twice.

stonesoup/detector/beamformers.py
```diff
@@ -0,0 +1,560 @@
...
+        p_noise = rand_val
+        return p_noise
+
+    def proposal(self, params, K, p_params):
+        """The proposal function for direction of arrival parameters. The parameters are initially
+        sampled from a uniform distribution. Gaussian probability densities are used to make new
+        proposals centred on the current parameter values.
+
+        """
+        p_K = 0
```
sdhiscocks, line 487: Not required

PhilClemson:
Thanks again for looking at this @sdhiscocks and apologies for taking so long to get back to you with the changes (unfortunately Christmas got in the way and then I was busy with other work in January).

I think I've fixed all the issues that were mentioned but it might be worth checking the new reader implementation with `itertools` is usable for streaming data.

PR #494, opened 2021-09-08
Added test_image

sgjkang4:
Added test_image class that test the single image using cfar or ccl


setup.py
```diff
@@ -27,7 +27,7 @@
       use_scm_version=True,
       install_requires=[
           'ruamel.yaml>=0.15.45', 'numpy>=1.17', 'scipy', 'matplotlib', 'utm', 'pymap3d', 'ordered-set',
-          'setuptools>=42', 'rtree',
+          'setuptools>=42', 'rtree', 'opencv-python'
```
sdhiscocks, line 30: I think this may be better in `extra_requires`. There are examples for this for `video`/`tensorflow`.
You can raise a warning on the relevant imports, like the tensorflow notebooks.

https://github.com/dstl/Stone-Soup/blob/c806374c4071071919be1b43cbe83db286873df7/stonesoup/detector/tensorflow.py#L5-L13

sdhiscocks (changes requested):
One change needed for dependencies, but otherwise looks good :+1:

setup.py
sglvladi: @sdhiscocks I'll be picking up this PR, since Charlotte has now finished her internship. Do you think this should go under the video extras?

sdhiscocks: Either _video_ or many a new group. _video_ may be best, so we don't have too many, and it makes sense for `opencv`

sdhiscocks (approved):

sdhiscocks:
Just spotted that the pickle files are fairly large. Wonder if there is an alternative? [numpy.savez](https://numpy.org/doc/stable/reference/generated/numpy.savez.html#numpy.savez)? Or maybe check properties of the resultant array (sum of array or diagonal matches expects)?

sdhiscocks:
> [numpy.savez](https://numpy.org/doc/stable/reference/generated/numpy.savez.html#numpy.savez)?

Compressed option [numpy.savez_compressed]( https://numpy.org/doc/stable/reference/generated/numpy.savez_compressed.html#numpy.savez_compressed)

sglvladi:
> Just spotted that the pickle files are fairly large. Wonder if there is an alternative? [numpy.savez](https://numpy.org/doc/stable/reference/generated/numpy.savez.html#numpy.savez)? Or maybe check properties of the resultant array (sum of array or diagonal matches expects)?

@sdhiscocks Good spot! I have cropped the test image, which leads to reduced size of the image file and the pickle files. Hope this does the trick. I also noticed that there was no documentation for the introduced classes, so I've added that as well.

Issue #326, opened 2020-11-11
SensorManager proposal


jmbarr:
![sensormanagerclass_v0 1](https://user-images.githubusercontent.com/31851985/98871316-9c2a0700-246c-11eb-9dce-8b3de89f7f60.png)

v0.1 of a proposed sensor manager class. Hopefully my hack-UML makes sense. Briefly, the new class encompasses a method which takes a set of possible sensor actions (_actionset_), a measurement model per sensor (perhaps derived from a sensor object, perhaps held internally), a set of state predictions (which reflect the sensor manager's view of the world), and an optimiser. This optimiser is a functional class which the sensor manager uses to return the chosen action. It requires a measure to optimise over (in future maybe more than one). 

An example of an optimiser might be a brute-force search, a measure could be estimated distance to target. A sensor manager combining these will query every track and observe the one estimated to be closest. Not a brilliant way of doing things, but an example. 

Red stars indicate classes or methods that don't yet exist. 

Comments; queries?

apiszcz:
Thought, consider adaptive methods for sensor control to optimize power levels/clutter level, etc. Assume this may be possible in the optimiser.

jmbarr:
> Thought, consider adaptive methods for sensor control to optimize power levels/clutter level, etc. Assume this may be possible in the optimiser.

Yes, I think the optimiser should be able to do constrained optimisation (or optimization if you like).  How exactly it does that is up for grabs. I'll try and create a class that solves some canonical problem and post for comment.

sdhiscocks:
I think it would be good to replace `StatePrediction` with `Track`, as then you'll have access to the entire track history, and track meta data. I can imagine use cases where this would be useful.

jmbarr:
![sensormanagerclass_v0 2](https://user-images.githubusercontent.com/31851985/100149047-83bae300-2e95-11eb-8297-74f1d824132e.png)


> I think it would be good to replace `StatePrediction` with `Track`, as then you'll have access to the entire track history, and track meta data. I can imagine use cases where this would be useful.

Of course you still need to make a prediction. So `Track` would have to come with `Predictor`, which would give you a `StatePrediction`.

jmbarr:
> ![sensormanagerclass_v0 2](https://user-images.githubusercontent.com/31851985/100149047-83bae300-2e95-11eb-8297-74f1d824132e.png)
> 
[...]

PR #503, opened 2021-09-28
Add Sensor Management

nperree-dstl:
Add sensor management capabilities to the main stone soup branch. 

This includes base classes for sensor management algorithms and examples of classes for optimized sensor management algorithms. It also includes an example of an uncertainty based reward function and actionable sensors compatible with the sensor management algorithms.

Covers some of the concepts discussed in issue #326 and discussion #441.


stonesoup/sensor/action/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+        return getattr(self.owner, self.attribute)
+
+    @property
+    def default_action(self):
+        """The default action to modify the property if there is no given action."""
+        raise NotImplementedError()
+
+
+class RealNumberActionGenerator(ActionGenerator):
+
```
jmbarr, line 75: Provide a description - what's this for?

sdhiscocks: This was to represent an action which is a choice of a real number (i.e. float) between and min and max value. This was to support optimisation where you want to constrain search space, without considering all options and smallest resolution.

I expect we'll also add a binary action generator, or vector action (e.g. position) in future.

stonesoup/sensormanager/base.py
```diff
@@ -0,0 +1,155 @@
...
+        raise NotImplementedError
+
+
+class RandomSensorManager(SensorManager):
+    """As the name suggests, a sensor manager which returns a random choice of action or actions
+    from the list available. Its practical purpose is to serve as a baseline to test against.
+
+    """
+
+    sensors: Set[Sensor] = Property(doc="The sensor(s) which the sensor manager is managing. "
```
jmbarr, line 59: Defined again? Isn't `sensors` attribute inherited?

docs/examples/Creating_Actionable_Sensor.py
```diff
@@ -0,0 +1,266 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+Creating an Actionable Sensor Example
```
jmbarr, line 5: Screenshot 2021-11-18 at 14 24 19

oharrald-Dstl: Odd. I can't see why that's being raised.

nperree-dstl: That's weird - it seems to load fine for me!

stonesoup/sensor/actionable.py
```diff
@@ -0,0 +1,179 @@
...
+        NotImplementedError
+            If sensor cannot be tasked.
+
+        Notes
+        -----
+        Base class returns True
+        """
+
+        if not self.validate_timestamp():
+            return
```
jmbarr, line 124: Does this mean it returns neither True or False if timestamp is invalid? If so should mention that possibility in the 'Returns' section above.

stonesoup/sensor/actionable.py
```diff
@@ -0,0 +1,179 @@
...
+        """Carry out actions up to a timestamp.
+
+        Parameters
+        ----------
+        timestamp: datetime.datetime
+            Carry out actions up to this timestamp.
+        """
+
+        if not self.validate_timestamp():
+            self.timestamp = timestamp
```
jmbarr, line 149: why set the timestamp here?

stonesoup/sensor/radar/radar.py
```diff
@@ -14,12 +14,14 @@
 from ...models.measurement.nonlinear import \
     (CartesianToBearingRange, CartesianToElevationBearingRange,
      CartesianToBearingRangeRate, CartesianToElevationBearingRangeRate)
+from ...sensor.actionable import ActionableProperty
```
jmbarr, line 17: Not sure how the changes in this file belong in sensor management branch

stonesoup/sensormanager/optimise.py
```diff
@@ -0,0 +1,102 @@
...
+
+import numpy as np
+from scipy.optimize import brute, basinhopping
+
+from . import BruteForceSensorManager
+from ..base import Property
+
+
+class _OptimizeSensorManager(BruteForceSensorManager):
+
```
jmbarr, line 12: I assume this gets left out of the Docs?

jmbarr: Otherwise needs an explanation.

stonesoup/sensormanager/optimise.py
```diff
@@ -0,0 +1,102 @@
...
+    returning the optimal configuration.
+
+    SciPy optimize provides functions which can minimize or maximize functions using a variety
+    of algorithms. The :meth:`scipy.optimize.brute` minimizes a function over a given range,
+    using a brute force method. This is done by computing the function's value at each point of
+    a multidimensional grid of points, to find the global minimum.
+
+    This brute force method also applies a polishing function to the result of
+    the brute force minimization. By default this is set as :func:`scipy.optimize.fmin` which
+    minimizes a function using the downhill simplex algorithm.
```
jmbarr, line 61: Could add a reference. Or a link to the scipy.optimize help.

stonesoup/sensormanager/tests/test_sensormanager.py
```diff
@@ -0,0 +1,167 @@
...
+            fov_angle=np.radians(30),
+            dwell_centre=StateVector([0.0]),
+            max_range=np.inf,
+        )}
+
+        for sensor_set in [sensorsA, sensorsB, sensorsC]:
+            for sensor in sensor_set:
+                sensor.timestamp = time_start
+
+        sensor_managerA = BruteForceSensorManager(sensorsA, reward_function.calculate_reward)
```
jmbarr, line 127: So the part where we pass `reward_function.calculate_reward` rather than just `reward_function` suggests to me that we don't need to provide `RewardFunction` classes in the codebase. They're just classes which execute a function no? So just provide the function - or just provide an example in a notebook and leave the crafting of the particular function to the experts.

jmbarr (changes requested):
Overall delivers what it should as far as the core sensor management stuff goes. But there are a lot of sensor-focussed adaptions which I don't follow and would have preferred in a separate PR. 
I do think the actionable sensor example should be clearer though.

stonesoup/sensormanager/tests/test_sensormanager.py
nperree-dstl: We don't necessarily need to provide a reward function class, it's there more as an example of how you can set one up that works with the rest of the sensor management structure and it is convenient to just import the class in notebooks. I could take it out and just include it in the tutorials (PR to follow this one) if that would be better? Or I can include it as just a function but I think the uncertainty reward function works nicely as a class so you don't need to pass a predictor/updater every time step.

stonesoup/sensormanager/optimise.py
nperree-dstl: Nope it's not included in the docs. I've updated the documentation slightly to make it clearer how the available classes work

docs/examples/Creating_Actionable_Sensor.py
```diff
@@ -0,0 +1,266 @@
...
+# one time.
+# To do so, we require an :class:`~.Action` and :class:`~.ActionGenerator`.
+# The action contains the logic of how a particular sensor property is modified. In this instance
+# we simply want to change the what sector the sensor is looking at, at a given time.
+# When the sensor is called to `act`, it calls `act` on the scheduled actions of each of its
+# properties that are :class:`~.ActionableProperty` types (the current implementation allows for
+# one scheduled action per actionable property in a sensor).
+#
+# The action's `act` method should return the new value of the modifiable property for the sensor.
+# The modfiable property for our sensor will be its `sector`. Therefore, our `act` code can be as
```
orosoman-dstl, line 24: *modifiable

oharrald-Dstl (approved):

stonesoup/sensormanager/base.py
```diff
@@ -0,0 +1,147 @@
...
+    """A sensor manager which returns a choice of action from those available. The sensor manager
+    iterates through every possible configuration of sensors and actions and
+    selects the configuration which returns the maximum reward as calculated by a reward function.
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def choose_actions(self, tracks_list, timestamp, nchoose=1, *args, **kwargs):
```
sdhiscocks, line 102: Does this need tracks in a list?

stonesoup/sensormanager/base.py
```diff
@@ -0,0 +1,147 @@
...
+    There is considerable freedom in both the theory and practice of sensor management and these
+    classes do not enforce a particular solution. A sensor manager may be 'centralised' in that
+    it controls the actions of multiple sensors, or individual sensors may have their own managers
+    which communicate with other sensor managers in a networked fashion.
+
+    """
+    sensors: Set[Sensor] = Property(doc="The sensor(s) which the sensor manager is managing. "
+                                        "These must be capable of returning available actions.")
+
+    reward_function: Callable = Property(
```
sdhiscocks, line 31: I think this needs better definition so it's clear what form this callable should be in. May be worth adding a definition of RewardFunction type. Something like:

```python
RewardFunction = Callable[
    [Mapping[Sensor, Sequence[Action]], Set[Track], datetime.datetime],
    float
    ]
```

stonesoup/sensormanager/base.py
```diff
@@ -0,0 +1,147 @@
...
+        default=None, doc="A function or class designed to work out the reward associated with an "
+                          "action or set of actions. This may also incorporate a notion of the "
+                          "cost of making a measurement. The values returned may be scalar or "
+                          "vector in the case of multi-objective optimisation. Metrics may be of "
+                          "any type and in any units.")
+
+    @abstractmethod
+    def choose_actions(self, *args, **kwargs):
```
sdhiscocks, line 39: I think its worth defining the inputs for this method to try and standardise it.

nperree-dstl: Good point - the RandomSensorManager takes in tracks_list even though it's not required, do we want to keep it that way and have the standard format as (self, tracks_list, timestamp, nchoose=1, *args, **kwargs) ? Alternatively, timestamp is the only input that I can think would always be required so I can just add that in

stonesoup/sensormanager/base.py
nperree-dstl: We do have a RewardFunction base class to go with the UncertaintyRewardFunction which defines reward functions in a bit more detail - could we use this here?

stonesoup/sensormanager/base.py
sdhiscocks: I think we should try to standardise it. No issues with RandomSensorManager just ignoring the variables it doesn't need.

stonesoup/sensormanager/base.py
sdhiscocks: Could do, but would make it less clear that you could just pass any callable/function.

Maybe we could make the `RewardFunction` class a Callable, and then define it with the above. That would make it cover both use cases.

stonesoup/sensormanager/base.py
jmbarr: Qs: what would you standardise at? Isn't `def choose_actions(self, *args, **kwargs):` standard (albeit flexible)?

I'd prefer not to pass variables that aren't used (yes, I may be responsible for that happening with the RandomSensorManager). It may not have an impact computationally, but it could end up being thoroughly confusing.

sdhiscocks: So we can leave space to expand, but having at least a base set of inputs (so like in the above: tracks and timestamps, and number of choices) then we can create a number of components that will be broadly compatible and interchangeable.

nperree-dstl: We could have timestamps and no. choices as base set of inputs and remove tracks_list (also remove from random one)?

stonesoup/sensormanager/base.py
nperree-dstl: No it probably doesn't, would it be better as a set or collection?

stonesoup/sensormanager/base.py
jmbarr: timestamp makes sense - you'd want to know when the choice is applicable. nchoose=1 would also seem an appropriate default.

sdhiscocks: I think `tracks` should be in the base set of inputs, and its dynamic information that represents your current understanding (noting that random sensor manager is really a special case, as making random choices is arguably not managing anything :wink: )

stonesoup/sensormanager/base.py
sdhiscocks: I think `Set` as that's what commonly used, and just call the variable `tracks`.

stonesoup/sensor/actionable.py
```diff
@@ -0,0 +1,179 @@
...
+    def _default_action(self, name, property_, timestamp):
+        """Returns the default action of the action generator associated with the property
+        (assumes the property is an :class:`~.ActionableProperty`."""
+        generator = property_.generator_cls(owner=self,
+                                            attribute=name,
+                                            start_time=self.timestamp,
+                                            end_time=timestamp)
+        return generator.default_action
+
+    def actions(self, timestamp: datetime.datetime, start_timestamp: datetime.datetime = None
```
oharrald-Dstl, line 67: The action-generators that this method creates will determine their corresponding actionable attribute initial value with `getattr(self.owner, self.attribute)`. That is the value of the attribute at the sensor's "current" timestamp, and not necessarily at the `start_timestamp` that is passed to this method. So this `start_timestamp` argument is irrelevant unless we change logic in the action generators so that they calculate what the attribute value WILL be at `start_timestamp` and determine possible actions from there.
I can't recall why we required this argument in the first place.

stonesoup/sensor/actionable.py
```diff
@@ -0,0 +1,179 @@
...
+                                                   end_time=timestamp))
+        return generators
+
+    def add_actions(self, actions: Sequence[Action]) -> bool:
+        """Add actions to the sensor
+
+        Parameters
+        ----------
+        actions: sequence of :class:`~.Action`
+            Sequence of actions that will be executed in order
```
oharrald-Dstl, line 106: There's no need for the actions to be in a sequence, a set would do fine. They aren't necessarily executed in the order they're passed as it shouldn't matter.

sdhiscocks: I can imagine reasonable use cases where order is important however e.g. look here, take measurement.

stonesoup/sensormanager/base.py
sdhiscocks: One goal of the framework is to enable ability to have a set of common components that can be easily interchanged. Whilst excepting that all possible combinations and use cases can't be covered, I think that have tracks as part of that initial baseline is reasonable (and otherwise the existing example we have is itself an exception).

stonesoup/sensor/actionable.py
oharrald-Dstl: True, we don't have that logic in the current implementation since we're actioning based on a schedule dictionary. Perhaps in a subclass? Or change the schedule to be ordered?

sdhiscocks: So true the current implementation of simulated sensors don't have a schedule as you say, but thing we are defining here is the interface, and such other implementations in future could decide to care about the order. I think we can leave the current implementation for now.

stonesoup/sensormanager/base.py
jmbarr: > One goal of the framework is to enable ability to have a set of common components that can be easily interchanged. Whilst excepting that all possible combinations and use cases can't be covered, I think that have tracks as part of that initial baseline is reasonable (and otherwise the existing example we have is itself an exception).

but we do have the random sensor manager in the existing example.

Maybe my understanding isn't correct but you're advocating:
`def choose_actions(self, tracks, timestamp, nchoose, *args, **kwargs):`
over 
`def choose_actions(self, *args, **kwargs)`. Is that right? (or close?)

I'd argue that the first example does not promote easy interchange for those (many) sensor managers that are not track-focussed. (Accepting that the difference is only passing an empty set vs passing an extra keyword, I think the preference should be for general, simple, base classes.) But maybe I'm missing something. Is there an obvious problem with a user-defined `choose_actions(timestamp, tracks=some_tracks)` (say)?

sdhiscocks: Yes. My main issue is with `*args` as positional arguments really need to be fixed (as position dependent). Maybe having `choose_actions(timestamp, nchoose, **kwargs)`? That works for your example, just need to be sure that keyword arguments are passed.

nperree-dstl: So put timestamp and nchoose into the base class function, and remove *args from all choose_actions() ?

jmbarr: > So put timestamp and nchoose into the base class function, and remove *args from all choose_actions() ?

Happy with that (nchoose=1 as default?). 

Realised I didn't really know what *args does.

docs/examples/Creating_Actionable_Sensor.py
```diff
@@ -0,0 +1,266 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
nperree-dstl: Oli made some changes to the example since this review - I think the content is now good to be merged with this PR but I agree it could still be improved with additional diagrams etc. I have created an issue for this #633 so I remember to revisit it.

jswright-dstl (approved):

jmbarr (approved):

PR #504, opened 2021-10-01
Add hidden Markov model implementation for classification

oharrald-Dstl:
Successive indirect observations of a target's hidden class are taken in order to attain a categorical distribution over a (discrete) set of possible hidden classes.

This branch introduces the 'categorical' components originally added in #460.

An example has been added to demonstrate use of the introduced components.

### DIFF
--- a/docs/examples/Classifying_Using_HMM.py
+++ b/docs/examples/Classifying_Using_HMM.py
@@ -1,0 +1,232 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+Classification Using Hidden Markov Model
+========================================
+This is a demonstration using the implemented Hidden Markov model to classify multiple targets.
+
+We will attempt to classify 3 targets in an undefined region.
+Our sensor will be all-seeing, and provide us with indirect observations of the targets such that,
+using the implemented hidden Markov Model (HMM), we should hopefully successfully classify exactly
+3 targets correctly.
+"""
+
+# %%
+# All Stone Soup imports will be given in order of usage.
+
+from datetime import datetime, timedelta
+import numpy as np
+
+# %%
+# Ground Truth
+# ^^^^^^^^^^^^
+# The targets may take one of three discrete hidden classes: 'bike', 'car' and 'bus'.
+# It will be assumed that the targets cannot transition from one class to another, hence an
+# identity transition matrix is given to the :class:`~.CategoricalTransitionModel`.
+#
+# A :class:`~.CategoricalState` class is used to store information on the classification/category
+# of the targets. The state vector will define a categorical distribution over the 3 possible
+# classes, whereby each component defines the probability that a target is of the corresponding
+# class. For example, the state vector (0.2, 0.3, 0.5), with category names ('bike', 'car', 'bus')
+# indicates that a target has a 20% probability of being class 'bike', a 30% probability of being
+# class 'car' etc.
+# It does not make sense to have a true target being a distribution over the possible classes, and
+# therefore the true categorical states will have binary state vectors indicating a specific class
+# (i.e. a '1' at one state vector index, and '0's elsewhere).
+# The :class:`~.CategoricalGroundTruthState` class inherits directly from the base
+# :class:`~.CategoricalState` class.
+#
+# While the category will remain the same, a :class:`~.CategoricalTransitionModel` is used here
+# for the sake of demonstration.
+#
+# The category and timings for one of the ground truth paths will be printed.
+
+from stonesoup.models.transition.categorical import CategoricalTransitionModel
+from stonesoup.types.groundtruth import CategoricalGroundTruthState
+from stonesoup.types.groundtruth import GroundTruthPath
+
+category_transition = CategoricalTransitionModel(transition_matrix=np.eye(3),
+                                                 transition_covariance=0.1 * np.eye(3))
+
+start = datetime.now()
+
+hidden_classes = ['bike', 'car', 'bus']
+
+# Generating ground truth
+ground_truths = list()
+for i in range(1, 4):
+    state_vector = np.zeros(3)  # create a vector with 3 zeroes
+    state_vector[np.random.choice(3, 1, p=[1/3, 1/3, 1/3])] = 1  # pick a random class out of the 3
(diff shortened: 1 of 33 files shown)


stonesoup/types/state.py
```diff
@@ -338,4 +338,50 @@ def covar(self):
...
+
 State.register(ParticleState)  # noqa: E305
+
+
+class CategoricalState(State):
+    r"""CategoricalState type.
+
+    State object representing an object in a categorical state space. A state vector
+    :math:`\mathbf{x}_i = P(\phi_i)` defines a categorical distribution over a finite set of
+    discrete categories :math:`\Phi = \{\phi_m|m\in\Z_{\ge0}\}`."""
```
nperree-dstl, line 351:
```suggestion
    discrete categories :math:`\Phi = \{\phi_m|m\in Z_{\ge0}\}`."""
```

docs/examples/Classifying_Using_HMM.py
```diff
@@ -0,0 +1,232 @@
...
+for states in np.vstack(ground_truths).T:
+    print(f"{new_state.timestamp:%H:%M:%S}", end="")
+    for state in states:
+        print(f" -- {new_state.category}", end="")
+    print()
+
+# %%
+# Measurement
+# ^^^^^^^^^^^
+# Using a hidden markov model, it is assumed the hidden class of a target cannot be directly
```
svidal-dstl, line 89:
```suggestion
# Using a Hidden Markov Model, it is assumed the hidden class of a target cannot be directly
```

stonesoup/predictor/categorical.py
```diff
@@ -0,0 +1,84 @@
...
+from ..base import Property
+from ..models.transition.categorical import CategoricalTransitionModel
+from ..predictor import Predictor
+from ..predictor._utils import predict_lru_cache
+from ..types.prediction import Prediction
+from ..types.state import CategoricalState
+
+
+class HMMPredictor(Predictor):
+    r"""Models the prediction step of a hidden Markov model"""
```
svidal-dstl, line 11:
```suggestion
    r"""Models the prediction step of a Hidden Markov Model"""
```

svidal-dstl (approved):

svidal-dstl (approved):
Misclicked re-request review, LGTM

docs/examples/Classifying_Using_HMM.py
```diff
@@ -0,0 +1,232 @@
...
+
+excess_tracks = len(tracks) - len(ground_truths)  # target value = 0
+sorted_tracks = sorted(tracks,
+                       key=lambda track: track.state_vector[np.argmax(track.state_vector)][0],
+                       reverse=True)
+best_tracks = sorted_tracks[:3]
+true_classifications = {ground_truth.category for ground_truth in ground_truths}
+track_classifications = {track.category for track in best_tracks}
+
+num_correct_classifications = len(true_classifications & track_classifications)
```
orosoman-dstl, line 229: If a={1,2}, b={2,1}, len(a & b)=2, even though the elements do not match up in terms of their order.  I can see that you've reordered the tracks so they shouldn't necessarily be order-wise equal, but is there a way to rework this so we check that each object is correctly categorised, as opposed to just the total number?

docs/examples/Classifying_Using_HMM.py
```diff
@@ -0,0 +1,232 @@
...
+
+# %%
+# Metric
+# ^^^^^^
+# Determining tracking accuracy.
+# In calculating how many targets were classified correctly, only tracks with the highest
+# classification certainty are considered.
+
+excess_tracks = len(tracks) - len(ground_truths)  # target value = 0
+sorted_tracks = sorted(tracks,
```
orosoman-dstl, line 222: hmm_output
This was my output from printing some values from here.  The issue I have is that we are being told there is 99.9% classification certainty for the top two tracks, but then 1 of them is wrong.  Obviously there is a possibility that this is random chance, but it happens every few times I run the code (without a seed).  Additionally, it seems strange that in this and other outputs, one track's probabilities are unchanged from the prior.

docs/examples/Classifying_Using_HMM.py
```diff
@@ -0,0 +1,232 @@
...
+#
+# The category and timings for one of the ground truth paths will be printed.
+
+from stonesoup.models.transition.categorical import CategoricalTransitionModel
+from stonesoup.types.groundtruth import CategoricalGroundTruthState
+from stonesoup.types.groundtruth import GroundTruthPath
+
+category_transition = CategoricalTransitionModel(transition_matrix=np.eye(3),
+                                                 transition_covariance=0.1 * np.eye(3))
+
```
orosoman-dstl, line 51: Would a seed be in order?  May have been a conscious omission but just in case you forgot

docs/examples/Classifying_Using_HMM.py
orosoman-dstl: One additional problem I noticed: my understanding is that the probabilities inside `track.state_vector` are in the order bike, car, bus, due to how `hidden_classes` is defined on line 54.  According to this, then, shouldn't `track_classifications` be picking bus twice?  Note this doesn't happen every rerun.  I've tried to work out why this is happening but I'm not sure unfortunately.

docs/examples/Classifying_Using_HMM.py
oharrald-Dstl: There's no determining what the correct ordering of the truths & tracks is. Since there are no kinematics to do a standard data association approach, we could do something along the lines of seeing which categorical distributions line up the best with each other, but in this instance I'm simply saying that we have the correct classification of target *somewhere*.

docs/examples/Classifying_Using_HMM.py
oharrald-Dstl: Good spot. There should be three elements in each set but I'd mistakenly used a set which won't add repeated elements. I'll replace with list.

docs/examples/Classifying_Using_HMM.py
oharrald-Dstl: I'd rather keep the example randomized if possible.

sdhiscocks (approved):

PR #506, opened 2021-10-01
Add nonlinear constant turn models. Update transition model class hierarchy

DaveKirkland:
Added nonlinear constant turn models. Moved nonlinear class up into the model i.e. all models are inherently nonlinear. Linear models are a subclass. Added tests for the StateVectors class.

Replaced CombinedLinearGaussianTransitionModel and CombinedNonlinearGaussianTransitionModel with CombinedGaussianTransitionModel which is now in the base class. This will break a lot of existing code that used these combined models but it is an easy fix e.g. replace CombinedLinearGaussianTransitionModel with CombinedGaussianTransitionModel.

### DIFF
--- a/stonesoup/models/transition/base.py
+++ b/stonesoup/models/transition/base.py
@@ -1,11 +1,14 @@
 # -*- coding: utf-8 -*-
 from abc import abstractmethod
+import copy
 from typing import Sequence
 
 from scipy.linalg import block_diag
+import numpy as np
 
 from ..base import Model, GaussianModel
 from ...base import Property
+from ...types.state import StateVector
 
 
 class TransitionModel(Model):
@@ -22,8 +25,79 @@ class TransitionModel(Model):
         pass
 
 
-class _CombinedGaussianTransitionModel(TransitionModel, GaussianModel):
+class CombinedGaussianTransitionModel(TransitionModel, GaussianModel):
+    r"""Combine multiple models into a single model by stacking them.
+
+    The assumption is that all models are Gaussian.
+    Time Variant, and Time Invariant models can be combined together.
+    If any of the models are time variant the keyword argument "time_interval"
+    must be supplied to all methods
+    """
     model_list: Sequence[GaussianModel] = Property(doc="List of Transition Models.")
+
+    def function(self, state, noise=False, **kwargs) -> StateVector:
+        """Applies each transition model in :py:attr:`~model_list` in turn to the state's
+        corresponding state vector components.
+        For example, in a 3D state space, with :py:attr:`~model_list` = [modelA(ndim_state=2),
+        modelB(ndim_state=1)], this would apply modelA to the state vector's 1st and 2nd elements,
+        then modelB to the remaining 3rd element.
+
+        Parameters
+        ----------
+        state : :class:`stonesoup.state.State`
+            The state to be transitioned according to the models in :py:attr:`~model_list`.
+
+        Returns
+        -------
+        state_vector: :class:`stonesoup.types.array.StateVector`
+            of shape (:py:attr:`~ndim_state, 1`). The resultant state vector of the transition.
+        """
+        temp_state = copy.copy(state)
+        ndim_count = 0
+        state_vector = np.zeros(state.state_vector.shape).view(StateVector)
+        # To handle explicit noise vector(s) passed in we set the noise for the individual models
+        # to False and add the noise later. When noise is Boolean, we just pass in that value.
+        if noise is None:
+            noise = False
+        if isinstance(noise, bool):
+            noise_loop = noise
+        else:
+            noise_loop = False
+        for model in self.model_list:
+            temp_state.state_vector =\
+                state.state_vector[ndim_count:model.ndim_state + ndim_count, :]
+            state_vector[ndim_count:model.ndim_state + ndim_count, :] += \
+                model.function(temp_state, noise=noise_loop, **kwargs)
+            ndim_count += model.ndim_state
+        if isinstance(noise, bool):
+            noise = 0
+        return state_vector + noise
+
+    def jacobian(self, state, **kwargs):
+        """Model jacobian matrix :math:`H_{jac}`
+
+        Parameters
+        ----------
+        state : :class:`~.State`
+            An input state
+
+        Returns
+        -------
+        :class:`numpy.ndarray` of shape (:py:attr:`~ndim_meas`, \
+        :py:attr:`~ndim_state`)
+            The model jacobian matrix evaluated around the given state vector.
+        """
+        temp_state = copy.copy(state)
+        ndim_count = 0
+        J_list = []
+        for model in self.model_list:
+            temp_state.state_vector =\
+                state.state_vector[ndim_count:model.ndim_state + ndim_count, :]
+            J_list.append(model.jacobian(temp_state, **kwargs))
+
+            ndim_count += model.ndim_state
+        out = block_diag(*J_list)
+        return out
 
     @property
     def ndim_state(self):
--- a/stonesoup/models/transition/linear.py
+++ b/stonesoup/models/transition/linear.py
@@ -7,7 +7,7 @@ import numpy as np
 from scipy.integrate import quad
 from scipy.linalg import block_diag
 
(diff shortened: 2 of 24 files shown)


stonesoup/models/transition/linear.py
```diff
@@ -30,27 +30,27 @@ def ndim_state(self):
...
+#         : :class:`numpy.ndarray` of shape\
+#         (:py:attr:`~ndim_state`, :py:attr:`~ndim_state`)
+#         """
 
-        transition_matrices = [
-            model.matrix(**kwargs) for model in self.model_list]
-        return block_diag(*transition_matrices)
+#         transition_matrices = [
+#             model.matrix(**kwargs) for model in self.model_list]
+#         return block_diag(*transition_matrices)
```
oharrald-Dstl, line 53: If the entire class is commented-out, it should probably be deleted. Though I think having the 'linear' combined model is still useful. Applying the combined transition via a single matrix multiplication will be quicker than the model-by-model logic of the combined parent class.

DaveKirkland: Yes, it looks like I forgot to remove that.
To handle time varying models, the total matrix has to be rebuilt at each call. Also because the total matrix is block diagonal, the full matrix multiply involves a lot of multiply by zero operations. It may make sense to do the smaller matrix multiplies and stitch together the result. It would probably be useful to benchmark the results to see if there is a significant performance hit.

DaveKirkland: @oharrald-Dstl @sdhiscocks  I've done some benchmarking and indeed the CombinedLinearGaussianTransitionModel is faster than the general implementation. I will add it back in, but utilize the new class hierarchy.

stonesoup/models/transition/nonlinear.py
```diff
@@ -1,46 +1,215 @@
...
+        turn_rate[turn_rate == 0.] = np.finfo(float).eps
+        dAngle = turn_rate * time_interval_sec
+        sv2 = StateVectors(
+            [sv1[0] + np.sin(dAngle)/turn_rate * sv1[1] - sv1[3] / turn_rate *
+             (1. - np.cos(dAngle)),
+             sv1[1] * np.cos(dAngle) - sv1[3] * np.sin(dAngle),
+             sv1[1] / turn_rate * (1. - np.cos(dAngle)) + sv1[2] + sv1[3] * np.sin(dAngle)
+             / turn_rate,
+             sv1[1] * np.sin(dAngle) + sv1[3] * np.cos(dAngle),
+             turn_rate])
```
sdhiscocks, line 112: Should this be using `sv1[0, :]`, etc. Also, I think by pre-calculating cos/sin, you can also save some space on the lines.
```suggestion
        cos_dAngle = np.cos(dAngle)
        sin_dAngle = np.sin(dAngle)
        sv2 = StateVectors([
            sv1[0, :] + sin_dAngle/turn_rate*sv1[1, :] - sv1[3, :]/turn_rate*(1. - cos_dAngle),
            sv1[1, :]*cos_dAngle - sv1[3, :]*sin_dAngle,
            sv1[1, :]/turn_rate*(1. - cos_dAngle) + sv1[2, :] + sv1[3, :]*sin_dAngle/turn_rate,
            sv1[1, :]*sin_dAngle + sv1[3, :]*cos_dAngle,
            turn_rate
        ])
```

stonesoup/models/transition/nonlinear.py
```diff
@@ -1,46 +1,215 @@
...
+
+        Returns
+        -------
+        : :class:`int`
+            The number of combined model state dimensions.
+        """
+        return sum(model.ndim_state for model in self.model_list) + 5
+
+    def function(self, state, noise=False, **kwargs) -> StateVector:
+        state_tmp = copy.deepcopy(state)
```
sdhiscocks, line 170: Could probably do a regular `copy` here, as long as you replace `state_vector` 3 lines down.

sdhiscocks (changes requested):
Looks good 👍 

Main issue is the `function` method needs to support vectorised input, which only requires minor change, and original `jacobian` method effectively tests this.

stonesoup/models/transition/base.py
```diff
@@ -22,9 +25,78 @@ def ndim_state(self) -> int:
...
+        J_list = []
+        for model in self.model_list:
+            temp_state.state_vector =\
+                state.state_vector[ndim_count:model.ndim_state + ndim_count, :]
+            J_list.append(model.jacobian(temp_state, **kwargs))
+
+            ndim_count += model.ndim_state
+        out = block_diag(*J_list)
+        return out
+
```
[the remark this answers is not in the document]
DaveKirkland: For large models this may be faster - I'll benchmark and see. I've updated the test to use approx(). Once I've done the benchmarking I'll commit the changes.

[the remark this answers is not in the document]
DaveKirkland: So I've done some benchmarking. Use a combination of 3  Constant Velocity models, the default jacobian calculation  (for the entire model combined) is 3 times slower that the above code, so I think it is better to leave this in.

stonesoup/models/transition/nonlinear.py
DaveKirkland: @sdhiscocks In working on the ConstantTurn model I found an issue with the covariance matrix that causes a problem with the Model.pdf(). I've changed the covariance matrix used in the CT model, so that it uses the discretized form of continuous time motion - this makes it consistent with the current linear form constant turn models. This avoids the issue I mentioned.

sdhiscocks (approved):

docs/source/auto_demos/Video_Processing.rst
```diff
@@ -1,28 +1,17 @@
-
```
oharrald-Dstl, line 1: Do you want to introduce edits to this file?

docs/source/auto_demos/Video_Processing.ipynb
```diff
@@ -15,7 +15,7 @@
       "cell_type": "markdown",
```
oharrald-Dstl, line 15: Do you want to introduce edits to this file?

oharrald-Dstl:
As there are now two `ConstantTurn` models, the files:
docs/source/auto_examples/Smooth_Platform_Coordinate_Transitions
stonesoup/simulator/transition.py
will require a specific model be referenced.

sdhiscocks:
> As there are now two `ConstantTurn` models, the files: docs/source/auto_examples/Smooth_Platform_Coordinate_Transitions stonesoup/simulator/transition.py will require a specific model be referenced.

Good point. I wonder whether it would better to change name of the models? What about `ConstantTurnRate` maybe for the linear one; or maybe call one `CoordinatedTurn` (or would that cause confusion)?

DaveKirkland:
> > As there are now two `ConstantTurn` models, the files: docs/source/auto_examples/Smooth_Platform_Coordinate_Transitions stonesoup/simulator/transition.py will require a specific model be referenced.
> 
> Good point. I wonder whether it would better to change name of the models? What about `ConstantTurnRate` maybe for the linear one; or maybe call one `CoordinatedTurn` (or would that cause confusion)?

@oharrald-Dstl @sdhiscocks 
I was thinking of 'KnownTurnRate' and also 'KnownTurnRateSandwich' for the linear models and leaving the nonlinear models as 'ConstantTurn' and 'ConstantTurnSandwich' - thoughts?

DaveKirkland:
@sdhiscocks I refreshed the commit with the new class names and merged with the main branch. I'm not sure why it has changes for the Video_Processing.ipynb and associated rst and gif files - I didn't changed them.

I also tried rebasing with HEAD~2, but this picks up more than 20 previous commits and I wasn't quite sure how to handle that.

sdhiscocks:
> @sdhiscocks I refreshed the commit with the new class names and merged with the main branch. I'm not sure why it has changes for the Video_Processing.ipynb and associated rst and gif files - I didn't changed them.
> 
> I also tried rebasing with HEAD~2, but this picks up more than 20 previous commits and I wasn't quite sure how to handle that.

I reverted the changes to the video processing demo, and then squash it back into your original commit, rebased onto *main*. Also added a extra commit, just to fix documentation failure.

oharrald-Dstl (approved):

PR #522, opened 2021-10-19
Create Orbital class for OrbitalState to inherit from

nperree-dstl:
Orbital methods have moved into an "Orbital" class which OrbitalState and GaussianOrbitalState now inherit from.

### DIFF
--- a/stonesoup/types/orbitalstate.py
+++ b/stonesoup/types/orbitalstate.py
@@ -8,6 +8,7 @@ import numpy as np
 
 from ..base import Property
 from ..functions.orbital import keplerian_to_rv, tru_anom_from_mean_anom
+from . import Type
 from .array import StateVector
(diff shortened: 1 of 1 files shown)


stonesoup/types/orbitalstate.py
```diff
@@ -177,95 +173,89 @@ def __init__(self, state_vector, *args, **kwargs):
...
 
                 kwargs['timestamp'] = timestamp
-                super().__init__(state_vector, *args, **kwargs)
 
             else:
                 raise TypeError("State vector and metadata cannot both be empty")
 
         # Otherwise check that the state vector is the right size
-        elif len(state_vector) != 6:
+        elif len(self.state_vector) != 6:
```
idorrington-dstl, line 189: Should this be hard-coded as 6? Can it not be made more general?

PR #535, opened 2021-11-15
Add Clutter Model that can be used with radar Sensors

ekhunter123:
A useful feature of Stone Soup is the creation of platforms (moving or stationary) which can hold various sensors on them. These sensors are able to measure a given groundtruth using a measurement model. The proposed clutter model introduces clutter (false alarms/detections) to these measurements, making the simulations more realistic. The clutter can be distributed according to any provided distribution function.

This pull request follows up on this discussion: #485 and implements the third idea listed.

An example usage of the ClutterModel in 2D can be found [here](https://gist.github.com/ekhunter123/a19154ea7da33a46a7eaca86b84055ab).
An example usage of the ClutterModel in 3D can be found [here](https://gist.github.com/ekhunter123/671fc33bd1803bc4b041979db8053ca2).

As I have written it, the clutter model can be added to the RadarBearingRange and the RadarElevationBearingRange sensors. Other sensors could be edited to include an optional clutter model, I just don't have enough experience in sensors to do this.

I would love any comments or suggestions on this model. As discussed in some of the user conferences, I hope this model will be a good place to start and build off of!


apiszcz:
Thank you, for the notebook sample, I am interested in testing this capability and need more time on this topic along with sensor creation.

stonesoup/models/measurement/clutter.py
```diff
@@ -0,0 +1,272 @@
...
+from ...types.angle import Bearing, Elevation
+from ...types.groundtruth import GroundTruthState
+from ...types.array import CovarianceMatrix
+from ...functions import cart2pol, cart2sphere
+from ...types.array import StateVector, StateVectors
+from ...types.numeric import Probability
+from ...types.state import State
+
+
+class ClutterBearingRange(MeasurementModel):
```
sdhiscocks, line 19: I wonder if we should make these unique models, rather than `MeasurementModel` subclass?

stonesoup/models/measurement/clutter.py
```diff
@@ -0,0 +1,272 @@
...
+            else:
+                random_vector = self.generate_normal_vector()
+            
+            # Subtract the mounting sensor's position from the clutter,
+            # then convert to polar coordinates.
+            if position is not None:
+                clutter_vector = random_vector - position[0:2].flatten()
+
+            rho, phi = cart2pol(*clutter_vector)
+            clutter_vector = np.array([Bearing(phi), rho])
```
sdhiscocks, line 104: Could the measurement model not be used to do this? I guess possibly not as state space may contain more than position?

stonesoup/models/measurement/clutter.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Returns
+        -------
+        : :class:`numpy.array`
+            A uniform random vector with length 3.
+        """
+        x = uniform.rvs(self.state_space[0][0], self.state_space[0][1])
+        y = uniform.rvs(self.state_space[1][0], self.state_space[1][1])
+        z = uniform.rvs(self.state_space[2][0], self.state_space[2][1])
+        return np.array([x, y, z])
```
sdhiscocks, line 260: Could probably make parent method more generic, by sampling `n` times from uniform distribution based on size of size of `state_space`.

```python
return np.array([uniform.rvs(*space) for space in self.state_space])
```

stonesoup/models/measurement/clutter.py
```diff
@@ -0,0 +1,272 @@
...
+    """
+
+    clutter_rate: float = Property(
+        default=1.0,
+        doc="The average number of clutter points per time step. The actual "
+            "number is Poisson distributed")
+    distribution: np.string_ = Property(
+        default='uniform',
+        doc="The type of distribution that the clutter follows. Supported "
+            "distributions are: uniform, normal (Gaussian)")
```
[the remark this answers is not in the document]
ekhunter123: Good idea! I added a definition for the `Distribution(Enum)` class in clutter.py. Is that what you were thinking? I don't have much experience with Enums.

stonesoup/models/measurement/clutter.py
```diff
@@ -0,0 +1,272 @@
...
+    state_space: list = Property(
```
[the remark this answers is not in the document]
ekhunter123: Yes that would be better. I changed it!

stonesoup/models/measurement/clutter.py
```diff
@@ -0,0 +1,272 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from scipy.stats import uniform
```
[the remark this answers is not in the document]
ekhunter123: That would be neater. They are both `numpy` now.

stonesoup/models/measurement/clutter.py
ekhunter123: Hmm I'm not sure. Which part specifically were you thinking could be done by the measurement model? Practically the state space should have the same dimension as the position. But I don't know if that is enforced.

stonesoup/models/measurement/clutter.py
ekhunter123: Yes, great idea. Thanks

stonesoup/models/measurement/clutter.py
sdhiscocks: So the measurement model already has a functions to go from state space to measurement space, as well as logic to handle position offsets (and rotation, that I just realised is missing here).

I'm just wondering if there is a way to use the measurement model that would then remove the need to carry out coordinate conversion step, and such could make it more flexible on which sensors you could use the clutter model (by sharing the measurement model with the sensor).


Maybe something like this? (not tested):
```suggestion
            state = State([0.0] * self.measurement_model.ndim_state, timestamp=timestamp)
            state.state_vector[self.measurement_model.mapping, 0] += random_vector
            
            clutter_vector = self.measurement_model.function(state)
```

ekhunter123: Oh yes, I like that. And you're right, it now includes rotation which I hadn't thought of before. This change also means that the function() function is now the same for the `ClutterBearingRange` (2D) as for the `ClutterElevationBearingRange` (3D). I implemented this change.

ekhunter123: This now means that the only difference between `ClutterBearingRange` and `ClutterElevationBearingRange` is the attribute definitions. Do you think I should make the documentation more generic and then have just one class, maybe simply called `ClutterModel`?

stonesoup/sensor/radar/radar.py
```diff
@@ -62,6 +68,12 @@ def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray,
                                       timestamp=truth.timestamp,
                                       groundtruth_path=truth)
             detections.add(detection)
+        
+        # Generate clutter at this time step
+        if self.clutter_model is not None: 
+            self.clutter_model.measurement_model = measurement_model
```
oharrald-Dstl, line 74: Could do this in constructor (`__init__`) (and on line 231)

stonesoup/models/measurement/clutter.py
```diff
@@ -0,0 +1,191 @@
...
+        # necessary to get the proper timestamp. If there is no
+        # groundtruth return a set of no Clutter.
+        if not ground_truths:
+            return set()
+
+        # Generate the clutter for this time step
+        timestamp = next(iter(ground_truths)).timestamp
+        clutter = set()
+        for _ in range(np.random.poisson(self.clutter_rate)):
+            if self.distribution == Distribution.uniform:
```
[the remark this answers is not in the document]
ekhunter123: Thanks for the comment. What do you mean by "the introduction of the `Distribution` type"? 
I had been using [this ](https://www.geeksforgeeks.org/enum-in-python/) website which indicated that both `is` and `==` are supported comparisons by Enum. I can change it to `is` if that would be better!

stonesoup/models/measurement/clutter.py
```diff
@@ -0,0 +1,191 @@
...
+    def generate_uniform_vector(self):
+        """Generate a random vector in the uniform distribution according
+        to the model's defined state space.
+
+        Returns
+        -------
+        : :class:`numpy.array`
+            A uniform random vector with length equal to self.ndim_meas.
+        """
+        arr = np.array([np.random.default_rng().uniform(*space)
```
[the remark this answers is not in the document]
ekhunter123: I was thinking that the way I implemented it made it easier on the user as it would save them the trouble of researching random generators and deciding whether to use numpy or scipy etc. But I do see the benefits of your suggestion. It would definitely make it more versatile to different distributions and remove the need for logic and checks. If someone else comments in the next day or so with a reason to keep it the way it is then I will keep it. Otherwise I'll change it!

stonesoup/sensor/radar/radar.py
ekhunter123: Good idea. Unfortunately, the sensor's measurement model isn't defined until its `measure()` function gets called. I think this was done so that the measurement model would hold the correct position and rotation offset at each time step. However, this means that the `ClutterModel.measurement_model` attribute can't be set in it's own constructor or in the radar sensor's constructor. We might actually want the measurement model to change at each time step for the same reasons as the radar's one changes. 

The only way I see around this would be to move the definition of the measurement model to a constructor for the radar. However this might mess up the measurement model for the reasons described above.

oharrald-Dstl: Very good point. Its good as is.

stonesoup/models/measurement/clutter.py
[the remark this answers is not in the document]
ekhunter123: Hi. I started to make this change yesterday but I ran into a problem. `np.random.default_rng().multivariate_normal(some_mean, some_covariance)` is a numpy array. It is not a function, meaning that if it is passed as a distribution then there is actually nothing to call to get more random vectors. This would happen with all such distribution functions I believe. If there is some other way to pass in a function with the parameters already set please let me know! :)

One way around this would be to make a function like the one below and pass that in
```
def distribution(mean, covariance):  
    return np.random.default_rng().multivariate_normal(mean, covariance)
```

Or I could have two parameters in the `ClutterModel` called `distribution` and `dist_params` . Then the user could set 
```
distribution = np.random.default_rng().multivariate_normal
dist_params = [mean, covariance]
```
and then the `ClutterModel` would say 
`clutter_vector = self.distribution(*self.dist_params)`

Is there a preferred way to do this? Thanks for your help :)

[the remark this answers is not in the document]
oharrald-Dstl: I suppose it gets difficult where you are calling the rng several times for the uniform distribution, with varying args, but once for the normal distribution.

What you've suggested sounds good. i.e. you have properties:
    * `distribution` where user could pass `np.random.default_rng().multivariate_normal` or `np.random.default_rng().uniform` etc.
    * `dist_params` where user could pass `(mean, covariance)` or `[(mean, cov), (mean2, cov2)]` or `space` or `[space, space2, space3]`

You could then do:
random_vector = np.array([dist_method(*arg) for arg in dist_args])
(as you would still need to make sure you get an array back if you are working with multiple spaces with uniform dist)
Though this is starting to look confusing.

I'm maybe being too picky here. The current implementation you have is fine. Just wondered if there was 1. a way, 2. need for, genericising the distribution.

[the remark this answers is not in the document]
ekhunter123: Ok thanks, I'll incorporate this into the code. The more I think about it the more I agree that a generic version would be good. It means we don't have to do all the type checking, and leaves it up to the user. I'll add a default uniform distribution to make it easier for people who don't want to do something fancy or want to use it quickly.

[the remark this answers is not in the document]
ekhunter123: Update: this idea has been implemented. I also updated the example gists so that they run with the newly defined model.

stonesoup/models/measurement/clutter.py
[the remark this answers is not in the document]
ekhunter123: Resolving this as it is outdated and the `Distribution` enum is no longer used.

stonesoup/models/measurement/clutter.py
ekhunter123: Due to the other changes made, there is now just one model that incorporates 2D and 3D. It is called `ClutterModel`.

stonesoup/models/measurement/clutter.py
ekhunter123: I'm definitely happy to do that, it wouldn't be hard since they only inherit the base `MeasurementModel`, so all I would need to do is add the `ndim_state` and `mapping` attributes. Would you like me to make that change? Thanks :)

stonesoup/sensor/radar/radar.py
```diff
@@ -135,7 +135,7 @@ def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray,
         for truth in ground_truths:
             # Transform state to measurement space and generate
             # random noise
-            measurement_vector = measurement_model.function(truth, noise=False, **kwargs)
+            measurement_vector = measurement_model.function(truth, noise=noise, **kwargs)
```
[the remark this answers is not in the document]
ekhunter123: Sorry, my bad! I don't know how that change got in there. I corrected it.

stonesoup/models/measurement/clutter.py
ekhunter123: Following our conversation on Tuesday, I made the change. The class actually doesn't need the `ndim_state` or `mapping` attributes as it will get all relevant information from the `MeasurementModel` once it is used.

oharrald-Dstl:
Need to add the clutter models to documentation.
In `docs/source/stonesoup.models.measurement.rst` you'll need to add an entry for the clutter model file.
E.g.
`
Clutter
-------
.. automodule:: stonesoup.models.measurement.clutter
    :show-inheritance:
    :inherited-members:
`

stonesoup/sensor/radar/radar.py
```diff
@@ -62,6 +68,12 @@ def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray,
                                       timestamp=truth.timestamp,
                                       groundtruth_path=truth)
             detections.add(detection)
+        
+        # Generate clutter at this time step
+        if self.clutter_model is not None:
```
oharrald-Dstl, line 73: Whitespace character at the end of this line needs to be removed.

stonesoup/sensor/radar/radar.py
```diff
@@ -213,6 +225,12 @@ def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray,
                                       timestamp=truth.timestamp,
                                       groundtruth_path=truth)
             detections.add(detection)
+
```
oharrald-Dstl, line 228: Whitespace characters in this blank line need to be removed.

stonesoup/sensor/radar/radar.py
```diff
@@ -213,6 +225,12 @@ def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray,
                                       timestamp=truth.timestamp,
                                       groundtruth_path=truth)
             detections.add(detection)
+        
+        # Generate clutter at this time step
+        if self.clutter_model is not None:
```
oharrald-Dstl, line 230: Trailing whitespace character here needs to be removed.

oharrald-Dstl (changes requested):
Minor comments regarding FLAKE8 errors. To stick to PEP8 (coding style) standards, these should be addressed.
Separate comments made about adding the clutter models to documentation and writing some tests for the new code.
Once done, this will be a great addition.

stonesoup/sensor/radar/radar.py
```diff
@@ -20,6 +20,7 @@
 from ...types.groundtruth import GroundTruthState
 from ...types.numeric import Probability
 from ...types.state import State, StateVector
+from ...models.clutter import ClutterModel
```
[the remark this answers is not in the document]
ekhunter123: Thanks for reminding me, I just moved clutter.py to the models folder, not the measurement folder. After the meeting last week, I changed it so that it is not a subclass of the measurement model.

ekhunter123:
> Need to add the clutter models to documentation. In `docs/source/stonesoup.models.measurement.rst` you'll need to add an entry for the clutter model file. E.g. Clutter ------- .. automodule:: stonesoup.models.measurement.clutter :show-inheritance: :inherited-members:

@oharrald-Dstl I have created a new file stonesoup.models.clutter.rst which you can find [here](https://github.com/ekhunter123/Stone-Soup/blob/main/docs/source/stonesoup.models.clutter.rst). I decided to put it in its own file, rather than in stonesoup.models.measurement.rst because it is not a subclass of the measurement model anymore. Please let me know if there is a problem with this or the formatting of the file, I've never made a .rst file before. Thanks!

ekhunter123:
> Some tests for these new models would be good.

@oharrald-Dstl Thanks for the suggestion. I have never made tests before, but I tried to copy the format of the tests for the measurement models. I also put the clutter.py and its tests in a new folder. You can find it all [here](https://github.com/ekhunter123/Stone-Soup/tree/main/stonesoup/models/clutter). I apologize in advance for all mistakes! I wasn't sure what to put in the tests, but I test the `ndim` attribute and the `function()` function. Let me know if there is anything else I should test!

oharrald-Dstl (changes requested):
Mainly comments to get tests working.
Once sorted, this looks like it should be good to go.

stonesoup/models/clutter/tests/test_models.py
```diff
@@ -0,0 +1,65 @@
...
+                            dist_params=dist_params)
+
+    meas_model = CartesianToElevationBearingRange(
+            ndim_state=6,
+            mapping=(0, 2, 4),
+            noise_covar=np.eye(3))
+    model_test.measurement_model = meas_model
+
+    truth = State(StateVector([1, 1, 1]), timestamp=datetime.now)
+    clutter = model_test.function(truth)
```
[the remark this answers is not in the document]
ekhunter123: That's a good point. I can see it either way. The measurement models take a single state and return a single state. The clutter model will still return a set of states. The prototype started as a ClutterSensor and so I used the same parameters as the [`Sensor` parent class. ](https://github.com/ekhunter123/Stone-Soup/blob/main/stonesoup/models/clutter/clutter.py). I never got around to changing it since then. As the ground truth is only used for a timestamp it could even be simplified to just take in a timestamp, rather than passing a single (or set of) state(s).

What do you think would fit best with the other models? It would be nice if they all had roughly the same parameters/design.

stonesoup/models/clutter/tests/test_models.py
```diff
@@ -0,0 +1,65 @@
...
+)
+@pytest.mark.parametrize(
+    "ModelClass, clutter_rate, dist,\
+     dist_params",
+    [
+        (   # 2D meas
+            ClutterModel,
+            5,
+            np.random.default_rng().uniform,
+            ((-50, 50), (-50, 50)),
```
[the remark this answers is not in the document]
ekhunter123: Thanks for all your help, I've fixed many of these test problems so I will resolve the conversations for now.

oharrald-Dstl:
One last issue: error raised that the file `docs/source/stonesoup.models.clutter.rst` isn't in a toctree.
Just need to add `stonesoup.models.transition` to the list of folders in `docs/source/stonesoup.models.rst` so that clutter models appear in our table of contents.
Eg.

Models
======

.. toctree::

    stonesoup.models.clutter
    stonesoup.models.control
    stonesoup.models.measurement
    stonesoup.models.transition

.. automodule:: stonesoup.models
    :no-members:

.. automodule:: stonesoup.models.base
    :show-inheritance:

oharrald-Dstl (approved):

jswright-dstl (approved):

sdhiscocks (approved):

PR #543, opened 2021-11-29
Add method to copy prior state properties of a platform across to a new state

oharrald-Dstl:
The motivation behind this change is to allow the use of GroundTruthState types in platforms.
Previously, a platform's (movement controller's) `move` method hard-coded a standard `State` append.
This lead to loss of metadata when initiating a platform with a GroundTruthState, since the metadata property was not copied across to the new state.

### DIFF
--- a/stonesoup/movable/movable.py
+++ b/stonesoup/movable/movable.py
@@ -1,20 +1,18 @@
 # -*- coding: utf-8 -*-
-import copy
 import datetime
 from abc import abstractmethod, ABC
 from functools import lru_cache
-
-from math import cos, sin
-from typing import Sequence, Tuple
+from typing import Sequence, Tuple, Any
 
 import numpy as np
+from math import cos, sin
 from scipy.linalg import expm
 
-from stonesoup.functions import cart2sphere, cart2pol, build_rotation_matrix, rotz
-from stonesoup.types.array import StateVector
 from stonesoup.base import Property
-from stonesoup.types.state import State, StateMutableSequence
+from stonesoup.functions import cart2sphere, cart2pol, build_rotation_matrix, rotz
 from stonesoup.models.transition import TransitionModel
+from stonesoup.types.array import StateVector
+from stonesoup.types.state import State, StateMutableSequence
 
 
 class Movable(StateMutableSequence, ABC):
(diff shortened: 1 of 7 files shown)


stonesoup/movable/movable.py
```diff
@@ -49,6 +47,45 @@ def __init__(self, *args, **kwargs):
         if not self.states:
             raise ValueError('States must not be empty: it must contain least one state.')
 
+    @staticmethod
+    def from_state(state: State, *args: Any, **kwargs: Any) -> State:
```
etfrogers-dstl, line 51: This method doesn't seem to use any behavior of the Moveable class (and you say its behaviour) is similar to `Prediction` and `Update` classes. 

This method feels like it belongs in a parent class, probably `State`. I'll think about this and see if I can refactor.

etfrogers-dstl (changes requested):

etfrogers-dstl (commented):
I have updated this PR, and now it could do with review by someone else

etfrogers-dstl (approved):

jswright-dstl (approved):

stonesoup/types/state.py
```diff
@@ -33,6 +34,136 @@ def ndim(self):
...
+        `GaussianStatePrediction` is defined) such that
+        `class_mapping[Prediction][GaussianState] = GaussianStatePrediction`. If this method is
+        then called like below
+
+        >>>> gaussian_state = GaussianState(some_arguments)
+        >>>> new_prediction = Prediction.from_state(gaussian_state, *args, **kwargs)
+
+        then the `from_state` method will look up the class mapping and see that
+        `Prediction.from_state()` called with a `GaussianState` input should return a
+        `GaussianStatePrediction` object, and therefore,the type of `new_prediction` will be
```
jswright-dstl, line 135: "therefore,the type" -> "therefore, the type"

etfrogers-dstl: Thanks. Should be addressed

stonesoup/movable/movable.py
```diff
@@ -1,24 +1,22 @@
...
 import datetime
 from abc import abstractmethod, ABC
 from functools import lru_cache
-
-from math import cos, sin
-from typing import Sequence, Tuple
+from typing import Sequence, Tuple, MutableSequence, Optional
 
 import numpy as np
+from math import cos, sin
```
sdhiscocks, line 8: This is standard library.

sdhiscocks (approved):

Issue #495, opened 2021-09-08
Particle Filter - Don't resample at every time step


sdhiscocks:
Currently, Stone Soup only has one resampler, which resamples at every time step.  A resampler should (possibly as a wrapper for existing resamplers) asses weight degeneracy (e.g. with effective sample size), and then return original particles if resampling isn't required.

PR #563, opened 2021-12-13
Introduce resampler that checks if resampling required

orosoman-dstl:
This change is in response to #495 
The new ESSResampler class checks whether the ESS (Effective Sample Size) is below a given threshold before proceeding, otherwise simply returning the original particles without any resampling.

### DIFF
--- a/stonesoup/resampler/particle.py
+++ b/stonesoup/resampler/particle.py
@@ -49,17 +49,45 @@ class SystematicResampler(Resampler):
         return new_particles
 
 
-class ESSResampler(SystematicResampler):
+class ESSResampler(Resampler):
+    """ESSResampler class
+        This wrapper uses a :class:`~.Resampler` to resample the particles inside
+        an instant of :class:`~.Particles`, but only after checking if this is necessary
+        by comparing Effective Sample Size (ESS) with a supplied threshold (numeric).
+        Kish's ESS is used, as in Section 3.5 of [1]_.   Doucet-Johansen, 2008.
+
+        References
+        ----------
+        .. [1] Doucet A., Johansen A.M., 2009, Tutorial on Particle Filtering \
+        and Smoothing: Fifteen years later, Handbook of Nonlinear Filtering, Vol. 12.
+
+        """
 
     threshold: float = Property(default=None,
-                                doc='Threshold compared with ESS to decide whether to resample')
+                                doc='Threshold compared with ESS to decide whether to resample. \
+                                    Default is number of particles divided by 2, \
+                                        set in resample method')
+    resampler: Resampler = Property(default=SystematicResampler,
+                                    doc='Resampler to wrap, which is called \
+                                        when ESS below threshold')
 
     def resample(self, particles):
+        """
+        Parameters
+        ----------
+        particles : list of :class:`~.Particle`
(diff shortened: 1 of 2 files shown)


jswright-dstl (approved):

stonesoup/resampler/particle.py
```diff
@@ -42,7 +43,23 @@ def resample(self, particles):
...
         new_particles = Particles(state_vector=particles.state_vector[:, index],
-                                  weight=[weight]*n_particles,
+                                  weight=[weight] * n_particles,
                                   parent=Particles(state_vector=particles.state_vector[:, index],
                                                    weight=particles.weight[index]))
         return new_particles
+
+
+class ESSResampler(SystematicResampler):
+
```
sdhiscocks, line 53: This could do with a docstring, to explain a little about what it does, and any references you've used.

stonesoup/resampler/particle.py
```diff
@@ -42,7 +43,23 @@ def resample(self, particles):
...
         index = weight_order[np.searchsorted(cdf, np.log(u_j))]
         new_particles = Particles(state_vector=particles.state_vector[:, index],
-                                  weight=[weight]*n_particles,
+                                  weight=[weight] * n_particles,
                                   parent=Particles(state_vector=particles.state_vector[:, index],
                                                    weight=particles.weight[index]))
         return new_particles
+
+
+class ESSResampler(SystematicResampler):
```
sdhiscocks, line 52: I wonder if this class could be a wrapper that could be used with other resamplers.
```python
class ESSResampler(Resampler):
    resampler: Resampler = Property(doc="Resampler to wrap, which is called when ESS below threshold")
    ...
```

stonesoup/resampler/particle.py
```diff
@@ -42,7 +43,23 @@ def resample(self, particles):
...
+
+    threshold: float = Property(default=None,
+                                doc='Threshold compared with ESS to decide whether to resample')
```
sdhiscocks, line 55: Would be good to add to doc that default of `None` means threshold is number of particles / 2.

stonesoup/resampler/particle.py
```diff
@@ -42,7 +43,45 @@ def resample(self, particles):
...
+
+        Returns
+        -------
+        particles : list of :class:`~.Particle`
+            The particles, either unchanged or resampled, depending on weight degeneracy
+        """
+        if not isinstance(particles, Particles):
+            particles = Particles(particle_list=particles)
+        if self.threshold is None:
+            self.threshold = len(particles) / 2
```
oharrald-Dstl, line 83: Something to add to the class' `__init__` method, so that this is not needed at every call to `resample`.

orosoman-dstl: The problem is, len(particles) is unknown when we create the ESSResampler object.  Typically, the object is created with something like: 
resampler = ESSResampler()
And then later on we have something like:
new_particles = resampler.resample(old_particles)
So the only way I could see to fix this is to maybe make the number of particles a Property in the base Resampler class.

oharrald-Dstl: Ah got it. That makes sense.

oharrald-Dstl (approved):

PR #564, opened 2021-12-14
WIP: Gaussian Processes

sglvladi:
This PR stems from work performed on attempting to fit Gaussian Processes into the existing framework. 

More specifically, a `SimpleGaussianProcess` transition model has been added, along with an example use-case. 

Some notable points for discussion:
- Currently, the model expects a `Track` instead of a `State` object as an input to the `function()` method, while the same is also necessary when evaluating the `matrix()` and `covar()` methods, even though the model is inherently a `LinearGaussianTransitionModel`. 
  - The main reason for the above is that the last `num_lags` timestamps are necessary for computing the Kernel matrix.
  - Even though this is currently compatible with most hypothesisers, since the actual `Track` instance is forwarded to the underlying predictors (see [here](https://github.com/dstl/Stone-Soup/blob/main/stonesoup/hypothesiser/distance.py#L73)), predictors like the `KalmanPredictor` would not forward this to the transition model. 
- The model expects a `start_time` argument on initialisation, which is used to compute the respective timestamps (in terms of number of seconds since `start_time`). 
  - This is not ideal since different tracks may require different values for `start_time`.
  - Note that `start_time` is not (necessarily) the same as the timestamp of the first track state. 

**NOTE: This PR is Work In Progress**
The main purpose of the PR at its current stage is to motivate thoughts and discussions on how this can be taken forward.

### DIFF
--- a/examples/gp_example.py
+++ b/examples/gp_example.py
@@ -1,0 +1,58 @@
+"""
+Stone Soup Gaussian Process Transition Model Example
+====================================================
+
+
+"""
+
+from datetime import datetime, timedelta
+
+from stonesoup.types.array import StateVector
+from stonesoup.types.state import State
+from stonesoup.types.track import Track
+from stonesoup.models.transition.gaussian_process import SimpleGaussianProcess
+
+init_time = datetime.now()
+init_state = State(StateVector([1, 0, 0]), timestamp=init_time+timedelta(seconds=1))
+track = Track([init_state])
+
+gp = SimpleGaussianProcess(num_lags=3, sigma=1, start_time=init_time)
+
+# Time: 1.2 sec
+# =====================================================
+time_interval = timedelta(seconds=0.2)
+new_timestamp = track.state.timestamp+time_interval
+# Compute transition and covariance matrices (just for display)
+F = gp.matrix(track, time_interval=time_interval)
+Q = gp.covar(track, time_interval=time_interval)
+# Run the track through the model to compute new state vector
+sv = gp.function(track, time_interval=time_interval)
+# Append state to track
+track.append(State(sv, timestamp=new_timestamp))
+
+# Time: 2 sec (1.2 + 0.8)
+# =====================================================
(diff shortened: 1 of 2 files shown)


sdhiscocks:
> Some notable points for discussion:
> 
> * Currently, the model expects a `Track` instead of a `State` object as an input to the `function()` method, while the same is also necessary when evaluating the `matrix()` and `covar()` methods, even though the model is inherently a `LinearGaussianTransitionModel`.
>   
>   * The main reason for the above is that the last `num_lags` timestamps are necessary for computing the Kernel matrix.
>   * Even though this is currently compatible with most hypothesisers, since the actual `Track` instance is forwarded to the underlying predictors (see [here](https://github.com/dstl/Stone-Soup/blob/main/stonesoup/hypothesiser/distance.py#L73)), predictors like the `KalmanPredictor` would not forward this to the transition model.

This is also compounded by the cache which pulls the state from the track for Kalman based predictors: https://github.com/dstl/Stone-Soup/blob/fbb48342253056d88b477160439fdadfcbc48c2b/stonesoup/predictor/_utils.py#L23-L25
The reason for passing a Track to the Predictor, was for the case where a predictor may use history of the track as part of the process, but the Kalman predictors at least don't cover the case that the model may use the history.

> * The model expects a `start_time` argument on initialisation, which is used to compute the respective timestamps (in terms of number of seconds since `start_time`).
>   
>   * This is not ideal since different tracks may require different values for `start_time`.
>   * Note that `start_time` is not (necessarily) the same as the timestamp of the first track state.
> 

I'm not sure on the best approach to solve this. One less than ideal way would be to create a predictor that wrapped existing predictors, which could operate on the track, extract out the timestamps and work out the required `start_time` for the track (not clear on how `start_time` is defined), then either replace or modify the model linked to the wrapped predictor.

Maybe another solution is to pass the track as a keyword argument? This could be done in `predict_lru_cache`?

PR #568, opened 2021-12-20
Implement multi-frame assignment algorithm

joldf:
This pull request adds support for multi-frame assignment (MFA).

* This implementation is heavily based on the previous pull request for MFA #344. However, while that was written in MATLAB using the MATLAB wrapper to call it from Stone Soup, this version is written in pure Python and does not require a MATLAB installation.
* As with the original MATLAB implementation, this is based on the article: Xia, Y., Granström, K., Svensson, L., García-Fernández, Á.F., and Williams, J.L., 2019. Multiscan Implementation of the Trajectory Poisson Multi-Bernoulli Mixture Filter. *J. Adv. Information Fusion*, 14(2), pp. 213–235.
* The actual logic of the algorithm is contained within two files, `mfa_init.py` and `mfa_step.py`, which are intentionally independent of any other Stone Soup code so that they could be used as-is in any other code base where they might be useful.
* The results are track oriented (or, in the language of the above paper, trajectory orientated). The hypotheses for each track are stored its tag field (as in the MATLAB implementation). Each time step, the algorithm computes the best overall hypothesis (i.e. the best combination of track hypotheses) based on the whole sliding window but does not directly expose this result. However it would be easy to add if a suitable way of expressing the result could be found. Ultimately, they get exposed indirectly by the n-scan pruning (which prunes track hypotheses that don't match the best hypotheses before the sliding window).
* There are two optimisation steps that call out to external libraries. The first is the Hungarian algorithm in the dual subproblem; this uses [`linear_sum_assignment` in `scipy`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linear_sum_assignment.html), which is an existing dependency of Stone Soup. The other is solving the linear programming primal problem, which is done using [Google OR Tools](https://developers.google.com/optimization); this is a mature library with a liberal licence (Apache v2.0), and can also be called from C++ which would aid if this MFA implementation were to be translated to C++. In this pull request, OR Tools is added as a required dependency for Stone Soup, but that could be omitted in which case you would get a reasonably self-explanatory error if you tried to import the MFADataAssociator.

### DIFF
--- a/docs/examples/MFA_example.py
+++ b/docs/examples/MFA_example.py
@@ -1,42 +1,22 @@
 #!/usr/bin/env python
 
 """
-Multi Frame Assignment example
+Multi-Frame Assignment example
 ==============================
-This notebook includes an example of the Multi Frame Assignment (MFA) algorithm [#]_.
+This notebook includes an example of the Multi-Frame Assignment (MFA) algorithm [#]_.
+
+The multi-frame assignment algorithm maintains and prunes hypotheses over *N*-Scans. This enables
+the global optimum assignment choice to be selected over *N* multiple frames/timesteps/scans.
 """
-import matplotlib
-# %%
-# Initial imports
-# ---------------
-
-import numpy as np
-from scipy.stats import chi2
-import datetime
-from matplotlib import animation
-from matplotlib.patches import Ellipse
-
-from stonesoup.dataassociator.mfa import MFADataAssociator
-from stonesoup.hypothesiser.mfa import MFAHypothesiser
-from stonesoup.models.transition.linear import CombinedLinearGaussianTransitionModel, \
-                                               ConstantVelocity
-from stonesoup.types.groundtruth import GroundTruthPath, GroundTruthState
-from stonesoup.types.detection import TrueDetection
-from stonesoup.types.detection import Clutter
-from stonesoup.models.measurement.linear import LinearGaussian
-from stonesoup.types.state import TaggedWeightedGaussianState
-from stonesoup.types.track import Track
-from stonesoup.types.update import GaussianMixtureUpdate
-from stonesoup.types.mixture import GaussianMixture
-from stonesoup.types.numeric import Probability
-from stonesoup.predictor.kalman import KalmanPredictor
-from stonesoup.updater.kalman import KalmanUpdater
-from stonesoup.hypothesiser.probability import PDAHypothesiser
 
 # %%
 # Global parameters
 # -----------------
+import numpy as np
+from scipy.stats import chi2
+
 np.random.seed(2001)  # Used to set the seed for the numpy random number generator
+
 prob_detect = 0.9  # Prob. of detection
 gate_level = 8  # Gate level
 prob_gate = chi2.cdf(gate_level, 2)  # Prob. of gating, computed from gate level for hypothesiser
@@ -48,8 +28,8 @@ slide_window = 3  # Slide window; used by MFA data associator
 # %%
 # Plotting functions
 # ------------------
-# These custom plotting functions are used to show the different assignments
-# present withing the sliding window.
+# First we'll need some custom plotting functions, used to show the different assignments
+# present within the slide window.
 
 def plot_covar(state, ax, color=None):
     w, v = np.linalg.eig(
@@ -83,24 +63,32 @@ def plot_tracks(tracks, ax, slide_window=None):
             parents.append(component)
             mini_tracks.append(parents)
 
-        drawn_states = []
-        for t in mini_tracks:
+        drawn_states = set()
+        for mini_track in mini_tracks:
             # Avoid re-plotting drawn trajectory
-            t_tmp = [x for x in t if x not in drawn_states]
-            if len(t_tmp) < len(t):
-                t_tmp.insert(0, t[-(len(t_tmp)+1)])
-            artists.extend(ax.plot([state.state_vector[0, 0] for state in t_tmp],
-                                   [state.state_vector[2, 0] for state in t_tmp], plot_style))
-            for state in t_tmp:
-                # Avoid re-plotting drawn error ellipses
-                if state not in drawn_states:
-                    artists.append(plot_covar(state, ax, plot_style[0]))
-                    drawn_states.append(state)
+            states_to_plot = [state for state in mini_track if state not in drawn_states]
+            # Insert state before so line is drawn
+            if len(states_to_plot) < len(mini_track):
+                states_to_plot.insert(0, mini_track[-(len(states_to_plot)+1)])
+            artists.extend(ax.plot([state.state_vector[0, 0] for state in states_to_plot],
+                                   [state.state_vector[2, 0] for state in states_to_plot],
+                                   plot_style))
+            # Avoid re-plotting drawn error ellipses
+            for state in set(states_to_plot) - drawn_states:
+                artists.append(plot_covar(state, ax, plot_style[0]))
+                drawn_states.add(state)
+
     return artists
 
 # %%
 # Models
 # ------
+# Create models for target motion :math:`[x \ \dot{x} \ y \ \dot{y}]` with near constant velocity
+# and measurements :math:`[x \ y]`.
+from stonesoup.models.transition.linear import (
+    CombinedLinearGaussianTransitionModel, ConstantVelocity)
+from stonesoup.models.measurement.linear import LinearGaussian
+
 transition_model = CombinedLinearGaussianTransitionModel([ConstantVelocity(0.005),
                                                           ConstantVelocity(0.005)])
 measurement_model = LinearGaussian(ndim_state=4, mapping=(0, 2),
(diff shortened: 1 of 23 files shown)


stonesoup/hypothesiser/mfa.py
```diff
@@ -0,0 +1,60 @@
...
+                        weight=new_weight,
+                        state_vector=hypothesis.prediction.state_vector,
+                        covar=hypothesis.prediction.covar,
+                        timestamp=hypothesis.prediction.timestamp
+                    )
+                hypotheses.append(hypothesis)
+        # Create Multiple Hypothesis and add to list
+        hypotheses = MultipleHypothesis(hypotheses)
+
+        return hypotheses
```
orosoman-dstl, line 60:
```suggestion
        return hypotheses

```
Needs newline to pass flake8 requirements

stonesoup/hypothesiser/mfa.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+
+        # Check to make sure all detections are obtained from the same time
+        timestamps = set([detection.timestamp for detection in detections])
+        if len(timestamps) > 1:
+            raise ValueError("All detections must have the same timestamp")
+
+        hypotheses = list()
+        detections_list = list(detections)
+        for component in track.state.components:
```
orosoman-dstl, line 41: This line is breaking things when I try to run an example.  I can't find any mention of `state.components` anywhere, and when I try to run something like `print(state.components)` for a valid State or Gaussian instance I get an `AttributeError`.  Also, there are compatibility issues as to which underlying hypothesiser is used: if we initialise using `DistanceHypothesiser`, for example, this class' `hypothesise` method takes the whole `Track` as a parameter.  I'm not sure which underlying hypothesiser is most appropriate, as I'm still getting to grips with how the algorithm functions, but some kind of way of dealing with someone attempting to use the wrong one is probably needed, or some guidance in the documentation.

orosoman-dstl (commented):
Some tests are definitely needed - will have a more detailed look on Monday.

stonesoup/hypothesiser/mfa.py
orosoman-dstl: I think you maybe meant `for component in track.states:`

orosoman-dstl:
I tried running the example script linked in PR #344, and that failed with the following error:
mfa_error
I think this is to do with the timestamp being nested inside an object inside the Gaussian Mixture.  Are you able to provide a working example?

stonesoup/hypothesiser/mfa.py
joldf: This seems to assume that `track.state` is a `GaussianMixture`, which does have a `components` attribute, and I think later code assumes it consists of `TaggedWeightedGaussianState` objects. I could perhaps put a couple of `assert isinstance(...)` statements here to assure that, but that would go against Python's duck typing philosophy (if the user came up with their own classes with the same interface they would be prevented from passing that in). Maybe a better solution would be to wrap relevant statements in `try: ... except AttributeError:` and re-throw the exception with a more useful error message – would that suit?

stonesoup/dataassociator/mfa.py
```diff
@@ -0,0 +1,60 @@
...
+from .mfa_init import init_hyp_info, Hyp
+from .mfa_step import MAX_ITERATION_COUNT, AlgorithmState, algorithm_step, prune_hypotheses
+
+
+class MFADataAssociator(DataAssociator):
+    """Data associator using multi-frame assignment algorithm over a sliding window."""
+
+    hypothesiser: Hypothesiser = Property(
+        doc='Generate a set of hypotheses for each prediction-detection pair')
+    slide_window: int = Property(doc='Length of MFA slide window')
```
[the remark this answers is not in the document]
joldf: Agreed, I'll put a default of 3 in as that's what the orginal MATLAB implementation of MFA seemed to use as a default.

stonesoup/dataassociator/mfa.py
```diff
@@ -0,0 +1,60 @@
...
+from .base import DataAssociator
+from ..base import Property
+from ..hypothesiser import Hypothesiser
+from ..types.multihypothesis import MultipleHypothesis
+from .mfa_init import init_hyp_info, Hyp
+from .mfa_step import MAX_ITERATION_COUNT, AlgorithmState, algorithm_step, prune_hypotheses
+
+
+class MFADataAssociator(DataAssociator):
+    """Data associator using multi-frame assignment algorithm over a sliding window."""
```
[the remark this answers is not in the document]
joldf: Capitalising the words in a spelt out initialism is not standard English (unless the term is a proper noun anyway e.g. UN) according to all style guides I've looked at on the subject (e.g. Chicago, The Economist). But it does seem to be a curiously common habit in the defence industry and in any case I'm still happy to make the change if you wish.

stonesoup/hypothesiser/mfa.py
oharrald-Dstl: You could use `hasattr(obj, "components")`, but I think just adding a note to documentation is all that's needed

stonesoup/dataassociator/mfa.py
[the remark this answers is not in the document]
orosoman-dstl: Ok fair enough.  I just thought it might add clarity so someone can quickly see what "MFA" actually stands for, but it's not a big deal so I'll leave it up to you.

joldf:
Sorry about the trouble to get this working, especially the lack of a link in the pull request text to the example script you found referenced from #344 – I had intended to include a link to it, but there was a lot of detail I wanted to mention and I ended up missing that part by accident.

I have just tried the script on a fresh clone of this branch (in case I somehow pushed a different version than I tested locally) and it did work for me. I ran these commands:

    git clone https://github.com/joldf/Stone-Soup stone_soup
    cd stone_soup
    git checkout
    git checkout multi_frame_assignment_python
    curl -O https://gist.githubusercontent.com/sglvladi/2aeebd831f1264541266f777d00c241b/raw/f228a9b023246510935c14297ebf9809035279fd/MFA_example.py
    python MFA_example.py

I also tried merging latest changes from master in case things were broken by more recent commits, and that also worked fine:

    git checkout main
    git pull https://github.com/dstl/Stone-Soup
    git checkout multi_frame_assignment_python
    git merge main
    # Fix merge conflicts in graphical tool and commit
    python MFA_example.py

I set a breakpoint at line 126 of `kalman.py`, where you got that exception `AttributeError: 'GaussianMixture' object has no attribute 'timestamp'`. But, for me, the `prior` object has type `TaggedWeightedGaussianState`, which does have a `prior.timestamp` attribute. Additionally, the `timestamp` parameter is not `None` anyway, so that expression is not even evaluated. Do you have any ideas on what could be different about your test setup?

orosoman-dstl:
> 

Ah yes it looks like it does indeed work when using the github version of the branch.  I made some very minor changes on my local version which must have caused the error above, so sorry I didn't revert these before checking properly (I wrongly thought it wouldn't make any difference).  On a slightly related note though, the best way we can be confident this code works (now and in the future when new code is added) is with some unit tests.  You could probably make heavy use of the example script to make that easier.

stonesoup/dataassociator/mfa.py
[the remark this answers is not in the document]
joldf: Good point. I'll put the abbreviation in the comment too: "Data associator using multi-frame assignment (MFA) algorithm over a sliding window"

stonesoup/__init__.py
```diff
@@ -13,5 +13,6 @@
 © Crown Copyright 2017-2021 Defence Science and Technology Laboratory UK
 © Crown Copyright 2018-2021 Defence Research and Development Canada / Recherche et développement pour la défense Canada
 © Copyright 2018-2021 University of Liverpool UK
+© Copyright 2021 Roke Manor Research Ltd UK
```
orosoman-dstl, line 16: You may wish to change this to 2022 or 2021-2022

stonesoup/types/mixture.py
```diff
@@ -56,6 +61,40 @@ def append(self, component):
...
+        means = self.means
+        weights = self.weights / Probability.sum(self.weights)
+        return np.average(means, axis=1, weights=weights)
+
+    @property
+    def covar(self):
+        _, covar = gm_reduce_single(self.means, self.covars,
+                                    self.weights)
+        return covar
+
```
orosoman-dstl, line 97: I think it might be good to add a timestamp property, so that we can easily access the timestamp of the components.  This is quite important to avoid errors when calculating metrics, for example.  Something like:
`@property
def timestamp(self):
    return next((component.timestamp for component in self.components), None)`
Maybe a check to confirm all the component timestamps are equal would be sensible to add somewhere.

sdhiscocks:
@joldf Could you look at adding some unit tests for the code please? Also, could we include tutorial/example, maybe derived from #344?

sdhiscocks:
I've rebased this onto `main`, done a minor restructure, added some flake8 fixes, a bug fix for track ordering, and added an animated example based on the one referenced in #344.

TODO:
- [ ] Unit tests
- [ ] Add additional information about MFA to the example.

sdhiscocks:
Added a few more changes to add some consistency with changes that had occurred on `main` branch.

Once merged, I'll raise an issue in regards of how measurement history is stored, as use of indices in tags, I don't think is the ideal approach.

docs/examples/MFA_example.py
```diff
@@ -0,0 +1,241 @@
...
+                                                      timestamp=start_time,
+                                                      weight=Probability(1), tag=[])])
+tracks = OrderedSet((Track([prior1]), Track([prior2])))
+
+# %%
+# And finally run through the scans, and plot the track and assignments
+# within slide window at each frame/timestep.
+from matplotlib import animation
+from matplotlib.patches import Ellipse
+
```
nperree-dstl, line 200:
```suggestion
import matplotlib

matplotlib.rcParams['animation.html'] = 'jshtml'

```

nperree-dstl (approved):
Could possibly do with more detailed documentation in the example but otherwise fine.

Issue #469, opened 2021-07-27
Metrics should be returned in a dictionary


oharrald-Dstl:
Currently, the method of attaining a particular metric from the standard metric manager class is by doing something along the lines of:
`metric_i_want = {metric for metric in metrics if metric.title == "something"}.pop()`

Replacing the returned set with a dictionary, keyed by sensible metric names would be better.

sdhiscocks:
Closed with 30addea

PR #570, opened 2022-01-05
Return metrics in a dictionary

orosoman-dstl:
This branch makes metrics return as a dictionary, with the keys being equal to `metric.title`.   
In response to #469.

### DIFF
--- a/docs/examples/ParticleFlow.py
+++ b/docs/examples/ParticleFlow.py
@@ -201,8 +201,8 @@ for predictor, updater, colour, filter, particle_count \
         [siap_gen], associator=TrackToTruth(association_threshold=np.inf))
     metric_manager.add_data(tracks={track}, groundtruth_paths={truth})
 
-    pa[filter] = {metric for metric in metric_manager.generate_metrics()
-                  if metric.title.startswith("SIAP Position Accuracy at times")}.pop()
+    metrics = metric_manager.generate_metrics()
+    pa[filter] = metrics.get("SIAP Position Accuracy at times")
 
 # %%
 # Positional Accuracy
--- a/stonesoup/metricgenerator/manager.py
+++ b/stonesoup/metricgenerator/manager.py
@@ -85,8 +85,7 @@ class SimpleManager(MetricManager):
             if not isinstance(metric_list, list):
                 metric_list = [metric_list]
             for metric in metric_list:
-                key = metric.title.replace(' at times', '')  # Shorten key a little
-                metrics[key] = metric
+                metrics[metric.title] = metric
         return metrics
 
     def list_timestamps(self):
--- a/stonesoup/metricgenerator/manager.py
+++ b/stonesoup/metricgenerator/manager.py
@@ -78,17 +78,16 @@ class SimpleManager(MetricManager):
         if self.associator is not None and self.association_set is None:
             self.associate_tracks()
 
-        metrics = set()
+        metrics = {}
         for generator in self.generators:
-            metric = generator.compute_metric(self)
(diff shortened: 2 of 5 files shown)


stonesoup/metricgenerator/manager.py
```diff
@@ -78,17 +78,16 @@ def generate_metrics(self):
...
-            else:
-                metrics.add(metric)
-
-        return set(metrics)
+            metric_list = generator.compute_metric(self)
+            # If not already a list, force it to be one below
+            if not isinstance(metric_list, list):
+                metric_list = [metric_list]
+            for metric in metric_list:
+                key = metric.title.replace(' at times', '')  # Shorten key a little
```
oharrald-Dstl, line 88: We'd end-up with a repeated key for SIAPs here. Eg. "SIAP Completeness" (which is an average completeness value over all timestamps) and "SIAP Completeness at times" (which is a list of completeness values, with a value for every timestamp).

stonesoup/metricgenerator/tests/test_basicmetrics.py
```diff
@@ -51,11 +51,10 @@ def test_basicmetrics():
                                            end_timestamp=start_time +
                                            datetime.timedelta(seconds=4)),
                                        generator=generator)}
-
     for metric_name in ["Number of targets",
                         "Number of tracks", "Track-to-target ratio"]:
         calc_metric = [i for i in correct_metrics if i.title == metric_name][0]
-        meas_metric = [i for i in metrics if i.title == metric_name][0]
+        meas_metric = [metrics[i] for i in metrics if i == metric_name][0]
```
oharrald-Dstl, line 57:
```suggestion
        meas_metric = metrics.get(metric_name)
```

Then check if meas_metric is `None` etc.

stonesoup/metricgenerator/manager.py
orosoman-dstl: Ah yes I realised this late yesterday - I'll keep the ' at times'.

docs/examples/ParticleFlow.py
```diff
@@ -201,8 +201,9 @@
         [siap_gen], associator=TrackToTruth(association_threshold=np.inf))
     metric_manager.add_data(tracks={track}, groundtruth_paths={truth})
 
-    pa[filter] = {metric for metric in metric_manager.generate_metrics()
-                  if metric.title.startswith("SIAP Position Accuracy at times")}.pop()
+    metrics = metric_manager.generate_metrics()
+    pa[filter] = {metrics.get(metric) for metric in metrics
+                  if metric.startswith("SIAP Position Accuracy at times")}.pop()
```
oharrald-Dstl, line 206: Since you can work with keys now, this would be simpler:
```suggestion
    pa[filter] = metrics.get("SIAP Position Accuracy at times")
```

stonesoup/metricgenerator/tests/test_basicmetrics.py
```diff
@@ -54,7 +54,7 @@ def test_basicmetrics():
     for metric_name in ["Number of targets",
                         "Number of tracks", "Track-to-target ratio"]:
         calc_metric = [i for i in correct_metrics if i.title == metric_name][0]
-        meas_metric = [metrics[i] for i in metrics if i == metric_name][0]
+        meas_metric = [metrics.get(i) for i in metrics if i == metric_name][0]
```
oharrald-Dstl, line 57:
```suggestion
        meas_metric = metrics.get(metric_name)
```

oharrald-Dstl (approved):

sdhiscocks (approved):

PR #576, opened 2022-01-11
Add composite tracking modules

oharrald-Dstl:
Introduces suite of composite state and tracking module types in order to infer the state(s) of an object that exists in several state spaces at once.


oharrald-Dstl:
Closed in favour of composite_tracking_modules branch PR, which contains categorical module improvements.

PR #583, opened 2022-01-17
Fix a minor mistake with generating clutter in tutorial 11

orosoman-dstl:
Just a miniscule change to fix an error I found in tutorial 11.  The clutter was being generated over too small an area compared with the `meas_range`, which this should fix.


nperree-dstl (approved):

jswright-dstl (approved):

ekhunter123 (approved):
Good catch, thanks for making this change!

PR #587, opened 2022-01-18
Add a RandomState parameter to the clutter model

### LINKED ISSUE #575: Radar cluttertest doesn't consider no clutter
On occasion, the radar clutter test raises an error due to no clutter being generated leading to len(detections) == 1.
See:
https://github.com/dstl/Stone-Soup/blob/31b5ff296dee2eff5732dd6fe6804c8d12384010/stonesoup/sensor/radar/tests/test_radar.py#L675

ekhunter123:
Add an optional numpy.random.RandomState parameter to the clutter model. This helps with reproducibility of code and closes #575.


jswright-dstl (approved):

stonesoup/models/clutter/clutter.py
```diff
@@ -42,6 +42,9 @@ class ClutterModel(Model, ABC):
         "The default defines the space for a uniform distribution in 2D. The call "
         "`np.array([self.distribution(*arg) for arg in self.dist_params])` "
         "must return a numpy array of length equal to the number of dimensions.")
+    random_state: Optional[np.random.RandomState] = Property(
+        default=np.random.RandomState(1),
```
sdhiscocks, line 46: The issue with having this as a default, is run to run you'll get the same results, and I think it'd be better that by default the global random state is used. Also, this default will be shared by all instances, which may have unexpected issues.

Something like in Gaussian Model may be better:
https://github.com/dstl/Stone-Soup/blob/2de2fc0a7075c9898dffb841658076cfbe781a7e/stonesoup/models/base.py#L211-L215

ekhunter123: That's a very good point, I hadn't thought of that. I've made this change like you suggested. However, I did keep the seed as either an integer seed or a numpy random state (or default None). I thought this way the user can use the same random state for the poisson random variable and the distribution attribute. For example:
```
rs = numpy.random.RandomState(seed=1)  
clutter_model = ClutterModel(  
     distribution=rs.uniform,  
     seed=rs,  
     ...  
)
```

Does that seem ok?

ekhunter123: Sorry, I don't quite understand why this change made the tests fail. It seems to be failing Test 39, installing dependencies.

sdhiscocks: > Sorry, I don't quite understand why this change made the tests fail. It seems to be failing Test 39, installing dependencies.

CircleCI seems to be having issues lately with the dependency cache getting corrupted. Seems I can't reset the cache on PRs from forks. Will have to look into a solution, but just ignore failures for now.

sdhiscocks (approved):

stonesoup/models/clutter/clutter.py
ekhunter123: ok, thank you!

PR #589, opened 2022-01-25
Make measurement model optional in measurement-based initiators

### LINKED ISSUE #588: Suggestion to make measurement model optional in SimpleMeasurementInitiator
The [SimpleMeasurementInitiator](https://stonesoup.readthedocs.io/en/v0.1b7/_modules/stonesoup/initiator/simple.html#SimpleMeasurementInitiator) is used to easily initiate a track based on a measurement. It has a mandatory `measurement_model` attribute, however in the following section of code you can see that it will try to use the measurement model attached to the detection in question.

https://github.com/dstl/Stone-Soup/blob/2de2fc0a7075c9898dffb841658076cfbe781a7e/stonesoup/initiator/simple.py#L90-L97

If the user knows that all detections will be assigned a measurement model, it then becomes redundant to make a "dummy" measurement model just to satisfy the required attribute. I propose something similar to the [KalmanUpdater](https://stonesoup.readthedocs.io/en/v0.1b5/_modules/stonesoup/updater/kalman.html#KalmanUpdater) which notes that the measurement model is not required if each detection has one. It enforces this requirement in the `_check_measurement_model()` function and raises an error if necessary. 

https://github.com/dstl/Stone-Soup/blob/2de2fc0a7075c9898dffb841658076cfbe781a7e/stonesoup/updater/base.py#L41-L45

I'm more than happy to implement this change if others think it is a good idea. I think it is a simple matter of adding a new else statement before line 97 (and giving a default None for the attribute).

ekhunter123:
A followup to close #588, we make the `measurement_model` attribute optional in initiators which are based on detections (namely the `SinglePointInitiator`, `SimpleMeasurementInitiator`, and `MultiMeasurementInitiator` classes). Detections are often paired with a measurement model, so the initiator will use that model in the new state. However, if the initiator is not given a measurement model and is passed a detection which also does not have a measurement model, it will raise a ValueError.

I have also made a test to ensure that the ValueError is raised correctly. I think there is no need to test the `MultiMeasurementInitiator` because the error will be raised in the `ExtendedKalmanUpdater` class, which is already tested. But I can add a test for this one as well if it would be best.

### DIFF
--- a/stonesoup/initiator/simple.py
+++ b/stonesoup/initiator/simple.py
@@ -166,15 +166,15 @@ class MultiMeasurementInitiator(GaussianInitiator):
     Does cause slight delay in initiation to tracker."""
 
     prior_state: GaussianState = Property(doc="Prior state information")
-    measurement_model: MeasurementModel = Property(
-        default=None,
-        doc="Measurement model. Can be left as None if all detections have a "
-            "valid measurement model.")
     deleter: Deleter = Property(doc="Deleter used to delete the track.")
     data_associator: DataAssociator = Property(
         doc="Association algorithm to pair predictions to detections.")
     updater: Updater = Property(
         doc="Updater used to update the track object to the new state.")
+    measurement_model: MeasurementModel = Property(
+        default=None,
+        doc="Measurement model. Can be left as None if all detections have a "
+            "valid measurement model.")
     min_points: int = Property(
         default=2, doc="Minimum number of track points required to confirm a track.")
     updates_only: bool = Property(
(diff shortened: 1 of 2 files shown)


ekhunter123:
Note, this might be a good time to add a docstring for the `skip_non_reversible` attribute in the SimpleMeasurementInitiator. It's not entirely clear what it is for.

ekhunter123:
In the test_multi_measurement() function in test_simple.py, it had previously created the measurement initiator like this
https://github.com/dstl/Stone-Soup/blob/a8760f9a34898a3e326456e65bc22bf915ac0a63/stonesoup/initiator/tests/test_simple.py#L279-L281
But since the measurement model is now optional, this messes with the order of parameters. So I had to change the test to do this
https://github.com/ekhunter123/Stone-Soup/blob/b6a52b1abcadea893bba973020226f12ae3f2e1d/stonesoup/initiator/tests/test_simple.py#L279-L282

If someone had previously made an initiator in this way, it won't work anymore. Does this mean that this is a breaking change?

oharrald-Dstl (approved):

sdhiscocks (approved):

PR #594, opened 2022-02-08
Add multi-distance hypothesiser

oharrald-Dstl:
The standard DistanceHypothesiser only allows a fixed measure to be used.  If one has incoming detections of BearingRange type, and some of Cartesian type, clearly the distance should be calculated differently.
The usual way around the issue would be to use a Mahalanobis measure and entirely avoid the units of measurement in the measurement space.
This PR adds a hypothesiser that allows the user to have multiple "distance" measures available, dependent on what the incoming detections' measurement models are.

I'll leave this PR as a discussion in the instance where this PR is redundant and there are better ways to go around this problem.

### DIFF
--- a/stonesoup/hypothesiser/distance.py
+++ b/stonesoup/hypothesiser/distance.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
-import warnings
-from typing import Sequence
+from typing import Mapping
 
 from .base import Hypothesiser
 from ..base import Property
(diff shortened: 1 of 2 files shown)


stonesoup/hypothesiser/distance.py
```diff
@@ -87,3 +91,65 @@ def hypothesise(self, track, detections, timestamp, **kwargs):
...
+        :attr:`hypothesisers`."""
+
+        # Keep track of detections attributed to each hypothesiser
+        hypothesisers_detections = {hypothesiser: set() for hypothesiser in self.hypothesisers}
+
+        # True detection hypotheses
+        for detection in detections:
+            try:
+                detection_measurement_model = detection.measurement_model
+                index_of_model = self.measurement_models.index(detection_measurement_model)
```
sdhiscocks, line 136: I'm not sure this will work where you have models which are more dynamic e.g. with changing translation or rotation offsets.
It would be better as an instance check? Maybe it could use MRO to find nearest match?

stonesoup/hypothesiser/distance.py
```diff
@@ -87,3 +91,65 @@ def hypothesise(self, track, detections, timestamp, **kwargs):
...
+    measurement models. When a detection is received of particular model, the class will attempt
+    to find the model's corresponding hypothesiser and use its output.
+    """
+
+    hypothesisers: Sequence[DistanceHypothesiser] = Property(
+        doc="Sequence of hypothesisers to utilise output of.")
+    measurement_models: Sequence[MeasurementModel] = Property(
+        doc="Sequence of measurement models that will be compared against incoming detections. "
+            "Must be same length as hypothesisers sequence."
+    )
```
sdhiscocks, line 113: Would these be better as `Mapping[MeasurementModel, Hypothesiser]`?

stonesoup/hypothesiser/distance.py
oharrald-Dstl: Good point. Perhaps limiting the user to only defining measurement model type, rather than instance would be better?

stonesoup/hypothesiser/distance.py
oharrald-Dstl: Yes, definitely.

oharrald-Dstl:
I've replaced the separate `hypothesisers` and `measurement_models` attributes in favour of a dictionary, keyed by the measurement model **types** and valued with the hypothesisers.
The reason behind not using particular measurement model instances is that a measurement model might be dynamic, and have its properties change over time. Therefore either an instance check or type check would be needed for correct keying.

The multi-hypothesiser now raises an error on detections without measurement models, or measurement models that have no corresponding hypothesiser in the `hypothesiser_dict` attribute.

PR #617, opened 2022-03-31
Restructure forward algorithm (in context of hidden Markov model)

oharrald-Dstl:
The most important change here is the use of the "emission matrix". Which has been correctly defined now as the matrix modelling conditional probabilities of receiving measurement classes given a particular hidden class.

Models now do not have additive noise (this was originally done to make the models interface similarly to other Stone Soup models, but was not necessary).

Some renaming also carried-out.

### DIFF
--- a/stonesoup/updater/categorical.py
+++ b/stonesoup/updater/categorical.py
@@ -1,96 +1,41 @@
 # -*- coding: utf-8 -*-
+
 import numpy as np
 
 from ..base import Property
-from ..models.measurement.categorical import CategoricalMeasurementModel
+from ..models.measurement.categorical import MarkovianMeasurementModel
 from ..types.prediction import MeasurementPrediction
-from ..types.state import CategoricalState
 from ..types.update import Update
 from ..updater import Updater
 
 
 class HMMUpdater(Updater):
-    r"""Models the update step of a hidden Markov model"""
+    r"""Hidden Markov model updater"""
 
-    measurement_model: CategoricalMeasurementModel = Property(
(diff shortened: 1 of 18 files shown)


stonesoup/updater/categorical.py
```diff
@@ -40,96 +96,35 @@ def _check_measurement_model(self, measurement_model):
...
-    def _get_emission_matrix(self, hypothesis):
-
-        measurement_model = self._check_measurement_model(hypothesis.measurement.measurement_model)
-
-        return measurement_model.emission_matrix
-
-    def predict_measurement(self, predicted_state, measurement_model=None, category_names=None,
-                            **kwargs):
-        r"""Predict the measurement implied by the predicted state mean
+    def predict_measurement(self, predicted_state, measurement_model, measurement, **kwargs):
```
sdhiscocks, line 106: Why is the `measurement` needed here? Are their attributes which should be part of the model?

oharrald-Dstl: It was passed here so that the measurement category names could be copied across to the predicted measurement.
Currently, measurement category names are added on by the HMMSensor, NOT by the MarkovianMeasurementModel.
Do you think the measurement model should contain these instead?

oharrald-Dstl: Have moved the `measurement_categories` to the measurement model. The HMMSensor now grabs this attribute and adds it to the yielded CategoricalDetections.

oharrald-Dstl:
Closed due to composite tracking module branch containing the same changes.

PR #621, opened 2022-04-11
Add sensor suite type

oharrald-Dstl:
The intention of this feature is to allow the user to append sensor information to its returned detections.

The SensorSuite type allows the user to define attributes that they would like to be detailed in a sensor's returned detections' metadata.
It can contain multiple sensors, and will yield all their returned detections in one set.

### DIFF
--- a/stonesoup/sensor/sensor.py
+++ b/stonesoup/sensor/sensor.py
@@ -1,11 +1,12 @@
 from abc import abstractmethod, ABC
-from typing import Set, Union
+from typing import Set, Union, Sequence
 
 import numpy as np
 
-from stonesoup.types.detection import TrueDetection
-from stonesoup.types.groundtruth import GroundTruthState
-from .base import PlatformMountable
+from ..base import Property
+from ..sensor.base import PlatformMountable
+from ..types.detection import TrueDetection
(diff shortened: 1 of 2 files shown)


jswright-dstl (approved):

stonesoup/sensor/sensor.py
```diff
@@ -38,3 +39,42 @@ def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray,
...
+
+            attributes_dict = {attribute_name: sensor.__getattribute__(attribute_name)
+                               for attribute_name in self.attributes_inform}
+
+            for detection in detections:
+                detection.metadata.update(attributes_dict)
+
+            all_detections.update(detections)
+
+        return all_detections
```
sdhiscocks, line 80: Not sure why you'd want to group all detections from all sensors together?

jswright-dstl: There are multiple examples of grouping measurements, i.e., AIS messages from multiple receivers. The aim of this is specifically aimed at adding information to the metadata of a measurement rather than the focus being on the multiple sensors.

sdhiscocks (approved):

Issue #199, opened 2020-04-18
Related Stone-Soup presentations?


apiszcz:
Is there a list of stonesoup related presentations?

https://www.fkie.fraunhofer.de/content/dam/fkie/de/documents/11thSymposiumSDF2017/PaulThomas_StoneSoup_An%20Open%20Source%20Framework%20for%20Tracking%20and%20State%20Estimation.pdf

https://www.gov.uk/government/news/dstl-shares-new-open-source-framework-initiative

https://isif-ostewg.org/stone-soup/

PR #625, opened 2022-04-22
Ensemble Kalman Filter Algorithm and Data Type

0sm1um:
Here is my implementation of the EnKF which I presented at last year's Fusion Conference.

A full writeup of the algorithm itself, as well as design philosophy can be found in my paper titled _"Implementation of Ensemble Kalman Filters in Stone Soup"_. The short version is that the EnKF is a numerical approximation of the Kalman Filter, which can also be used for nonlinear estimation.

The gist of this implementation is a new State Type is introduced known as `EnsembleState`. This state bears a resemblence to the GaussianState, except instead of being composed of a single `StateVector` and `CovarianceMatrix` it represents the state as a single `StateVectors` instance which is referred to as the `ensemble` in the code. This data type has class methods which allow the user to return the mean of the ensemble, and sample covariance matrix which allows users familliar with GaussianState to interact with it in all the ways they are used to.

This pull request includes this new state type, as well as a new `EnsemblePredictor` and `EnsembleUpdater`.


jwragg-dstl (changes requested):
Looks really good. Some documentation for the Ensemble KF is required e.g: .. automodule:: stonesoup.updater.EnsembleKalmanFiter
    :show-inheritance:

taken from the Stone Soup source documentation.

0sm1um:
> 

Thank you! I believe I have added the proper headers in the updater and predictor sections. I also fixed a docstring not rendering properly on the EnsembleState class.

I have two other questions I would like to ask.

First, is since I have my paper available to reference. Is there a prescribed way to reference or cite them them in the documentation itself?

Second, after speaking with a co author of my paper, is that I believe it may be necessary to update or add copyrights/liscence file which states something along the lines of:

"Copyright © 2020-2021 John Hiles
Portions of this work were created while John Hiles was employed by Wright State University under Contract No. FA8650-18-2-1645 to the Air Force Research Laboratory, and the University has released its rights to this Student work."

My understanding is that this is neccisary to ensure that the public release that I obtained is properly memorialized or acknowledged. Do you all have a set protocol on how you would like this to be done? I would suggest that I add a file on my branch which has this information and it can be removed after the merge.

sdhiscocks:
> I have two other questions I would like to ask.
> 
> First is since I have my paper available to reference; is there a prescribed way to reference or cite them them in the documentation itself?
> 

You can include a reference in the doc string. There are example in the code like [the square root Kalman filter](https://github.com/dstl/Stone-Soup/blob/d38a196fa91dc7015355903f3cfcf5d68792d782/stonesoup/updater/kalman.py#L379-L399).

We also have #199 on the TODO list, where we'd like to include list of references to papers that use or about Stone Soup. We'll be sure to include your paper.

> Second, after speaking with a co author of my paper, is that I believe it may be necessary to update or add copyrights/liscence disclaimer which states something along the lines of:
> 
> "Copyright © 2020-2021 John Hiles Portions of this work were created while John Hiles was employed by Wright State University under Contract No. FA8650-18-2-1645 to the Air Force Research Laboratory, and the University has released its rights to this Student work."
> 
> My understanding is that this is neccisary to ensure that the public release that I obtained is properly acknowledged. Do you all have a set protocol on how you would like this to be done? I would suggest that I add a file or put this in the LISCENCE file on my branch which has this information and it can be removed after the merge.

Please go ahead and edit the `LICENSE` file to include this on the PR. As the code will remain in the code base, it can remain in the file going forwards.

0sm1um:
I updated the liscence file and added refrence to our paper in the docstring(as well as making the docstrings just generally better).

sdhiscocks (approved):
Looks good 👍 

Testing, I think we could do with adding an initiator for ensemble states, even if initially like the [` GaussianParticleInitiator`](https://github.com/dstl/Stone-Soup/blob/59c636fb2b632e19901c4eeeed4500257566c165/stonesoup/initiator/simple.py#L224), where we can use existing Gaussian initiators and `EnsembleState.from_gaussian_state` method. I'll raise an issue for this, and can be done separately to this PR.

PR #628, opened 2022-05-02
Prevent track from birth component w/o detection

ekhunter123:
Problem: In Gaussian-mixture based trackers, tracks can currently be made from the birth component without a detection. 

The `GaussianMixtureHypothesiser` creates prediction-detection pairs as hypotheses. Each prediction is paired with each `TrueDetection` and also with a `MissedDetection`. If the prediction is from the birth component, the hypothesis will get assigned a new tag. Otherwise, the hypothesis will carry the tag from the predicted component. But if the hypothesis is from the birth component with a `MissedDetection`, then it should not get a new tag. It should keep the "birth" tag because it does not correspond to any real data. This problem is realized later, when the `PointProcessMultiTargetTracker` checks the tag of each component. If a component does not have the 'birth' tag, then it will try to make a track or add to an existing one. For this if-statement to be effective, we need the birth component to keep its birth tag.

To fix this, I added a new conditional when assigning tags to hypotheses.

---

Also, pointprocess.py use '0' as the tag of the birth component, , while these other files use 'birth':
https://github.com/dstl/Stone-Soup/blob/d38a196fa91dc7015355903f3cfcf5d68792d782/stonesoup/tracker/tests/test_pointprocess.py#L26

https://github.com/dstl/Stone-Soup/blob/d38a196fa91dc7015355903f3cfcf5d68792d782/stonesoup/hypothesiser/gaussianmixture.py#L64
I changed pointprocess.py to also use 'birth'.

It may be good to explicitly say that the birth component must have tag 'birth' in the docstring of the `birth_component`, and/or to check for it in the constructor and raise an error otherwise. Thoughts?

### DIFF
--- a/stonesoup/types/state.py
+++ b/stonesoup/types/state.py
@@ -420,6 +420,8 @@ class TaggedWeightedGaussianState(WeightedGaussianState):
     """
     tag: str = Property(default=None, doc="Unique tag of the Gaussian State.")
 
+    BIRTH = 'birth'
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.tag is None:
--- a/stonesoup/types/state.py
+++ b/stonesoup/types/state.py
@@ -421,6 +421,7 @@ class TaggedWeightedGaussianState(WeightedGaussianState):
     tag: str = Property(default=None, doc="Unique tag of the Gaussian State.")
 
     BIRTH = 'birth'
+    '''Tag value used to signify birth component'''
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/docs/tutorials/11_GMPHDTutorial.py
+++ b/docs/tutorials/11_GMPHDTutorial.py
@@ -332,7 +332,7 @@ for truth in start_truths:
             state_vector=truth.state_vector,
             covar=covar**2,
             weight=0.25,
-            tag='birth',
+            tag=TaggedWeightedGaussianState.BIRTH,
             timestamp=start_time)
     tracks.add(Track(new_track))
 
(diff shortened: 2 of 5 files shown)


sdhiscocks:
> It may be good to explicitly say that the birth component must have tag 'birth' in the docstring of the `birth_component`, and/or to check for it in the constructor and raise an error otherwise. Thoughts?

Sounds good. I think we should probably make an explicit definition for it to avoid issues with assuming what the value is.
```diff
diff --git a/stonesoup/types/state.py b/stonesoup/types/state.py
index b86c957f..968bbd18 100644
--- a/stonesoup/types/state.py
+++ b/stonesoup/types/state.py
@@ -420,6 +420,8 @@ class TaggedWeightedGaussianState(WeightedGaussianState):
     """
     tag: str = Property(default=None, doc="Unique tag of the Gaussian State.")
 
+    BIRTH = 'birth'
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.tag is None:

```

such you can check with  for example `if component.tag == component.BIRTH:`

jwragg-dstl (commented):
All reviewed. Nothing to add from me. Looks to retain the 'Birth Tag' correctly. Thank you very much.

jwragg-dstl (approved):
Just adding the approval. Thank you.

stonesoup/types/state.py
```diff
@@ -420,6 +420,8 @@ class TaggedWeightedGaussianState(WeightedGaussianState):
     """
     tag: str = Property(default=None, doc="Unique tag of the Gaussian State.")
 
+    BIRTH = 'birth'
```
sdhiscocks, line 423: Just noted this hasn't appeared in the documentation. Adding something like this should work:
```suggestion
    BIRTH = 'birth'
    '''Tag value used to signify birth component'''
```

ekhunter123: Good catch, thanks!

jswright-dstl (approved):

PR #636, opened 2022-05-12
Add Nonparametric PDA and gate out non valid measurements

sdhiscocks:
This adds the ability to use nonparametric PDA for clutter spatial density, where it is estimated based on assumption that measurements within a defined validation region represent the clutter level.

This change also gates out non valid measurements by default, similar to the behaviour in the `DistanceHypothesier`, and per description in original PDA paper.

### DIFF
--- a/stonesoup/hypothesiser/probability.py
+++ b/stonesoup/hypothesiser/probability.py
@@ -1,4 +1,9 @@
-from scipy.stats import multivariate_normal as mn
+from functools import lru_cache
+
+from scipy.stats import multivariate_normal, chi2
+from scipy.linalg import det
+from scipy.special import gamma
+import numpy as np
 
 from .base import Hypothesiser
 from ..base import Property
@@ -21,7 +26,10 @@ class PDAHypothesiser(Hypothesiser):
     predictor: Predictor = Property(doc="Predict tracks to detection times")
     updater: Updater = Property(doc="Updater used to get measurement prediction")
     clutter_spatial_density: float = Property(
-        doc="Spatial density of clutter - tied to probability of false detection")
+        default=None,
+        doc="Spatial density of clutter - tied to probability of false detection. Default is None "
+            "where the clutter spatial density is calculated based on assumption that "
+            "all but one measurement within the validation region of the track are clutter.")
     prob_detect: Probability = Property(
         default=Probability(0.85),
         doc="Target Detection Probability")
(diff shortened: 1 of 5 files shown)


stonesoup/hypothesiser/probability.py
```diff
@@ -118,6 +137,7 @@ def hypothesise(self, track, detections, timestamp, **kwargs):
 
         # True detection hypotheses
         for detection in detections:
+            valid_measurement = False
```
sglvladi, line 140: Maybe move this above line 154 with some comments about gating?

stonesoup/hypothesiser/probability.py
```diff
@@ -126,18 +146,57 @@ def hypothesise(self, track, detections, timestamp, **kwargs):
...
+    @lru_cache()
+    def _validation_region_volume(cls, prob_gate, meas_pred):
+        n = meas_pred.ndim
+        gate_threshold = cls._gate_threshold(prob_gate, n)
+        c_z = np.pi**(n/2) / gamma(n/2 + 1)
+        return c_z * gate_threshold**(n/2) * np.sqrt(det(meas_pred.covar))
+
+    @staticmethod
+    @lru_cache()
+    def _inv_cov(meas_pred):
```
sglvladi, line 196: Not too keen on single-line functions, but I'm guessing this is done to take advantage of caching?

stonesoup/hypothesiser/probability.py
```diff
@@ -126,18 +146,57 @@ def hypothesise(self, track, detections, timestamp, **kwargs):
...
+                valid_measurement = True
+
+            if self.include_all or valid_measurement:
+                probability = pdf * self.prob_detect
+                if self.clutter_spatial_density is None:
+                    # Note: will divide by validated measurements count later...
+                    probability *= self._validation_region_volume(
+                        self.prob_gate, measurement_prediction)
+                else:
+                    probability /= self.clutter_spatial_density
```
sglvladi, line 165:
```suggestion
                if self.clutter_spatial_density is not None:
                    probability /= self.clutter_spatial_density
```

Might be more readable if the calculation for the case ``self.clutter_spatial_density is None`` is all done in one place on lines 175-177, given that we are iterating over the hypotheses and making the check anyway. The comment is helpful, but I think keeping the "equations" together is more helpful.

sglvladi:
Looks good overall 👍 Added some minor notes on making the code more readable.

Note on earlier discussion: I don't think we can avoid gating when doing the non-parametric version. We could avoid it for the parametric one, but then it would be confusing to say we do gating in one case but not the other.

stonesoup/hypothesiser/probability.py
sdhiscocks: Yes. However, there are other modifications with Mahalanobis measure that include built in caching. I may rebase of main and then can take advantage of that and remove the need for this function.

sglvladi (approved):

jswright-dstl (approved):

PR #637, opened 2022-05-13
Add option to specify number of targets in MultiTargetGroundTruthSimulator

ekhunter123:
I added two parameters to the MultiTargetGroundTruthSimulator class. Together, these parameters give the user more control over the number of targets, and their locations, that exist in the simulation. Previously, the number of targets could only be controlled using the birth rate and death probability.

`preexisting_states` is a list of state vectors where the user wants a target to start. These targets will all be initiated at time 0 and will last at least 1 time step (they will not die at time 0). After that, their survival is random according to the death probability.

`initial_number_targets` is an integer dictating the number of targets that the user wants to be randomly created at time 0. This makes it possible to start the simulation with, for example, 10 targets randomly spread in the state space. Previously, the number of targets existing at time 0 varied randomly according to the birth rate. Like the `preexisting_states` targets, these simulated states will not die at time 0. 

**Note:** I tried to make the docstrings clear that `initial_number_targets` is not simply the length of `preexisting states`. Please feel free to make suggestions if this needs clarification. Alternatively, make suggestions if the names of the parameters is not clear.

Thanks to @sdhiscocks for his suggestions and edits about this!

I made this in response to my discussion https://github.com/dstl/Stone-Soup/discussions/630#discussion-4054726.

### DIFF
--- a/stonesoup/simulator/simple.py
+++ b/stonesoup/simulator/simple.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from typing import Optional
 import datetime
-from typing import Sequence
+from typing import Sequence, Tuple
 
 import numpy as np
 from ordered_set import OrderedSet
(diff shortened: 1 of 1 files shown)


stonesoup/simulator/simple.py
```diff
@@ -92,18 +92,58 @@ class MultiTargetGroundTruthSimulator(SingleTargetGroundTruthSimulator):
         default=0.1, doc="Probability of track dying in each time step. Default 0.1.")
     seed: Optional[int] = Property(default=None, doc="Seed for random number generation."
                                                      " Default None")
+    preexisting_states: Tuple = Property(
```
sdhiscocks, line 95: I think the type for this should be `Collection[GaussianState]`

sdhiscocks (approved):

Whaoger (approved):

Whaoger (approved):

stonesoup/simulator/simple.py
ekhunter123: I made a commit to change this. However I made it a Collection[StateVector] because they won't have covariances matrices. I also wanted to make it clear that you are supposed to pass in state vectors only. If it was a State, then it would require a timestamp. I think that could be confusing because it might lead people to think they can dictate when they want the target to appear in the ground truth. But what I have done here has all the states begin at time t=0.

jswright-dstl (approved):

stonesoup/simulator/simple.py
sdhiscocks: Ah, I misread the code and thought it was using covar to sample, but that was `initial_state` I was looking at. 👍

PR #584, opened 2022-01-17
Added 3D Plotting Functionality to plotter.py

PACarniglia:
Adds three-dimensional plotting functionality to the plotter.py function. Plotting of truth data, measurements (including clutter and true detection distinction), and tracks with error bars are supported. The underlying 2D plotting functionality is unchanged. Plotting dimensionality is now selected by the user when instantiating new Plotter objects by passing a plotter.py Dimension Enum object. Currently Dimension.TWO and Dimension.THREE are supported but can be extended in the future.

The legend is now populated with a dictionary where the keys show up as labels and values associated with those keys appear as the handles. This implementation avoids duplicate legend entries if plotting the same data multiple times.

A unit test file, test_plotter.py, was also created to ensure subsequent changes to plotter.py do not break current functionality.

PR #638, opened 2022-05-13
Create MTT 3D Platform example

PACarniglia:
This is a supplementary example to be added to the online Stone Soup documentation. This example uses a platform-based simulation in 3D. A single, ground-based radar sensor attached to a stationary platform takes range/bearing/elevation measurements of airborne targets, represented using moving platforms with motion described by 3D CV models. Targets are birthed and die with 5% probability. The ground-based radar measures both true detections and clutter which is generated using the new ClutterModel implemented by @ekhunter123 in discussion #535. Track management processes are used in conjunction with a JPDA to perform 3D multi-target tracking on the randomly generated tracking scenario. Results are plotted in 3D using the new 3D visualization functionality added in #584. The GOSPA, SIAP, and uncertainty metrics are used to assess the quality of the MTT results. Documentation throughout provides a step-by-step instruction for users to follow along and replicate the results on their own.

The file is Sphinx-ready and can be compiled on a local machine to view the .html file it produces for QA purposes. It is highly desired to have this documentation on the website prior to Monday, May 16 @ 1:30 pm EST so the I2MTC tutorial attendees can follow along at the same location as the rest of the Stone Soup documentation and the hyperlinks in the tutorial navigate to the corresponding pages elsewhere in the documentation. This tutorial was prepared with the assistance of @ekhunter123 and @divyr

### DIFF
--- a/docs/examples/MTT_3D_Platform.py
+++ b/docs/examples/MTT_3D_Platform.py
@@ -1,0 +1,396 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+Multi-Target Tracking in 3D Using Platform Simulation
+=====================================================
+In the Stone Soup library, simulations can be set up and run using special
+:class:`~.FixedPlatform` and :class:`~.Sensor` objects. Simulated data can be preferable to
+real data as the user has more control over the tracking scenario and real
+data can be difficult or costly to acquire.
+"""
+
+# %%
+# Creating the Sensor
+# -------------------
+# We will begin by importing many relevant packages for the simulation.
+
+from datetime import datetime
+from datetime import timedelta
+import numpy as np
+import random
+
+
+# Stone Soup imports:
+from stonesoup.types.state import State, GaussianState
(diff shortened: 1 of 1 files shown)


PACarniglia:
If possible, it would also be ideal if line numbers could be included in the online documentation for the code as it improves the ability to reference a specific line.

sglvladi (approved):
Looks good 👍

docs/examples/MTT_3D_Platform.py
```diff
@@ -0,0 +1,396 @@
...
+
+from stonesoup.metricgenerator.manager import SimpleManager
+metric_manager = SimpleManager(
+    [ospa_generator, siap_generator, uncertainty_generator],
+    associator=associator
+)
+
+
+# %%
+# Since we saveed the groundtruth and tracks before, we can easily add them
```
hpritchett-dstl, line 346: Minor spelling mistake with "saveed" instead of "saved".

sdhiscocks (approved):

PR #581, opened 2022-01-14
Modify ParticleState so it has StateVectors and remove unnecessary Particles class

nperree-dstl:
Currently in stone soup we have a `ParticleState` which is not actually a state and contains `Particles` which have `StateVectors` and weights. This can make working with the particle filter a bit confusing as there are so many layers of particle types.

This pull request changes this structure so that everything needed is in the `ParticleState` and the `Particles` class has been removed. The new `ParticleState` contains `StateVectors` and weights meaning it can be substituted for anywhere that `Particles` was previously used.

`ParticleState` can be initiated using `StateVectors` and a sequence of weights or by using a list of particles (`Particle` types). 

This closes issue #537

PR #644, opened 2022-05-19
Modify EnsembleState to use StateVectors

sdhiscocks:
Similar to change in #581, this uses `StateVectors` as the `EnsembleState` state vector. As well as being more consistent, this also enables use of the vectorised models.


sdhiscocks:
@0sm1um - Wondering if you could also review this PR?

0sm1um:
This is a sensible change from what I can see. When I originally started working on the `ensemblestate` vectorized models did not exist.

hpritchett-dstl (approved):

nperree-dstl (approved):

PR #646, opened 2022-05-20
Add max range to simple Radars

hpritchett-dstl:
Currently the _RadarBearingRange_ and _RadarElevationBearingRange_ have no attribute to set its max range.

This PR adds a max_range, and adds in logic to only produce detections for states that lie within the sensor's FOV (similar to how _RadarRotatingBearingRange_ does it).

### DIFF
--- a/stonesoup/sensor/radar/radar.py
+++ b/stonesoup/sensor/radar/radar.py
@@ -189,6 +189,7 @@ class RadarElevationBearingRange(RadarBearingRange):
         doc="The sensor noise covariance matrix. This is utilised by "
             "(and follow in format) the underlying "
             ":class:`~.CartesianToElevationBearingRange` model")
+    max_range: float = Property(doc="The maximum detection range of the radar (in meters)")
 
     @property
     def measurement_model(self):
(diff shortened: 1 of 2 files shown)


stonesoup/sensor/radar/radar.py
```diff
@@ -189,6 +189,7 @@ class RadarElevationBearingRange(RadarBearingRange):
         doc="The sensor noise covariance matrix. This is utilised by "
             "(and follow in format) the underlying "
             ":class:`~.CartesianToElevationBearingRange` model")
+    max_range: float = Property(doc="The maximum detection range of the radar (in meters)")
```
sdhiscocks, line 192: `default=np.inf`?

hpritchett-dstl: Yes I was thinking that, but wanted to be consistent with _RadarRotatingBearingRange_, maybe both need a change?

stonesoup/sensor/radar/radar.py
```diff
@@ -206,7 +209,23 @@ def measure(self, ground_truths: Set[GroundTruthState], noise: Union[np.ndarray,
...
+            if noise is True:
+                measurement_noise = measurement_model.rvs()
+            else:
+                measurement_noise = noise
+
+            # Check if state falls within range of sensor
+            range_t = measurement_vector[2, 0]  # Elevation(0), Bearing(1), Range(2)
+            # Do not measure if state is out of range
+            if range_t > self.max_range:
+                continue
```
sdhiscocks, line 224: An simple extra test to cover this line would be good.

sdhiscocks (approved):

jswright-dstl (approved):

PR #649, opened 2022-05-26
Add parameter to limit number of components in GaussianMixtureReducer

ekhunter123:
In this PR I've added two new parameters to the `GaussianMixtureReducer`: `max_number_components` and `truncating`. Together, these let the user limit the number of components being held in the gaussian mixture. After the pruning and merging is complete, only the `max_number_components` components with the highest weights will be kept.

Having this option makes the reducer more consistent with the literature:
- [Original GMPHD filter proposed by Vo and Ma](https://ieeexplore.ieee.org/document/1710358) 
- [Followup paper by Clark, Panta, and Vo](https://ieeexplore.ieee.org/document/4086095)

There are a few notes I want to discuss and get opinions on.
1. In Clark, Panta, and Vo's paper about the GMPHD filter and its implementation, they added a step in the pruning where the weights from the pruned components are "redistributed" to the remaining components. This was not done in Vo and Ma's original GMPHD filter, but does make sense. It is done in Stone Soup here
https://github.com/dstl/Stone-Soup/blob/ee9914b339bb16abb3b1c3b7a7b09489fb75da7c/stonesoup/mixturereducer/gaussianmixture.py#L82-L85

Later in Clark et al.'s paper they give the pseudocode for keeping only a certain number of components. But the weight redistribution is not done for those deleted components. Does anyone happen to know why this is? I have added the redistribution to the code in my PR, but it is easy to take out if we want to keep consistent with the papers.

2. The two papers linked above are not consistent about the naming of the steps. Vo and Ma call the entire reduction process "pruning". Clark et al. split it into "pruning" and "merging" (which is done in Stone Soup), but consider limiting the maximum number of components to be part of "merging". It seems to me that it is a different idea, so I've added a third function called `truncate()` which is called if the `truncating` parameter is True. If there are other ideas about the naming of this function or of the parameters please feel free to comment! 

3. The merging process iteratively looks at the component with the highest weight (component A), and looks to see if there is another component (component B) that is close enough to be merged. Suppose m_A, m_B, P_A, and P_B are the means and covariances of components A and B, and U is the merging threshold. In the literature, this distance is calculated as

 In the `GaussianMixtureReducer.merge()` function, the distance between two states is calculated using the

[scipy mahalanobis](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.mahalanobis.html#scipy.spatial.distance.mahalanobis) function called with m_A, m_B, and P_A. Note that Scipy expects you to pass in the inverse of the covariance matrix, which we don't do. This means it is calculated as 

The difference is that it uses P_A rather than the inverse of P_B, and takes the square root. 
Is there a reason that it is done this way in Stone Soup? I'm not trying to be critical, I just want to understand. Thanks!


sdhiscocks:
@Campbell101: would welcome your comments on this.

Campbell101:
@sdhiscocks @ekhunter123 Sorry, was swamped then on holiday there! Sure I'll look at it this week if you still need me to

sdhiscocks:
> @sdhiscocks @ekhunter123 Sorry, was swamped then on holiday there! Sure I'll look at it this week if you still need me to

Thanks @Campbell101. It would be great to get your thoughts.

Campbell101:
@ekhunter123 I think this is a great idea and well worth doing! To answer your questions:

1. I can't say why it wasn't included in the conference paper (Clark et al.). It shouldn't matter too much if the redistribution of the pruned weights is included as its usually fairly insignificant (as long as the pruning threshold is low enough) when compared to the rest of the mixture. I would personally keep it in as it doesn't really make any computational difference to take it out.
2. I would call the entire procedure "mixture reduction" and splitting it into separate steps seems appropriate for this stone soup implementation. The user might want to pick and choose what techniques or merging/pruning metrics they want. Truncation is a good name.
3. Good catch, this is a bug! I believe it should use the Mahalonobis distance from the stone soup metrics package.

ekhunter123:
Thanks for your comments, @Campbell101. I have reached out to Daniel Clark about my question in 1. For now we can leave the redistribution in there.

I added another commit to fix that little bug with the Mahalanobis distance. I can make that fix into a separate PR if that would be better, @sdhiscocks. Note that the distance is now squared to keep in line with the literature. This will have an impact on any existing code which uses the GM-PHD (it is equivalent to using the square root of a previously-used merging threshold).

I am also going to make this a real pull request now, not just a draft. Thanks for your help!

sdhiscocks:
> Thanks for your comments, @Campbell101. I have reached out to Daniel Clark about my question in 1. For now we can leave the redistribution in there.
> 
> I added another commit to fix that little bug with the Mahalanobis distance. I can make that fix into a separate PR if that would be better, @sdhiscocks. Note that the distance is now squared to keep in line with the literature. This will have an impact on any existing code which uses the GM-PHD (it is equivalent to using the square root of a previously-used merging threshold).
> 
> I am also going to make this a real pull request now, not just a draft. Thanks for your help!

I think a separate PR would be preferable, but not essential, as we can more clearly document the breaking change (it terms of parameters that need to be set).

stonesoup/mixturereducer/gaussianmixture.py
```diff
@@ -161,8 +175,7 @@ def merge(self, components_list):
             # (modifying list in loop, so copy used)
             for component in remaining_components.copy():
                 # Calculate distance between component and best component
-                distance = dist.mahalanobis(
-                    best_component.mean[:, 0], component.mean[:, 0], best_component.covar)
+                distance = (Mahalanobis(state1=component, state2=best_component))**2
```
sdhiscocks, line 178:
```suggestion
                distance = (Mahalanobis()(state1=component, state2=best_component))**2
```
May also just want to define `Mahalanobis()` once to avoid recreating every loop.

sdhiscocks (approved):
Thanks @ekhunter123 

There is one warning on the docs build still, but this is fixed in `main` so you can ignore that.

nperree-dstl (approved):

jswright-dstl (approved):

Issue #655, opened 2022-06-10
Ability to pass figure arguments or provide existing axis to Plotter


sdhiscocks:
It would be useful to pass custom arguments to figure initialisation (e.g. fig size) to the Plotter. Also, it would be useful to be able to pass in an existing axis object, which is then plotted on. This would enable creating figures with multiple subplots.

sdhiscocks:
Figure related arguments resolved with #661.

Would still be good to pass existing axis objects in order to make subplots.

csherman-dstl:
#1266

PR #661, opened 2022-06-20
Add additional functionality into Plotter class.

jswright-dstl:
This PR adds the ability to pass custom arguments to figure initialisation as highlighted in #655. For example, `figsize` can be passed to Plotter in initialisation to change the figure size. 
Additionally, the method to set equal axes in 3d plots. This uses the data in the figure to set the axes.

### DIFF
--- a/stonesoup/plotter.py
+++ b/stonesoup/plotter.py
@@ -1,6 +1,5 @@
 import warnings
 from itertools import chain
-from typing import Optional
 
 import numpy as np
 from matplotlib import pyplot as plt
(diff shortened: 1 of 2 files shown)


stonesoup/plotter.py
```diff
@@ -40,6 +42,8 @@ class Plotter:
     ----------
     dimension: enum \'Dimension\'
         Optional parameter to specify 2D or 3D plotting. Default is 2D plotting.
+    figsize: tuple
+        Optional parameter to specify the size of the figure. Default is (10, 6).
```
sdhiscocks, line 46: I think this should be documented like the other methods.

https://github.com/dstl/Stone-Soup/blob/bfecfd6596ab1d877dc3d895ad3fbce1f7aa459f/stonesoup/plotter.py#L95-L96

sdhiscocks (approved):

hpritchett-dstl (approved):

PR #666, opened 2022-06-27
Add example of track-to-track fusion

ekhunter123:
Add example of track-to-track fusion using the new Bayesian formulation of the covariance intersection algorithm from Clark and Campbell [(link)](https://www.techrxiv.org/articles/preprint/Integrating_Covariance_Intersection_into_Bayesian_multi-target_tracking_filters/19351931).

I have run Sphinx to check the formatting, here is what it looks like as a PDF: 
[Track2Track_Fusion_Example.pdf](https://github.com/dstl/Stone-Soup/files/8992963/Track2Track_Fusion_Example.pdf)

This work was done with Daniel Clark, Sean O'Rourke (@orourkesm-afrl), and Bhashyam Balaji (@bhashyambalaji).


docs/examples/Track2Track_Fusion_Tutorial.py
```diff
@@ -0,0 +1,876 @@
...
+# %%
+# The rest of this example will continue as follows:
+#   * Create a simulator for the ground truth
+#   * Create 2 radar simulators, one on the ground and one that is airborne
+#   * Make a JPDA tracker for the first radar, and a Gaussian mixture linear complexity with
+#     cumulants (GM-LCC) tracker for the second. These will mimic the situation where the radar
+#     sensors outputs tracks instead of raw measurements.
+#   * Create a GM-PHD tracker that will perform measurement fusion, using all measurements from
+#     both radars. This is created to compare with the covariance intersection method.
+#   * Define the ChernoffUpdater class, which implements the covariance intersection
```
nperree-dstl, line 84:
```suggestion
#   * Define the :class:`ChernoffUpdater` class, which implements the covariance intersection
```

docs/examples/Track2Track_Fusion_Tutorial.py
```diff
@@ -0,0 +1,876 @@
...
+#   * Define a Detector class which converts tracks into Gaussian measurements
+#   * Create a GM-PHD tracker using the ChernoffUpdater for the covariance intersection
```
nperree-dstl, line 86:
```suggestion
#   * Create a GM-PHD tracker using the :class:`ChernoffUpdater` for the covariance intersection
```

docs/examples/Track2Track_Fusion_Tutorial.py
```diff
@@ -0,0 +1,876 @@
...
+from stonesoup.models.clutter.clutter import ClutterModel
+clutter_model = ClutterModel(
+    clutter_rate=2.0,
+    distribution=np.random.default_rng().uniform,
+    dist_params=((-600.0, 600.0), (-600.0, 600.0), (250.0, 750.0))
+)
+
+# %%
+# The first radar will be airborne, at an altitude of approximately 3000 m. It makes detections with
+# an elevation, bearing, and range measurement model. By setting the max_range to 3500, we can
```
nperree-dstl, line 140:
```suggestion
# an elevation, bearing, and range measurement model. By setting the `max_range` to 3500, we can
```

docs/examples/Track2Track_Fusion_Tutorial.py
```diff
@@ -0,0 +1,876 @@
...
+from stonesoup.platform.base import FixedPlatform
+sensor2_platform = FixedPlatform(
+    State([2000, 0, 50, 0, 0, 0]),
+    position_mapping=[0, 2, 4],
+    sensors=[radar2]
+)
+
+# %%
+# Now we can pass the platforms into a detection simulator. At each timestep, the simulator will
+# return the detections from the sensor1 platform, then the detections from the sensor2 platform.
```
nperree-dstl, line 199:
```suggestion
# return the detections from the `sensor1_platform`, then the detections from the `sensor2_platform`.
```

docs/examples/Track2Track_Fusion_Tutorial.py
```diff
@@ -0,0 +1,876 @@
...
+    updater=deepcopy(updater),
+    reducer=deepcopy(reducer),
+    birth_component=deepcopy(birth_component),
+    extraction_threshold=0.90,
+)
+
+# %%
+# 5: Define the Chernoff Updater Class
+# ------------------------------------
+# The ChernoffUpdater is where we will implement the covariance intersection. As an
```
nperree-dstl, line 412:
```suggestion
# The :class:`ChernoffUpdater` is where we will implement the covariance intersection. As an
```

docs/examples/Track2Track_Fusion_Tutorial.py
```diff
@@ -0,0 +1,876 @@
...
+# :class:`~.Updater`, the ChernoffUpdater will need functions `predict_measurement()` and
```
nperree-dstl, line 413:
```suggestion
# :class:`~.Updater`, the :class:`ChernoffUpdater` will need the functions `predict_measurement()` and
```

docs/examples/Track2Track_Fusion_Tutorial.py
```diff
@@ -0,0 +1,876 @@
...
+# --------------------
+# Next, we will plot all of the resulting tracks and measurements. This will be done in two plots.
+# The first plot will show all of the data, and the second plot will show a closer view of one
+# resultant track.
+#
+# These plots are done in 2D to make them more readable. They are still very busy, so you are
+# invited to run it yourself and remove some of the items from the plot. Or, if you are running
+# the notebook locally, add the next line to your code to make it an interactive plot.
+#
+# %matplotlib widget
```
nperree-dstl, line 740: Is this needed in the docs?

nperree-dstl (commented):
Tutorial looks good! Just some really small documentation comments from looking at the pdf version.

docs/examples/Track2Track_Fusion_Tutorial.py
ekhunter123: Maybe not? I added it as an excuse for the very messy plot, it is a lot easier to see using the widget mode. But I can remove it if you think it doesn't quite fit!

ekhunter123: I've changed the wording a little. I think it is still useful to have in an example like this. I'm hoping it will encourage users to explore the plots and results more closely on their own, as that is the best way to fully understand the material.

ekhunter123:
Now that the ChernoffUpdater and the track feeder are merged into main, I could remove their class definitions in this example. Or is it helpful to have them here for clarity?

docs/examples/Track2Track_Fusion_Tutorial.py
nperree-dstl: No it's okay, if it's needed for the plotting leave it in - I just wasn't sure if it did anything on the documentation version

nperree-dstl:
> Now that the ChernoffUpdater and the track feeder are merged into main, I could remove their class definitions in this example. Or is it helpful to have them here for clarity?

Yeah I would probably remove them from the example as it is quite a long one and let people explore the classes themselves if they want to

jswright-dstl (approved):

ekhunter123:
I removed the class definitions for the ChernoffUpdater and the Tracks2GaussianDetectionFeeder, and reworked the explanation to accommodate this. Here is the latest Sphinx rendering

[T2TFusionExample.pdf](https://github.com/dstl/Stone-Soup/files/9243102/T2TFusionExample.pdf)

nperree-dstl (approved):

docs/examples/Track2Track_Fusion_Example.py
```diff
@@ -0,0 +1,784 @@
...
+    plotter.ax.set_position([pos.x0, pos.y0, pos.width * 0.7, pos.height])
+    k = list(plotter.legend_dict.keys())
+    k[2], k[3] = k[3], k[2]
+    v = list(plotter.legend_dict.values())
+    v[2], v[3] = v[3], v[2]
+    plotter.ax.legend(handles=v, labels=k, loc='lower center', bbox_to_anchor=(0.5, -0.5))
+
+plotter1.fig.show()
+
+track = track_fusion_tracks.pop()
```
sdhiscocks, line 674: Building on CircleCI currently fails as `track_fusion_tracks` is an empty set.

sdhiscocks (changes requested):

docs/examples/Track2Track_Fusion_Example.py
ekhunter123: Ah I see, thanks Steve! I added an if statement to check if the track_fusion_tracks has any tracks in it before plotting. I've run the notebook so many times on my own, and it very rarely happens that there are no tracks. So I think this is an adequate solution.

sdhiscocks:
I'm not sure why, but when running on CircleCI, no tracks are produced. But it works fine locally.

I'll rebase on `main` to see if that resolves it.

sdhiscocks:
So the issue was in certain configurations, the prune threshold was slightly too high. I recreated this on another Linux system running Python 3.10. I suspect this is caused by floating point precision discrepancy, either with Linux, Python 3.10, or version of NumPy or 3rd party libraries. Increase the threshold seems to have resolved the issue.

sdhiscocks (approved):

Issue #668, opened 2022-06-27
Ensemble Predictor does not call vectorized models correctly


0sm1um:
The code in Pull Request #644 introduced a bug to the `EnsemblePredictor` which causes the Ensemble Kalman Filter to quickly diverge/fail.

Upon calling the model, the transition model essentially populates the prediction ensemble with a `StateVectors` instance of N identical vectors. In other words an nxN `StateVectors` matrix of identical column vectors.

In Pull Request #644, the `EnsembleState` and `EnsemblePredictor` were changed to utilize the vectorized models which were introduced to speed up the Particle Filter in PR #365.

0sm1um:
The fix for this is already implemented in PR #669

sdhiscocks:
Thanks @0sm1um. Looks like more fundamental issue of calls to the `rvs` method not setting the sample size. Luckily this seems to only effect particle filter (which has same `num_samples` fix you used); and of course the Ensemble filter.

Here's a proposed alternative fix, which deals with the issue more generally.
```diff
diff --git a/stonesoup/models/base.py b/stonesoup/models/base.py
index b4b1bd8c..27bc5471 100644
--- a/stonesoup/models/base.py
+++ b/stonesoup/models/base.py
@@ -134,7 +134,7 @@ class LinearModel(Model):
         """
         if isinstance(noise, bool) or noise is None:
             if noise:
-                noise = self.rvs(**kwargs)
+                noise = self.rvs(num_samples=state.state_vector.shape[1], **kwargs)
             else:
                 noise = 0
 
diff --git a/stonesoup/models/measurement/linear.py b/stonesoup/models/measurement/linear.py
index d03e195a..3fe32d8c 100644
--- a/stonesoup/models/measurement/linear.py
+++ b/stonesoup/models/measurement/linear.py
@@ -75,7 +75,7 @@ class LinearGaussian(MeasurementModel, LinearModel, GaussianModel):
 
         if isinstance(noise, bool) or noise is None:
             if noise:
-                noise = self.rvs()
+                noise = self.rvs(num_samples=state.state_vector.shape[1], **kwargs)
             else:
                 noise = 0
 
diff --git a/stonesoup/models/measurement/nonlinear.py b/stonesoup/models/measurement/nonlinear.py
index 64ccb5ad..054a33e4 100644
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
[...]

PR #669, opened 2022-06-27
Ensemble Square Root Updater

0sm1um:
Here is my implementation of the Ensemble Square Root filter Updater which I presented at last year's Fusion Conference.

A full writeup of the algorithm itself, as well as design philosophy can be found in my paper titled "[Implementation of Ensemble Kalman Filters in Stone Soup](http://dx.doi.org/10.23919/FUSION49465.2021.9626923)". The contribution here (in this particular pull request) is specifically a new updater. The predictor of the EnKF and EnSRF are shared, but the EnSRF assimilates measurements without adding noise which typically results in lower sampling error and thus higher accuracy.

One other thing to note. Is that this pull request also fixes issue #668 which caused both the `EnsemblePredictor` and the `predict_measurement` method in the EnKF (the latter of which is inherited by the EnSRF).


stonesoup/updater/ensemble.py
```diff
@@ -118,19 +114,18 @@ def predict_measurement(self, predicted_state, measurement_model=None,
         measurement_model : :class:`~.MeasurementModel`
             The measurement model. If omitted, the model in the updater object
             is used
-
```
sdhiscocks, line 121: Some of the blank lines are required, as they are breaking formatting in some places.

jwragg-dstl (approved):

0sm1um:
It should be noted that this pull request now includes the changes outlined in the discussion of Issue #668.

This include changes to numerous `model` classes, as well as the Partifle filter. Currently these changes only affect the Ensemble and Particle filters respectivley.

sdhiscocks (approved):

PR #676, opened 2022-07-22
Enable use of kDTree with Gaussian Mixture Hypothesiser

sdhiscocks:
This enables use of kDTree with Gaussian Mixture Hypothesiser, adds ability to set tree search distance based on track covariance and fixes existing tree data associator tests.


jswright-dstl (approved):

ekhunter123 (approved):
This seems like an excellent addition. It will be much more efficient than the basic gating previously done using the `missed_distance` parameter.

Issue #673, opened 2022-07-07
Iterating over StateVectors


jmbarr:
You may wish to do: 

`for a_state_vector in state_vectors:
    do some operation on a_state_vector`

For a set of `StateVector` in a `StateVectors` object.

But currently that doesn't work - you get an array of each state vector element for all 'particles'. To get an individual state vector you need to do `for a_state_vector in state vectors.T:` and then cast `a_state_vector` to the `StateVector` type.

Could/should we overwrite the `__iter__()` method in `StateVectors` so it does the sensible thing of running through the state vectors by default?

sdhiscocks:
Completed in #677

PR #677, opened 2022-07-22
State vectors iterator

jmbarr:
Does as suggested in #673. Iterating over `StateVectors` now yields a `StateVector`. Note that this breaks a few tests (which have been amended), so may also break iterations which previously accommodated the old way. 

Also fails doc tests. No idea why that is and doesn't appear to be to do with these changes.


sdhiscocks:
> Also fails doc tests. No idea why that is and doesn't appear to be to do with these changes.

Looks like these changes cause the _Moving Platform Simulator_ example to hang.

stonesoup/types/array.py
```diff
@@ -180,14 +185,15 @@ def _average(state_vectors, axis=None, weights=None, returned=False):
         elif axis == 1:  # Need to handle special cases of averaging potentially
             state_vector = StateVector(
                 np.empty((state_vectors.shape[0], 1), dtype=state_vectors.dtype))
-            for dim, row in enumerate(state_vectors):
+            for dim, row in enumerate(state_vectors.T):
```
sdhiscocks, line 188:
```suggestion
            for dim, row in enumerate(np.asarray(state_vectors)):
```

stonesoup/types/array.py
```diff
@@ -180,14 +185,15 @@ def _average(state_vectors, axis=None, weights=None, returned=False):
...
                 type_ = type(row[0])  # Assume all the same type
                 if hasattr(type_, 'average'):
                     # Check if type has custom average method
                     state_vector[dim, 0] = type_.average(row, weights=weights)
                 else:
                     # Else use numpy built in, converting to float array
-                    state_vector[dim, 0] = type_(np.average(np.asfarray(row), weights=weights))
+                    state_vector[dim, 0] = type_(np.average(np.asfarray(row).flatten(),
+                                                            weights=weights))
```
sdhiscocks, line 196:
```suggestion
                    state_vector[dim, 0] = type_(np.average(row, weights=weights))
```

jswright-dstl (approved):

nperree-dstl (approved):

stonesoup/models/measurement/tests/test_combined.py
```diff
@@ -72,10 +72,10 @@ def test_rvs(model):
...
     rvs_state = model.rvs(10)
     assert rvs_state.shape == (4, 10)
-    assert all(isinstance(state, Bearing) for state in rvs_state[0])
-    assert all(not isinstance(state, Bearing) for state in rvs_state[1])
-    assert all(isinstance(state, Bearing) for state in rvs_state[2])
-    assert all(not isinstance(state, Bearing) for state in rvs_state[3])
+    assert all(isinstance(state[0], Bearing) for state in rvs_state[0])
+    assert all(not isinstance(state[0], Bearing) for state in rvs_state[1])
+    assert all(isinstance(state[0], Bearing) for state in rvs_state[2])
+    assert all(not isinstance(state[0], Bearing) for state in rvs_state[3])
```
sdhiscocks, line 78: I was a little confused by this change, but now looking more closely, as I can see it's an issue caused by ending up with `StateVectors` type when slicing in certain ways. The new iteration method has highlight this bug here.

I'll push commit here with fix for that, which probably will also resolve some other issues.

PR #687, opened 2022-08-05
Drop # -*- coding: utf-8 -*-and # coding: utf-8 from files

jonasdlindner:
Because Python 3 is already in encoded in UTF-8 (https://peps.python.org/pep-3120/) and Python 2 is not supported anymore, I suggest dropping these lines.


sdhiscocks (approved):
Thanks @jonasdlindner.

I added a commit to fix one of the examples building rather than using the cached version.

Also, ignore the failed tests, this is caused by issue with `pytest-flake8` plugin, with a commit to fix as part of #685 which hopefully will be merged soon as well.

jswright-dstl (approved):

PR #691, opened 2022-08-11
Modify uncertainty reward so sensors measure all tracks together

sdhiscocks:
This can be important in case of sensor handling obscuration, or limited resolution. It also offers a small performance benefit.

### DIFF
--- a/stonesoup/sensormanager/reward.py
+++ b/stonesoup/sensormanager/reward.py
@@ -1,16 +1,17 @@
-from copy import deepcopy
-import numpy as np
-import datetime
-
 from abc import ABC
+import copy
+import datetime
 from typing import Mapping, Sequence, Set
-from stonesoup.base import Base, Property
-from stonesoup.predictor.kalman import KalmanPredictor
-from stonesoup.updater.kalman import ExtendedKalmanUpdater
-from stonesoup.types.track import Track
-from stonesoup.types.hypothesis import SingleHypothesis
-from stonesoup.sensor.sensor import Sensor
-from stonesoup.sensor.action import Action
+
+import numpy as np
+
+from ..base import Base, Property
+from ..predictor.kalman import KalmanPredictor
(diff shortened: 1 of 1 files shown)


stonesoup/sensormanager/reward.py
```diff
@@ -80,57 +81,43 @@ def __call__(self, config: Mapping[Sensor, Sequence[Action]], tracks: Set[Track]
...
-
-                previous_cov_norm = np.linalg.norm(previous.covar)
-
-                predicted_track = Track(previous, init_metadata=dict(Length=3, Width=1))
+        # Create dictionary of predictions for the tracks in the configuration
+        predicted_tracks = set()
+        for track in tracks:
+            predicted_track = copy.copy(track)
+            predicted_track.states = copy.copy(predicted_track.states)
+            predicted_track.metadatas = copy.copy(predicted_track.metadatas)
```
nperree-dstl, line 99: Why do you need to copy everything separately here and why are they copies of themselves? (Probably just me not understanding!)

nperree-dstl (approved):

stonesoup/sensormanager/reward.py
sdhiscocks: When you append states it'll modify the original list (and metadatas list), which you don't want. But if you deep copy the entire track to avoid this, it'll also copy all the states of the track which isn't required and super slow for any track with a lengthy history.

I should add a comment to make this clear in the code.

sdhiscocks: In fact, having thought about this, I think we should implement a proper shallow copy method like a list/dictionary would have. I'll create a new PR for this once this is merged. It'll be same as here effectively, but part of the class, so a plain `copy.copy(track)` will ensure you get a shallow copy, and adding/removing states won't effect the original.

jswright-dstl (approved):

PR #693, opened 2022-08-11
Changed ruamel version requirement to 0.16.5

gawebb-dstl:
Version 0.16.5 is needed to allow multiple types in ruamel.yaml.Yaml which serialise.py uses


nperree-dstl (approved):
Seems to pass tests for me locally

jwragg-dstl (approved):

PR #697, opened 2022-08-17
Update install instructions to make compatible with zsh

DomonkosCs:
I've recently tried to install StoneSoup using the provided `python -m pip install -e .[dev]` command, and it didn't work. I've found out, that when using `zsh`, the location and the extra have to be enclosed by quotation marks. For the sake of other `zsh` users, I think a quick instruction could be given in the docs.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -62,6 +62,14 @@ following:
     cd Stone-Soup
     python -m pip install -e .[dev]
 
+If you are using ``zsh``:
+
+.. code::
+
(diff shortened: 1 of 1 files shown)


sdhiscocks:
Thanks @DomonkosCs. I think to keep things simple, we could just have the quotations marks adding to the original command, as that'll then make that one command compatible with all.

DomonkosCs:
Ok, good idea!

sdhiscocks (approved):

jswright-dstl (approved):

PR #704, opened 2022-08-24
Move additional filter tutorials to subsection and fix binder post build script

sdhiscocks:
binder build script was missing `plotly` dependency


nperree-dstl (approved):
Looks good! While we're at it can we add `sphinx_gallery_thumbnail_number = 3` to the Information Filter tutorial so it's the same as the others?

jswright-dstl (approved):

Issue #354, opened 2021-01-06
Add tutorial/example of adding a custom Detection Reader


sdhiscocks:
Many use cases requiring using custom data, beyond just CSV, so an example of this would be very useful.

mharris-dstl:
Addressed in [707](https://github.com/dstl/Stone-Soup/pull/707).

PR #707, opened 2022-08-31
Added an example within the documentation for custom readers supporting pandas DataFrames.

BenjaminFraser:
Added a new example (`Custom_Pandas_Dataloader.py`) within the documentation in docs/examples for the definition of custom Readers that support pandas DataFrames. 

This allows a wide range of data formats supported by pandas to be taken advantage of for Ground Truth Readers and Detection Readers, without the need manually define custom data ingestion processes for each type, e.g. JSON, XML, Parquet, HDF5, .txt, .zip.

Given its similarity to the requirements of the custom reader documentation example (#354), I've linked this pull request to that, which hopefully is not a problem.

These classes do have the disadvantage of requiring the entire dataset in memory. However, it seems that the ability to directly use pandas DataFrames is a feature several users of Stonesoup have shown interest in, which is understandable given the flexibility and processing functionalities this can provide.

The example in `Custom_Pandas_Dataloader.py` includes the definitions of `DataFrameGroundTruthReader` and `DataFrameDetectionReader` classes. Each of these inherit from the existing `GroundTruthReader` class, along with a custom defined `_DataFrameReader` class. 

These classes operate similarly to the existing `CSVGroundTruthReader` and `CSVDetectionReader` classes, except they take as input a pandas DataFrame already read into memory, rather than a path to .csv file. They also have modified generator functions for producing the time and paths / detections. 

These have been useful for some work I've done using Stonesoup for some UAV-based non-cooperative radar research, and so hopefully they are also of value to other members of the community!

**Update on progression and fixes to aspects of this PR, as of 22 Oct 22:**
1. Added pandas to dev of setup.cfg.
2. Updated references to Stone Soup to be consistent - two words throughout documentation.
3. Added demonstration of ground truth reader after initialisation by outputting first iteration to docs.
4. Added support for fields already in DateTime format.
5. Added pandas_reader.py within reader directory with the three new classes: _DataFrameReader, DataFrameGroundTruthReader, and DataFrameDetectionReader. Tests are still to be developed for these (hence failing on the draft commits currently).
6. Added tests in test_pandas_reader.py within stonesoup/reader/tests. 

A point noted with the tests is that there is currently full coverage of all classes defined in pandas_reader.py, however Codecov flags the pandas import check (which raises an import error if pandas is not installed) as failed. 

To-do / enhancements:
1. Take advantage of pandas grouping to make code more efficient (as suggested by Steven below).
2. Link documentation example to the classes defined within pandas_reader.py, using something such as [`inspect.getsource`](https://docs.python.org/3/library/inspect.html#inspect.getsource).


sdhiscocks:
Thanks for the contribution @BenjaminFraser.

I see docs are failing to build due to _pandas_ being missing dependency. If you could add _pandas_ the `dev` dependencies in `setup.py` that should resolve it: 
https://github.com/dstl/Stone-Soup/blob/435883a67045f72161355e9a5cbb44bcacfa67b1/setup.py#L31-L35

It'd be good to have the readers in the main code base (probably with an optional dependency on _pandas_) so users can easily access them. And also good to keep the example you've created as both a how to use them, but also, in reference to #354, to show how to create custom readers. (Minor issue of if they are modified, we'll have to be sure to update in both places, unless in the example could do something with [`inspect.getsource`](https://docs.python.org/3/library/inspect.html#inspect.getsource))

BenjaminFraser:
That's no problem at all, and including the Readers within the main code base sounds like a good idea! The only sticking point was including it with pandas as an optional dependency, but I'll look into that, which should hopefully be straightforward enough.

I'll take a look later when I have the chance and put together another PR for those points!

sdhiscocks:
> The only sticking point was including it with pandas as an optional dependency, but I'll look into that, which should hopefully be straightforward enough.

We've done this before by simply raising an error on importing of dependencies.
https://github.com/dstl/Stone-Soup/blob/5276c1b6b541487203806c6fbc8a0547a9ece762/stonesoup/reader/opensky.py#L5-L10

docs/examples/Custom_Pandas_Dataloader.py
```diff
@@ -0,0 +1,280 @@
...
+    """A custom detection reader for DataFrames containing detections.
+
+    DataFrame must have headers with the appropriate fields needed to generate
+    the detection. Detections at the same time are yielded together, and such assume file is in
+    time order.
+
+    Parameters
+    ----------
+    """
+    dataframe: pd.DataFrame = Property(doc="DataFrame containing the ground truth data.")
```
sdhiscocks, line 243:
```suggestion
    dataframe: pd.DataFrame = Property(doc="DataFrame containing the detection data.")
```

docs/examples/Custom_Pandas_Dataloader.py
```diff
@@ -0,0 +1,280 @@
...
+    
+    @BufferedGenerator.generator_method
+    def detections_gen(self):
+        detections = set()
+        previous_time = None
+        for row in self.dataframe.to_dict(orient="records"):
```
sdhiscocks, line 249: Could possibly also group by time field here.

docs/examples/Custom_Pandas_Dataloader.py
```diff
@@ -0,0 +1,280 @@
...
+support data contained within Pandas DataFrames, rather than
+loading directly from a .csv file using :class:`~.CSVGroundTruthReader` or 
+:class:`~.CSVDetectionReader`. 
+
+The benefit is that this allows us to use the versatile data loading
+capabilities of pandas to read from many different data source types 
+as needed, including .csv, JSON, XML, Parquet, HDF5, .txt, .zip and more. 
+The resulting DataFrame can then simply be fed into the defined 
+`DataFrameGroundTruthReader` or `DataFrameDetectionReader` for further processing
+in Stonesoup as required.
```
orosoman-dstl, line 16:
```suggestion
in Stone Soup as required. 
```

docs/examples/Custom_Pandas_Dataloader.py
```diff
@@ -0,0 +1,280 @@
...
+
+from typing import Sequence, Collection
+
+from datetime import datetime, timedelta
+from dateutil.parser import parse
+
+# %%
+# Data Frame Reader
+# ^^^^^^^^^^^^^^^^^
+# Similarly to Stonesoups :class:`~._CSVFrameReader`, we'll define a `_DataFrameReader`
```
orosoman-dstl, line 51:
```suggestion
# Similarly to Stone Soup's :class:`~._CSVFrameReader`, we'll define a `_DataFrameReader`
```

docs/examples/Custom_Pandas_Dataloader.py
```diff
@@ -0,0 +1,280 @@
...
+        else:
+            local_metadata = {field: row[field]
+                              for field in self.metadata_fields
+                              if field in row}
+        return local_metadata
+
+    def _get_time(self, row):
+        if self.time_field_format is not None:
+            time_field_value = datetime.strptime(row[self.time_field], self.time_field_format)
+        elif self.timestamp is True:
```
orosoman-dstl, line 84:
```suggestion
        elif self.timestamp:
```
You don't need the "is True" for a bool.

docs/examples/Custom_Pandas_Dataloader.py
BenjaminFraser: Morning Oliver, that's no problem at all and totally understood - consistency is important. I'll have a look and ensure this is all consistent for any further changes made.

docs/examples/Custom_Pandas_Dataloader.py
```diff
@@ -0,0 +1,280 @@
...
+    dataframe: pd.DataFrame = Property(doc="DataFrame containing the ground truth data.")
+    path_id_field: str = Property(doc='Name of column to be used as path ID')
+
+    @BufferedGenerator.generator_method
+    def groundtruth_paths_gen(self):
+        """ Generator method for providing each row of ground truth data. """
+        groundtruth_dict = {}
+        updated_paths = set()
+        previous_time = None
+        for row in self.dataframe.to_dict(orient="records"):
```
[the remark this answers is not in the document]
BenjaminFraser: I'm sure we could do this easily enough, although I've not included it in the current commit yet though. Before doing this, I could do with confirming the precise details in order to avoid accidentally changing the current logic of the generator.

If we simply order by path_id and time, and then iteratively yield each time and updated_paths (detections for DataFrameDetectionReader), is that doing precisely the same functionality as the current groundtruth_paths_gen (and detections_gen)?

sdhiscocks (approved):

spike-dstl (approved):

PR #708, opened 2022-08-31
Migrate from setup.py to pyproject.toml based build

sdhiscocks:
Pip (and other tools) now consider `setup.py` legacy, and such should migrate to `pyproject.toml` (and associated `setup.cfg` for `setuptools`).

This also removes dependency on `setuptools_scm_git_archive` as this is now integrated into `setuptools_scm >= 7`.


jswright-dstl (approved):

setup.cfg
```diff
@@ -0,0 +1,56 @@
...
+    numpy>=1.17
+    ordered-set
+    pymap3d
+    ruamel.yaml>=0.16.5
+    rtree
+    scipy
+    utm
+
+[options.extras_require]
+dev =
```
hpritchett-dstl, line 32: With #707, we will also need pandas here. But maybe that should be addressed in that PR.

sdhiscocks: Yeah, that PR adds the dependency. This'll cause a merge conflict, so on merge of 707 I'll add pandas onto this list.

hpritchett-dstl (approved):

PR #715, opened 2022-09-14
Add ImmutableMixIn and Freezable classes

erogers-dstl:
For discussion: I wrote these in a separate repo for use in a internal project. 

Are these something of use to the community?

### DIFF
--- a/stonesoup/base.py
+++ b/stonesoup/base.py
@@ -466,3 +466,161 @@ class Base(metaclass=BaseMeta):
     if sys.version_info < (3, 11):
         def __getstate__(self):
             return self.__dict__
+
+
+class ImmutableMeta(BaseMeta):
+    """Metaclass for immutable Stone Soup objects. New classes using this metaclass have all
+    the same Properties as any parent class, but all these properties are set to read-only."""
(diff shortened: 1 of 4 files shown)


etfrogers-dstl:
Also: it is possible this shouldn't be in `base.py`. Would a new file be more appropriate?

stonesoup/base.py
```diff
@@ -457,3 +457,164 @@ def __repr__(self):
...
+    Immutability is inherited: that is, all subclasses of a class which inherits from this class
+    will have *all* their properties readonly, even ones defined by the subclass.
+    """
+    def __eq__(self, other):
+        if self._is_hashable:
+            return (type(self) is type(other)
+                    and all(getattr(self, name) == getattr(other, name)
+                            for name in type(self).properties))
+        else:
+            return self is other
```
sdhiscocks, line 510:
```suggestion
    def __eq__(self, other):
        if self is other:
            return True
        if self._is_hashable:
            return (type(self) is type(other)
                    and all(getattr(self, name) == getattr(other, name)
                            for name in type(self).properties))
        return False
```

sdhiscocks (approved):

PR #717, opened 2022-09-14
Change 2d assignment implementation to SciPy version

sdhiscocks:
SciPy changed from Hungarian to the same algorithm that was provided to us from David Crouse, based on modified Jonker-Volgenant algorithm. It's preferable to use SciPy C implementation, than our Python NumPy implementation.


etfrogers-dstl (approved):
Seems sensible

sdhiscocks:
Should have added, that I tested this on a large cost matrix and got identical results with both implementations.

sdhiscocks:
Should have added, that I tested this on a large cost matrix and got identical results with both implementations.

nperree-dstl (approved):

gawebb-dstl:
It’s worth noting that the implementations are slightly different. This difference doesn’t affect the `GNNWith2DAssignment` class but may affect other functions/classes which use the algorithm 
For non-square cost arrays the numpy implementation explicitly uses ‘-1’ to signify that a non-match while the scipy implementation ignores indices that do not match. They are implicitly absence.
The other difference is the `row_ind` (`row4col` in the numpy implementation) output from the scipy implementation is sorted while the numpy implementation isn’t. 

The following script highlights the difference

```
from stonesoup.dataassociator._assignment import assign2D
from scipy.optimize import linear_sum_assignment
import numpy as np
from numpy.random import default_rng

def numpy_version(C, maximise=True):
    C2 = np.copy(C)
    gain, col4row, row4col = assign2D(C2, maximise)

    if gain.size <= 0:
        return None, None
    else:
        return col4row, row4col

def scipy_version(C, maximise=True):
    C2 = np.copy(C)
    try:
        row4col, col4row = linear_sum_assignment(C2, maximise)
    except ValueError:
        return None, None
    else:
        return col4row, row4col

def compare_versions(C, print_stuff=False):
    n_c4r, n_r4c = numpy_version(C)
    s_c4r, s_r4c = scipy_version(C)

    c4r_match = np.array_equal(n_c4r, s_c4r)
    r4c_match = np.array_equal(n_r4c, s_r4c)

    if print_stuff:
        if c4r_match and r4c_match:
            print("Everything matches")
        else:
            if not c4r_match:
                print("Col4row:", "Numpy answer", n_c4r, "doesn't equal Scipy answer", s_c4r, "for cost matrix size", C.shape)
            if not r4c_match:
                print("Row4col:", "Numpy answer", n_r4c, "doesn't equal Scipy answer", s_r4c, "for cost matrix size", C.shape)
            #print("For Cost Matrix", C)

    return c4r_match and r4c_match

cost_matrix = np.array([[5, 2], [4,11], [16, 3]])
numpy_col4row, numpy_row4col = numpy_version(cost_matrix)
scipy_col_ind, scipy_row_ind = scipy_version(cost_matrix)

print("Cost Matrix:", cost_matrix)
print("Numpy result:", numpy_col4row, numpy_row4col)
print("Scipy result:", scipy_col_ind, scipy_row_ind)
print("Numpy col4row", numpy_col4row, "explicitly ‘-1’ to signify that there is no match in the first row")
# You could get the cost via either of these two methods for the numpy implementation
costs = [cost_matrix[i, j] for i, j in enumerate(numpy_col4row) if j != -1]
costs = [cost_matrix[i, j] for j, i in enumerate(numpy_row4col) if i != -1]

# The scipy version makes it more convenient to get the cost values
costs = cost_matrix[scipy_row_ind, scipy_col_ind]


print()
print("Other random examples")

rng = default_rng()
rints = rng.integers(low=2, high=10, size=10)

print("Square arrays")
for size_of_array in rints:
    C = np.random.rand(size_of_array, size_of_array)
    do_they_match = compare_versions(C, True)


print("Non-Square arrays")
rints2 = rng.integers(low=2, high=10, size=(10,2))
for (i,j) in rints2:
    C = np.random.rand(i, j)
    do_they_match = compare_versions(C, True)


```

PR #719, opened 2022-09-16
Add Multi-model and Rao-Blackwellised versions of particle filter

sdhiscocks:
This is a squash and clean up of commits from pull request #271, adapted to work with changes on the `main` branch.

TODO:
- [x] Vectorise particles
- [x] Remove new resamplers (these shouldn't be required once vectorised, as can use `from_state` method)
- [x] Improve tests
- [ ] ~~Add example~~ (to be done in separate PR)


stonesoup/updater/particle.py
```diff
@@ -221,3 +224,207 @@ def predict_measurement(self, state_prediction, *args, **kwargs):
...
+                mean = model.function(
+                    parent_required_state_space, time_interval=time_interval, noise=False)
+
+                # Input the indices that were removed previously
+                for j in range(len(particle.state_vector)):
+                    if j not in position_mapping[i]:
+                        mean = np.insert(mean, j, 0, axis=0)
+
+                cov_matrices = [measurement_model.covar()
+                                for _ in range(len(model.model_list))]
```
sdhiscocks, line 405: Should this be transition model covariance (or transition model PDF method), rather than the measurement model?

sdhiscocks: I've changed this to the transition model now.

stonesoup/predictor/particle.py
```diff
@@ -91,3 +97,137 @@ def predict(self, prior, *args, **kwargs):
...
+            if j not in position_mapping:
+                new_state_vector = np.insert(new_state_vector, j, 0, axis=0)
+
+        return new_state_vector
+
+
+class RaoBlackwellisedMultiModelPredictor(MultiModelPredictor):
+    """ParticlePredictor class
+
+    An implementation of a Particle Filter predictor.
```
orosoman-dstl, line 188: Is there a reference that could go here?

stonesoup/predictor/tests/test_multimodel.py
```diff
@@ -0,0 +1,140 @@
...
+def test_multi_model():
+
+    # Define time related variables.
+    timestamp = datetime.datetime.now()
+    time_diff = 2  # 2sec.
+    new_timestamp = timestamp + datetime.timedelta(seconds=time_diff)
+
+    # Define prior state.
+    prior_particles = [
+        MultiModelParticle(np.array([[10], [10], [10], [10], [10], [10], [10], [10], [10]]),
```
orosoman-dstl, line 21: Seeing this huge list is making me think maybe it would be worth having a container class for Particle/MultiModelParticle, then perhaps a function that could generate something like this given a few parameters?  I'm not sure how that would work specifically, but this kind of thing definitely contravenes the standard coding advice about avoiding excessive repetition right?  Maybe a change that would be worth a new separate PR.  Cue "Nicely volunteered"!

sdhiscocks: `MultiModelParticleState` has also been added in this PR, which like `ParticleState` lets you just use `(no. dimension, no. particles)` arrays. I could update this test to use it.

orosoman-dstl: Oh cool, that was what I was looking for.  Yeah I reckon that'd be worth doing, as I know I often turn to tests as a way of seeing how the components are intended to be used.  If that is the case

hpritchett-dstl (commented):
I've attempted implementing the Multi-model and Rao-Blackwellised particle filters in the following notebook:
https://gist.github.com/hpritchett-dstl/a42124667bb4fe10e734d5c57c6db3d8
These filters are a great addition to Stone Soup. I only ran into a couple problems (assuming i'm using it correctly!) when making the notebook. Those problems are at the bottom of the notebook, I will also mention them at the relevant points in this PR.  😃

stonesoup/predictor/particle.py
```diff
@@ -91,3 +97,137 @@ def predict(self, prior, *args, **kwargs):
...
+            prior,
+            state_vector=copy.copy(prior.state_vector),
+            parent=prior,
+            particle_list=None,
+            timestamp=timestamp)
+
+        # Change the value of the dynamic value randomly according to the
+        # matrix
+        new_dynamic_models = np.array([
+            np.random.choice(len(self.transition_models), p=model_probabiliy)
```
hpritchett-dstl, line 218: Typo in "model_probabiliy"

hpritchett-dstl, line 218: I came across an issue with the following line of code, causing the error:
"_TypeError: Cannot cast array data from dtype('O') to dtype('float64') according to the rule 'safe'_"
This can be resolved by casting the array as a float, e.g:
```suggestion
            np.random.choice(len(self.transition_models), p=model_probabiliy.astype(float)
```

stonesoup/predictor/particle.py
```diff
@@ -91,3 +97,137 @@ def predict(self, prior, *args, **kwargs):
...
+    transition_model = None
+    transition_models: Sequence[TransitionModel] = Property(
+        doc="Transition models used to for particle transition, selected by model index on "
+            "particle."
+    )
+    transition_matrix: np.ndarray = Property(
+        doc="n-model by n-model transition matrix."
+    )
+    position_mappings: Sequence[Sequence[int]] = Property(
+        doc="Sequence of position mappings associated with each transition model.")
```
hpritchett-dstl, line 116: The term "position_mappings" causes a bit of confusion with me.  For example, if you have a transition model for a 4dim state vector (x, x_v, y, y_v) such as _CombinedLinearGaussianTransitionModel(ConstantVelocity(0.05), ConstantVelocity(0.05))_ you would want your position_mapping to be [0,1,2,3], not an intuitive [0, 2]. Maybe a change of name will make it more obvious? model_mappings?

sdhiscocks: Hopefully improved this in 67ae107

hpritchett-dstl (approved):

jswright-dstl (approved):

PR #721, opened 2022-09-22
Fix plotting error when a track only has a single state

sdhiscocks:
This was caused by squeeze removing extra dimension when only a single state was present. Using concatenate along correct dimension avoids this.

### DIFF
--- a/stonesoup/plotter.py
+++ b/stonesoup/plotter.py
@@ -310,9 +310,10 @@ class Plotter(_Plotter):
                 else:
                     not_update_indexes.append(n)
 
-            data = np.array(
+            data = np.concatenate(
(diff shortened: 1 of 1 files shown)


gawebb-dstl (approved):
Looks good to me. I tested it with this snippet of code
```
from stonesoup.plotter import Plotter
from stonesoup.types.state import State
import matplotlib.pyplot as plt

t1 = Track([State([1,2]),
            State([2,4])])
t2 = Track([State([1,4])])

plotter = Plotter()
plotter.plot_tracks(t1, mapping=[0,1])
try:
    plotter.plot_tracks(t2, mapping=[0, 1])
    print("Pass")
except IndexError:
    print("Fail")

plt.show()
```

jswright-dstl (approved):

PR #724, opened 2022-09-23
Fix bug with pre-fetching noise in cases of only a single target

sdhiscocks:


jswright-dstl (approved):

gawebb-dstl (changes requested):
In the scenario `len(ground_truths) == 1 and noise == True` (which happens in the Disjoint_Tracking_Classification.py example) 
Then `measurement_noise = True`
Which doesn't cause an error on the line `measurement_vector += measurement_noise` 
It just adds 1 to the `measurement_vector` 

Would this be a suitable solution:
```
if noise is True:
    # Pre-fetch noise values when more than one target
    if len(ground_truths) > 1:
        noise_vectors_iter = iter(measurement_model.rvs(len(ground_truths), **kwargs))
    else:
        noise_vectors_iter = iter([measurement_model.rvs(**kwargs)])
```

sdhiscocks:
> In the scenario `len(ground_truths) == 1 and noise == True` (which happens in the Disjoint_Tracking_Classification.py example) Then `measurement_noise = True` Which doesn't cause an error on the line `measurement_vector += measurement_noise` It just adds 1 to the `measurement_vector`
> 
> Would this be a suitable solution:
> 
> ```
> if noise is True:
>     # Pre-fetch noise values when more than one target
>     if len(ground_truths) > 1:
>         noise_vectors_iter = iter(measurement_model.rvs(len(ground_truths), **kwargs))
>     else:
>         noise_vectors_iter = iter([measurement_model.rvs(**kwargs)])
> ```

Thanks @gawebb-dstl, I hadn't spotted that. I've updated it for sensors where this is an issue.

gawebb-dstl:
@sdhiscocks I noticed you were having to rewrite the same code in many different places, so I tried to abstract that code to one class see #732

PR #727, opened 2022-09-23
Allow default values for MeasurementInitiators

gawebb-dstl:
Add default value to prior_state for `SimpleMeasurementInitiator` & `MultiMeasurementInitiator`

- The prior state is overwritten with 0s in the function so it makes little sense to require a value for it when a `StateVector` and `CovarianceMatrix` can initalised with zeros

- Slight breaking change to `MultiMeasurementInitiator` as the order of arguments has changed. No effect when initialising the object with keyword arguments

Add option to skip detections in `MultiMeasurementInitiator` with a non reversible measurement model

### DIFF
--- a/docs/demos/OpenSky_Demo.py
+++ b/docs/demos/OpenSky_Demo.py
@@ -195,7 +195,7 @@ from stonesoup.initiator.simple import MultiMeasurementInitiator
 from stonesoup.types.state import GaussianState
 
 initiator = MultiMeasurementInitiator(
-    GaussianState(
+    prior_state=GaussianState(
         np.array([[0], [0], [0], [0], [0], [0]]),   # Prior State
         np.diag([15**2, 100**2, 15**2, 100**2, 15**2, 20**2])),
     measurement_model=None,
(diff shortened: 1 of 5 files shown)


sdhiscocks:
> * The prior state is overwritten with 0s in the function so it makes little sense to require a value for it when a `StateVector` and `CovarianceMatrix` can initalised with zeros

So this currently only overwrites elements of the prior that are mapped to the measurement space. This could result in having some elements of the covariance default to zero, which would cause cause issues. I don't think this is suitable default behaviour.

> Add option to skip detections in `MultiMeasurementInitiator` with a non reversible measurement model

The multi-measurement initiator doesn't have any calls to the inverse function of the measurement model, so not sure why this would be required?

gawebb-dstl:
> > Add option to skip detections in `MultiMeasurementInitiator` with a non reversible measurement model
> 
> The multi-measurement initiator doesn't have any calls to the inverse function of the measurement model, so not sure why this would be required?

This additional option was used to filter the detections going into the initiator. 
Measurements with non-reversible measurement models generally don't have sufficient co-ordinate information to map to a single position/state. Therefore they're more likely to associate to clutter and form clutter tracks. 
This was an easy quick fix and may be more appropriate as a wrapper for the initiator instead

gawebb-dstl:
I'm not sure if `SimpleMeasurementInitiatorWithCovar` adds much, let me know what you think.

sdhiscocks:
Yeah, I'm not sure `SimpleMeasurementInitiatorWithCovar` is really needed. But certainly the filtered detections would be useful.

sdhiscocks:
@gawebb-dstl I'll close this one for now, but the filtered detections for initiators would be useful if you could open another PR for this.

PR #728, opened 2022-09-23
Make StateVectors Hashable

gawebb-dstl:
The most noticable difference would be the change to equals

A small code snippet to display the differences
```
sv1 = StateVector([1, 2])
sv2 = StateVector([2, 3])
sv3 = StateVector([1, 2])

print(sv1 == sv2)
print(sv1 == sv3)
print(sv1 is sv3)

x = {sv1: 1, sv2: 2, sv3: 3}
print(x[sv1], x[sv2], x[sv3])
```

### DIFF
--- a/stonesoup/types/array.py
+++ b/stonesoup/types/array.py
@@ -47,6 +47,12 @@ class Matrix(np.ndarray):
         else:
             return self._cast(result)
 
+    def __hash__(self):
+        return hash(self.tobytes())
+
+    def __eq__(self, other):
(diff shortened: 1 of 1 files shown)


sdhiscocks:
One requirement of hashable objects in Python, is the hash must never change in it's lifetime. Not easily done with NumPy array. It may be safer in cases where you need hashable state vector, to convert it to a tuple representation.

gawebb-dstl:
Good point, which I hadn't considered

PR #734, opened 2022-10-07
Add Accumulated State Densities Filter

sdhiscocks:
This is an update of #309 by @jogehl to latest version of Stone Soup.

Changes to core code are minor tidy up for readability, some minor changes based on comments from #309, and moving to separate sub-module from Kalman.

Tutorial has also been updated to take advantage of newer animation and plotly features.


sdhiscocks:
Code coverage highlighted that `_transition_function` wasn't used in predictor, so I did a little refactor to make use of the parent `_transition_function` and removed custom `pred` function. Extend and other variants in theory should be easier to add in future.

stonesoup/predictor/asd.py
```diff
@@ -0,0 +1,275 @@
...
+            p_pred[(t_index+1) * ndim:, (t_index+1) * ndim:] = P_right_lower
+
+            # get all timestamps which has to be recalculated beginning
+            # with the newest one
+            timestamps = sorted(prior.timestamps + [timestamp], reverse=True)
+            timestamps_to_recalculate = [ts for ts in timestamps if ts >= timestamp]
+            covars = \
+                [prior.multi_covar[i * ndim:(i+1) * ndim, i * ndim:(i+1) * ndim]
+                 for i in range(t_index)]
+            covars.append(p_pred_m)
```
sdhiscocks, line 206: I think this should be `p_pred_m_cr` rather than `p_pred_m`. Changing this, it seems to give better and more stable results, whereas without changing it, the estimates looking wrong.

sdhiscocks: I've changed this in ecfd1a6, which also includes changing the covariances stored in correlation matrices, as I believe this were the wrong way around.

jogehl: Sorry for the late answer. You are right. This was an error in my implementation.

sdhiscocks:
@jogehl Thanks again for your contribution, and sorry about significant delay getting to your original pull request. I wasn't sure if you were still active on GitHub.

I wanted to add you to the copyright authors list, but wasn't sure whether to put your name or company/institution that you are/were based at when you did this work?

jogehl:
@sdhiscocks if you like you can put the Fraunhofer FKIE to the contributors' list.

nperree-dstl (approved):
This all works/makes sense as far as I can tell. Could possibly do with a bit more explanation in the tutorial as to how the filter works but otherwise I think it looks good!

jswright-dstl (approved):

PR #738, opened 2022-10-12
Fixed TypeError being thrown on certain Gaussian states in the Plotter.plot_tracks function

gawebb-dstl:
For some Gaussian states the uncertainity can't be plotted by the `Plotter` class. You can see if you try running the following code
```
from stonesoup.plotter import Plotter
from stonesoup.types.state import GaussianState, StateVector

track = Track([GaussianState(state_vector=[0, 0], covar=[[10, -1], [1, 10]])])
Plotter().plot_tracks(track, mapping=[0, 1], uncertainty=True)
# TypeError: ufunc 'arctan2' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

This is caused by complex eignevalues or eigenvectors from the uncertainity matrix. I've put in a check for these in the `Plotter` class. States which cause complex eignevalues or eigenvectors are now skipped in the plot_tracks function and a warning is raised. A test has been created for this behaviour

### DIFF
--- a/stonesoup/plotter.py
+++ b/stonesoup/plotter.py
@@ -343,6 +343,10 @@ class Plotter(_Plotter):
                     HH = np.eye(track.ndim)[mapping, :]  # Get position mapping matrix
                     for state in track:
                         w, v = np.linalg.eig(HH @ state.covar @ HH.T)
+                        if np.iscomplexobj(w) or np.iscomplexobj(v):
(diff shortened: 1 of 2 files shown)


gawebb-dstl:
I think the same error is present in the `Plotterly` class but haven't tested it

sdhiscocks (approved):

jswright-dstl (approved):

PR #742, opened 2022-10-20
Progress Trackers Manually

gawebb-dstl:
All current trackers require a detection feeder. It can be useful to progress a tracker manually without a detection feeder (see Multi-Sensor Fusion: Covariance Intersection Using Tracks as Measurements example). 

All current trackers and start their `__next__` function with:
```
time, detections = next(self.detector_iter)
# Some tracking logic
```

I've abstracted this line to the new `TrackerWithDetector` class and added an `update_tracker` function which takes the time and detections from the detection feeder as an input. With this being a separate function, you can access it directly and bypass the detection feeder which may be easier in instances

### DIFF
--- a/docs/examples/Track2Track_Fusion_Example.py
+++ b/docs/examples/Track2Track_Fusion_Example.py
@@ -550,7 +550,7 @@ class DummyDetector(DetectionReader):
 
 # %%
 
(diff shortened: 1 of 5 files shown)


gawebb-dstl:
See PR #779 for an alternative implementation

PR #743, opened 2022-10-20
Modified dotproduct to work with StateVectors 

jmbarr:
Dotproduct function previously only worked with `StateVector`. This is designed to make it work with `StateVectors` and return an numpy.array of dimension 1x the number of `StateVector`s.

### DIFF
--- a/stonesoup/functions/__init__.py
+++ b/stonesoup/functions/__init__.py
@@ -633,7 +633,7 @@ def dotproduct(a, b):
 
     def _dotproductvectors(v1, v2):
         # This operates on a StateVector
-        oout=0
+        oout = 0
         for a_i, b_i in zip(v1, v2):
             oout += a_i*b_i
         return oout
--- a/stonesoup/functions/__init__.py
+++ b/stonesoup/functions/__init__.py
@@ -614,37 +614,43 @@ def build_rotation_matrix(angle_vector: np.ndarray):
 
 
 def dotproduct(a, b):
-    r"""Returns the dot (or scalar) product of two StateVectors.
+    r"""Returns the dot (or scalar) product of two StateVectors or two sets of StateVectors.
 
-    The result for vectors of length :math:`n` is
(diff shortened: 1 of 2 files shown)


stonesoup/functions/__init__.py
```diff
@@ -614,37 +614,43 @@ def build_rotation_matrix(angle_vector: np.ndarray):
...
+    # Decide whether this is a StateVector or a StateVectors
+    if type(a) is StateVector and type(b) is StateVector:
+        return _dotproductvectors(a, b)
+    elif type(a) is StateVectors and type(b) is StateVectors:
+        out = []
+        for aa, bb in zip(a, b):
+            out.append(_dotproductvectors(aa, bb))
+        return np.reshape(out, np.shape(np.atleast_2d(a[0, :])))
+    else:
+        raise ValueError("Inputs must be `StateVector` or `StateVectors` and of the same type")
```
sdhiscocks, line 653: The below should be equivalent using `np.sum`. Not sure 100% what shape you want the output to be, as currently `(1, ndim)` in the implementation above, but `(ndim, )` below...but adding `np.atleast_2d(...)` below would make it the same.

```suggestion
    if np.shape(a) != np.shape(b):
        raise ValueError("Inputs must be (a collection of) column vectors of the same dimension")

    # Decide whether this is a StateVector or a StateVectors
    if type(a) is StateVector and type(b) is StateVector:
        return np.sum(a*b)
    elif type(a) is StateVectors and type(b) is StateVectors:
        return np.asarray(np.sum(a*b, axis=0))
    else:
        raise ValueError("Inputs must be `StateVector` or `StateVectors` and of the same type")
```

jmbarr: Array out should be shape [1, ndim] to match StateVectors dimension.

nperree-dstl (approved):

sdhiscocks (approved):

PR #744, opened 2022-10-27
Add logpdf method to models, and implement in particle filter

sdhiscocks:
This avoids a number a nugatory calculations switching in/out of log space with Probability class.

Also modified mean/covar calculation in particle state to use float array to avoid similar issue caused by Probability class forcing use of object arrays.

### DIFF
--- a/stonesoup/models/base.py
+++ b/stonesoup/models/base.py
@@ -102,6 +102,24 @@ class Model(Base):
         """
         raise NotImplementedError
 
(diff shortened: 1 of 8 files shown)


hpritchett-dstl:
This looks good to me!

The change made to the calculation of the mean in the particle state resulted in a significant runtime improvement, with an example particle filter tracker going from 34.1s runtime to 11.8s, while reproducing identical results.

particle_state_mean_change

hpritchett-dstl (approved):

jswright-dstl (approved):

PR #746, opened 2022-11-09
Add average uncertainty method to UncertaintyRewardFunction

mharris-dstl:


stonesoup/sensormanager/reward.py
```diff
@@ -57,6 +57,7 @@ class UncertaintyRewardFunction(RewardFunction):
     predictor: KalmanPredictor = Property(doc="Predictor used to predict the track to a new state")
     updater: ExtendedKalmanUpdater = Property(doc="Updater used to update "
                                                   "the track to the new state.")
+    method: str = Property(default="sum", doc="Determines the type of uncertainty reward - use 'sum' or 'average'.")
```
sdhiscocks, line 60: Should probably avoid a "magic string" here:

- One option is to have an [`enum`](https://docs.python.org/3/library/enum.html);
- another is to have Boolean option of `average` instead, defaulting to `False`; 
- or maybe have this as a `Callable` defaulting to `np.sum`, but suggest `np.average` as an alternative in doc string (leaves option there other functions that then could be used e.g. `np.median` or custom lambda function).

sdhiscocks, line 60: This line is too long also, failing *flake8* tests.

sdhiscocks (changes requested):

nperree-dstl (approved):

sdhiscocks (approved):

sdhiscocks (changes requested):
Could you add a test to cover `method_sum` is `True`.

PR #751, opened 2022-11-28
TimeBasedPlotter default timestamps and Notebook animation docs

sdhiscocks:
- Use all timestamps by default for TimeBasedPlotter
  This no longer requires specifying the time stamps, as they default to
those from data being plotted
- Simplify docs for Notebook use of animations in TimeBasedPlotter
   Also ignore example gif file that is saved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ htmlcov/
 
 # pytest
 .pytest_cache/
+
+#Animation plotting example
+docs/examples/example_animation.gif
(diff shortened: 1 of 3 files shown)


jswright-dstl (approved):

gawebb-dstl (approved):
Its a small useful PR and the code looks good to me. I've commented one optional small change so that the figure appears when being ran from the command line/ pycharm

Issue #749, opened 2022-11-22
Rotation matrix calculation


edwheelhouse-dstl:
Stonesoup forms the rotation matrix in the wrong order in stonesoup.functions.__init__.py / build_rotation_matrix()

I've created a branch 'rotation_matrix_fix' with a potential solution ready for a pull request but it gets complicated.

In the new branch I've corrected the calculation and the tests which make the same mistake. I've written some more test cases and added them to stonesoup.models.measurement.tests.test_models.position_measurement_sets. The old code fails some of the new test cases and the new code passes them all.

The new tests highlighted another issue: numpy.isclose() doesn't work reliably with the Bearing class: two Bearings of -pi fail numpy.isclose()

It is because of how Bearing's parent, Angle, overrides __init__() so some stonesoup tests fail when they shouldn't. you could use a work around in the tests to compare the Bearing's _values instead of the Bearings themselves and that allows the tests to pass. However I think it would be better to deal with the cause. Currently Angle overrides __abs__() by creating a new Bearing and Angle.__init__() returns mod_angle(value) which, if value=pi, returns -pi. Therefore abs(Bearing(np.pi)) and abs(Bearing(-np.pi)) both return a negative number, which causes np.isclose() to return false and is probably not what was intended. Note abs(Bearing(anything other than pi or -pi)) returns a positive number as intended.

I think it might be better to alter stonesoup.functions.__init__.py/mod_bearing() so that it returns a value from the range that includes +pi rather than -pi as it does now. Doing that currently causes failure of stonesoup.types.test.test_angle.py/test_average() but that looks easy enough to fix by testing val = np.pi instead of -np.pi. However, I'm in danger of getting lost in this rabbit hole and going beyond what I really understand!

PR #762, opened 2023-01-19
Rotation matrix fix

edwheelhouse-dstl:
The rotation matrix is currently formed backwards as rotz(theta_z) @ roty(theta_y) @ rotx(theta_x). It needs to be in reverse order: x, y, z. This is because matrix @ operations are carried out from right to left and we need to rotate about z first, because only then can we know which way to rotate about y. This branch implements that and adds some test cases to prove it (some of which fail with the previous code). 

This does lead to a problem with Bearing, as discussed in Issue #749.

### DIFF
--- a/stonesoup/models/measurement/tests/test_models.py
+++ b/stonesoup/models/measurement/tests/test_models.py
@@ -339,9 +339,9 @@ position_measurement_sets = [((0, 1, 0, 0, 0, 0), (1, 0, 0, 0, 0, 0),
                               (-np.pi / 4, 0, np.sqrt(200), -1 / np.sqrt(2))),
                              ((0, 0, 0, 0, 0, 1), (0, 0, 0, 0, 10, 0),
                               (0, 0, 10, -1)),    # original tests up to here
-                             ((0, 2, 0, 0, 0, 1),(1, 2, 0, 0, 1, 1),
+                             ((0, 2, 0, 0, 0, 1), (1, 2, 0, 0, 1, 1),
                               (np.pi/4-np.arctan(0.5), 0, np.sqrt(2), 0)),
-                             ((0, 2, 0, 0, 0, -1),(1, 2, 0, 0, 1, -1),
+                             ((0, 2, 0, 0, 0, -1), (1, 2, 0, 0, 1, -1),
                               (np.pi/4+np.arctan(0.5), 0, np.sqrt(2), 0)),
                              ((0, -2, 0, 0, 0, 1), (1, -2, 0, 0, 1, 1),
                               (np.pi/4+np.arctan(0.5), np.pi, np.sqrt(2), 0)),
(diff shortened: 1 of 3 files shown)


stonesoup/platform/tests/test_platform_base.py
```diff
@@ -649,6 +649,11 @@ def test_range_and_angles_to_other(first_state, second_state, expected_measureme
                                transition_model=None)
 
     range_, azimuth, elevation = platform1.range_and_angles_to_other(platform2)
+    # Allow azimuth of -pi to equal pi, to avoid unnecessary assert failures
+    if azimuth == -np.pi:
+        azimuth = np.pi
+    if expected_measurement[1] == -np.pi:
```
etfrogers-dstl, line 655: Expected measurement is never `pi`, so lines 655-656 are not needed. Could potentially be removed.

stonesoup/models/measurement/tests/test_models.py
```diff
@@ -338,7 +338,77 @@ def fun(x):
                              ((0, 1, 0, 0, 0, 0), (10, 0, 0, 0, -10, 0),
                               (-np.pi / 4, 0, np.sqrt(200), -1 / np.sqrt(2))),
                              ((0, 0, 0, 0, 0, 1), (0, 0, 0, 0, 10, 0),
-                              (0, 0, 10, -1)),
+                              (0, 0, 10, -1)),    # original tests up to here
```
etfrogers-dstl, line 341: There are quite a lot of flake8 errors in this block. (long lines and whitespace errors)

etfrogers-dstl (commented):
Looks good overall. One suggested change (debatable) and a few flake8 errors to fix.

stonesoup/platform/tests/test_platform_base.py
edwheelhouse-dstl: Expected measurement is currently pi a few times, but not -pi. These lines are pre-emptive defence against someone writing new tests using -pi, since there doesn't seem to be a clear convention as to which to use I'd keep them. Unless pre-emptive defence = confusing clutter...?

stonesoup/models/measurement/tests/test_models.py
edwheelhouse-dstl: My mistake - forgot to check with flake8. All fixed.

stonesoup/platform/tests/test_platform_base.py
etfrogers-dstl: > Expected measurement is currently pi a few times, but not -pi. These lines are pre-emptive defence against someone writing new tests using -pi, since there doesn't seem to be a clear convention as to which to use I'd keep them. Unless pre-emptive defence = confusing clutter...?

Fair enough: agree with the argument.

etfrogers-dstl (approved):

etfrogers-dstl:
Note that the test failures on this PR will be fixed (I think) if #763 is applied first

sdhiscocks (approved):

PR #763, opened 2023-01-20
Fix Angle.__abs__ such that it always returns a positive value

etfrogers-dstl:
Address part of issue #749 that deals with `Bearing`s not playing nicely with `numpy.isclose`, which is due to `abs(Bearing(-pi))` returning a negative value.

### DIFF
--- a/stonesoup/types/angle.py
+++ b/stonesoup/types/angle.py
@@ -79,7 +79,24 @@ class Angle(Real):
         return self._value != other
 
     def __abs__(self):
-        return self.__class__(abs(self._value))
+        abs_val = self.__class__(abs(self._value))
+        if abs_val._value < 0:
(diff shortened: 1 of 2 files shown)


sdhiscocks (approved):

etfrogers-dstl:
Note that (as per the comment in the source file), there is still the oddity that `abs(Bearing(-pi)) != Bearing(+pi))` or equivalently that `abs(Bearing(-pi)) != Bearing(abs(-pi)))`

This is because `abs(Bearing(-pi)) == pi` (because `abs` forces it positive) and `Bearing(abs(-pi)) == -pi` (because the bearing modulo operation pushes `pi` to be stored as `-pi`).

edwheelhouse-dstl (approved):
Fine. The oddity is annoying, but hopefully it won't come up often!

PR #764, opened 2023-01-23
Add track stitching class and example.

spike-dstl:
Co-authored-by: Scott Florence <sflorence@dstl.gov.uk>

Pull request to add track stitching example and 'Stitcher' class to main stone soup repository.


sdhiscocks:
Code coverage could be improved. It looks like the *backwards* methods aren't tested currently.

stonesoup/tests/test_stitcher.py
```diff
@@ -0,0 +1,156 @@
...
+from stonesoup.dataassociator.neighbour import GNNWith2DAssignment
+from stonesoup.deleter.error import CovarianceBasedDeleter
+from stonesoup.deleter.multi import CompositeDeleter
+from stonesoup.deleter.time import UpdateTimeStepsDeleter
+from stonesoup.initiator.simple import MultiMeasurementInitiator
+from stonesoup.types.groundtruth import GroundTruthPath, GroundTruthState
+from stonesoup.stitcher import TrackStitcher
+
+start_time = datetime.now().replace(second=0, microsecond=0)
+np.random.seed(100)
```
sdhiscocks, line 22: Can't really use seed in test (without resetting it again). Probably best to use seed in models, and also capture example starting points from this seed for now.

docs/examples/Track_Stitcher_Example.py
```diff
@@ -0,0 +1,585 @@
...
+metric_manager.add_data(truths, set(all_tracks))
+
+plt.rcParams["figure.figsize"] = (10, 8)
+metrics = metric_manager.generate_metrics()
+
+siap_averages = {metrics.get(metric) for metric in metrics
+                 if metric.startswith("SIAP") and not metric.endswith(" at times")}
+siap_time_based = {metrics.get(metric) for metric in metrics if metric.endswith(' at times')}
+
+_ = SIAPTableGenerator(siap_averages).compute_metric()
```
sdhiscocks, line 585: Would be good to compare before and after, but I can see issue is that stitched tracks aren't interpolated. Probably should add that as separate component that could be used.

stonesoup/stitcher.py
```diff
@@ -0,0 +1,206 @@
+from datetime import timedelta
+from stonesoup.types.track import Track
+from stonesoup.types.detection import Detection
+from stonesoup.models.measurement.linear import LinearGaussian
+from copy import deepcopy
+from scipy.optimize import linear_sum_assignment
```
sdhiscocks, line 6:
```suggestion
from collections import defaultdict
from copy import deepcopy
from datetime import timedelta

from scipy.optimize import linear_sum_assignment

from .base import Base, Property
from .hypothesiser import Hypothesiser
from .models.measurement.linear import LinearGaussian
from .types.detection import Detection
from .types.track import Track
```

docs/examples/Track_Stitcher_Example.py
```diff
@@ -0,0 +1,585 @@
...
+        tracks = set(tracks)
+        for ids in combo:
+            x = []
+            for a in ids:
+                track = self.__get_track(a, tracks)
+                x = x + track.states
+                tracks.remove(track)
+            tracks.add(Track(x))
+
+        return tracks
```
sdhiscocks, line 478: Replace with just importing the new class.

sdhiscocks (changes requested):

stonesoup/tests/test_stitcher.py
```diff
@@ -0,0 +1,153 @@
...
+                associated_measurements.add(hypothesis.measurement)
+            else:
+                track.append(hypothesis.prediction)
+        del_tracks = deleter.delete_tracks(tracks)
+        tracks -= del_tracks
+        tracks |= initiator.initiate(measurements - associated_measurements,
+                                     start_time + timedelta(seconds=n))
+        historic_tracks |= del_tracks
+    historic_tracks |= tracks
+    return historic_tracks
```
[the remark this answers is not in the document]
sdhiscocks: Could just use the standard `MultiTargetTracker`

spike-dstl:
> Code coverage could be improved. It looks like the _backwards_ methods aren't tested currently.

stonesoup/stitcher.py
```diff
@@ -0,0 +1,229 @@
...
+    forward_hypothesiser: Hypothesiser = Property(
+        doc="Forward predicting hypothesiser.", default=None)
+    backward_hypothesiser: Hypothesiser = Property(
+        doc="Backward predicting hypothesiser.", default=None)
+    search_window: timedelta = Property(default=timedelta(seconds=30))
+
+    @staticmethod
+    @lru_cache()
+    def _extract_detection(track, backward=False):
+        state = track[0] if not backward else track[-1]
```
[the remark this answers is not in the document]
spike-dstl: This gives a different result than wanted: 
The original method gives: 
backward = False -> state = track[0]
backward = True -> state = track[-1]

The new method would give: 
backward = False -> state = track[-1]
backward = True -> state = track[0]

I think your point is correct though, and the below change should be correct?
```suggestion
    def _extract_detection(track, backward=False):
        state = track[-1] if not backward else track[0]
```

sdhiscocks (approved):

PR #769, opened 2023-02-14
Do not draw line between not_update states.

mattbrown11:
When plotting tracks, the previous behavior would draw a line between all of the states that weren't of type Update.


sdhiscocks (approved):
Thanks @mattbrown11. I'm surprised no one noticed this yet, but suspect as we changed most tutorial and example to use plotly just before this bug was introduced.

jswright-dstl (approved):

PR #776, opened 2023-03-02
Migrate to GitHub Actions

sdhiscocks:
This is a proposed change to migrate to GitHub Actions. The closer integration has some benefits over using CircleCI, such as easier linking and highlighting lines of code with flake8 errors, or which parts of the documentation have formatting errors. Having everything in one place and avoiding logging into another 3rd party service makes the checks easier for contributors to access.

One disadvantage is the documentation artefacts have to be downloaded, rather than viewed online. To consider is whether to keep CircleCI for this?

~~One minor issue, which I do not know the cause of, is GitHub Actions images (or possible a dependency) seems to have underflow errors (e.g. very small probabilities in things like particle filter going to zero), explicitly raised as an error, rather than ignored that seems to be default most everywhere else. Therefore there is an additional commit that explicitly sets underflow errors to be ignored in relevant places.~~


jswright-dstl (approved):
Looks good. Regarding the docs, could we host the built docs on a readthedocs subdomain or something similar to save having to sdownload them/keep CircleCI just for that use case?

sdhiscocks:
> Looks good. Regarding the docs, could we host the built docs on a readthedocs subdomain or something similar to save having to sdownload them/keep CircleCI just for that use case?

Looks like that is an [option](https://docs.readthedocs.io/en/stable/pull-requests.html). I'll look into it some more.

Also, on the issue with underflow, I've seen the issue elsewhere that on GitHub Actions, so I think it's a dependency that may be changing NumPy global default error state.

sdhiscocks:
> Also, on the issue with underflow, I've seen the issue elsewhere that on GitHub Actions, so I think it's a dependency that may be changing NumPy global default error state.

Fixed just an hour ago https://github.com/geospace-code/pymap3d/issues/79

PR #786, opened 2023-03-31
Modify SystematicResampler to enable down/upsampling

timothy-glover:
Added a modification to SystematicResampler to allow an optional argument "nparts" to be added to the resample method of the class. If this is not defined or equal to the length of the input ParticleState object, the resampler will function in the same way. If it is larger/smaller than the length of the input ParticleState object, the resampler will down/up sample to return the desired output.

### DIFF
--- a/stonesoup/resampler/particle.py
+++ b/stonesoup/resampler/particle.py
@@ -8,13 +8,15 @@ from ..types.state import ParticleState
 
 class SystematicResampler(Resampler):
 
-    def resample(self, particles):
+    def resample(self, particles, nparts=None):
         """Resample the particles
 
         Parameters
         ----------
         particles : :class:`~.ParticleState` or list of :class:`~.Particle`
             The particles or particle state to be resampled according to their weights
+        nparts : int
+            The number of particles to be returned from resampling
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


stonesoup/resampler/particle.py
```diff
@@ -35,15 +39,16 @@ def resample(self, particles):
...
 
         # Cycle through the cumulative distribution and copy the particle
         # that pushed the score over the current value
-        u_j = u_i + (1 / n_particles) * np.arange(n_particles)
+        u_j = u_i + (1 / nparts) * np.arange(nparts)
         index = weight_order[np.searchsorted(cdf, np.log(u_j))]
 
-        new_particles = particles[index]
-        new_particles.weight = np.full((n_particles, ), weight)
+        new_particles = particles[np.array(index)]
```
sdhiscocks, line 49: Not sure why `np.array` is added here? Is there case where `np.searchsorted` returns an `int`? If so, `np.atleast_1d` may be better to be clearer.

timothy-glover: A leftover fix for something I removed I think. It has been changed back to `particles[index]` now.

jswright-dstl (approved):

stonesoup/resampler/tests/test_particle.py
```diff
@@ -45,6 +45,17 @@ def test_systematic_even():
...
+    particles = [Particle(np.array([[i]]), weight=1/40) for i in range(40)]
+
+    resampler = SystematicResampler()
+
+    new_particles = resampler.resample(particles, 20)
+
+    # Check that the resampler downsamples the particles correctly
+    assert len(new_particles) == 20
+
+
```
sdhiscocks, line 58: Would be good to add a test for upsampling as well.

Could even parameterise this test for both up and downsampling

timothy-glover: The latest commit includes an upsampling test

PR #795, opened 2023-04-25
Plotting Improvements

gawebb-dstl:
- Added option to not convert detections when plotting 
- Added option to plot_tracks that all the tracks should have the same colour 
- Added option to specify dimension of plot with an int
- Added documentation


mharris-dstl (approved):
Looks good. Appreciate allowing the dimension number to be an int and documentation addition.

gawebb-dstl:
I've added the suggested changes from @rcgorman-dstl

PR #796, opened 2023-04-26
Polar Plotting

gawebb-dstl:
Create polar plots using plotly like [this](https://plotly.com/python/polar-chart/)

### DIFF
--- a/docs/examples/Polar_Plotting.py
+++ b/docs/examples/Polar_Plotting.py
@@ -12,7 +12,7 @@ angular one. Angular ground truth is created using measurements without noise. D
 """
 
 # %%
-# First some general imports and setup
+# First, include some standard imports and initialise the start time:
 from datetime import datetime
 from datetime import timedelta
 
@@ -34,19 +34,20 @@ start_time = datetime(2023, 1, 1)
 # %%
 # Generate Cartesian State Space Data
 # -----------------------------------
+# Two targets are created:
+#  #. Target 1 moves in a ‘C’ shape. First it moves west (negative x), then it starts a slow,
+#     long 180-degree turn moving south, until it is moving west (positive x).
+#  #. Target 2 moves in a straight line from north to south.
 
 # Create manoeuvre behaviours and durations for our moving platform
 straight_level = CombinedLinearGaussianTransitionModel(
     [ConstantVelocity(0.), ConstantVelocity(0.)])
 
-# Configure the aircraft turn behaviour
+# Configure target 1's turn behaviour
 turn_noise_diff_coeffs = np.array([0., 0.])
-turn_rate = np.deg2rad(5)  # specified in radians per seconds...
+turn_rate = np.deg2rad(5)  # specified in radians per second
 turn_model = KnownTurnRate(turn_noise_diff_coeffs=turn_noise_diff_coeffs, turn_rate=turn_rate)
-
-# Configure turn model to maintain current altitude
-turning = CombinedLinearGaussianTransitionModel(
-    [turn_model])
+turning = CombinedLinearGaussianTransitionModel([turn_model])
 
 manoeuvre_list = [straight_level, turning, straight_level]
 manoeuvre_times = [timedelta(seconds=12),
@@ -82,7 +83,7 @@ for t in timesteps:
 # %%
 # Display ground truth in Cartesian state space using the standard
 # :class:`~.Plotterly` plotting class:
-plotter_xy = Plotterly(title="God's Eye View of Targets")
+plotter_xy = Plotterly(title="Bird's Eye View of Targets")
 plotter_xy.plot_ground_truths(target_1, mapping=[0, 2])
 plotter_xy.plot_ground_truths(target_2, mapping=[0, 2])
 plotter_xy.fig
@@ -97,7 +98,13 @@ plotter_xy.fig
 # Create sensor:
 sensor = RadarBearingRange(ndim_state=4,
                            position_mapping=[0, 2],
-                           noise_covar=np.diag([np.radians(0.06), 25]))
+                           noise_covar=np.diag([np.radians(0.06), 50]),
+                           position=StateVector([0, 0]))
+
+# %%
+# Plot sensor's location on XY plot
+plotter_xy.plot_sensors({sensor}, mapping=[0, 1])
+plotter_xy.fig
 
 # %%
 # Measure each ground truth individually and without noise to create ground truth paths:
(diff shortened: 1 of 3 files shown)


mharris-dstl:
I've pushed some grammar changes (thought it was easier for everyone if I add them myself and put them in as one, but have some other thoughts for the example. 

- It needs a cover image putting in.
- It probably needs to show tracks being plotted, since currently is is only ground truth and detections but there is functionality in the source code for track plotting.
- The polar plots themselves could do with some more description in the titles. While polar plots look really cool, I personally have to concentrate quite hard to understand the information they're showing. Better descriptions would help a lot with that.
- The scenario has 2 targets in 3D moving in all 3 dimensions, but the static plot only shows 2 dimensions.
- Could potentially get a 2D animated plotter in there as well? I tried to add this but it hasn't got functionality to automatically plot moving platforms. But an animation of some sort would give a lot of intuition as to how the targets are moving.

mharris-dstl (changes requested):
I've pushed some grammar changes (thought it was easier for everyone if I add them myself and put them in as one) but have some other thoughts for the example. 

- It needs a cover image putting in.
- It probably needs to show tracks being plotted, since currently the example shows only ground truth and detections but there is functionality in the source code for track plotting.
- The polar plots themselves could do with some more description in the titles. While polar plots look really cool, I personally have to concentrate quite hard to understand the information they're showing. Better descriptions would help a lot with that.
- The scenario has 2 targets in 3D moving in all 3 dimensions, but the static plot only shows 2 dimensions.
- Could potentially get a 2D animated plotter in there as well? I tried to add this but it hasn't got functionality to automatically plot moving platforms. But an animation of some sort would give a lot of intuition as to how the targets are moving.

gawebb-dstl:
> * It needs a cover image putting in.

I think this will be done automatically by read the docs 


> * It probably needs to show tracks being plotted, since currently the example shows only ground truth and detections but there is functionality in the source code for track plotting.

There is PR in the pipeline to do tracking in angular state space. That will give us angle tracks. In that PR I'll update this example.


> * The polar plots themselves could do with some more description in the titles. While polar plots look really cool, I personally have to concentrate quite hard to understand the information they're showing. Better descriptions would help a lot with that.

**Agreed. I'll update the branch with that** 


> * The scenario has 2 targets in 3D moving in all 3 dimensions, but the static plot only shows 2 dimensions.

**I'll update the example to put the scenario in 2D**


> * Could potentially get a 2D animated plotter in there as well? I tried to add this but it hasn't got functionality to automatically plot moving platforms. But an animation of some sort would give a lot of intuition as to how the targets are moving.

I may look into this for a later PR

stonesoup/plotter.py
```diff
@@ -1361,6 +1361,238 @@ def hide_plot_traces(self, items_to_hide: set):
...
+        Parameters
+        ----------
+        truths : Collection of :class:`~.GroundTruthPath`
+            Collection of  ground truths which will be plotted. If not a collection,
+            and instead a single :class:`~.GroundTruthPath` type, the argument is modified to be a
+            set to allow for iteration.
+        mapping: list
+            List of items specifying the mapping of the position components of the state space.
+        truths_label: str
+            Label for truth data. Default is "Ground Truth"
```
mharris-dstl, line 1463:
```suggestion
            Label for truth data. Default is "Ground Truth".
```

stonesoup/plotter.py
```diff
@@ -1361,6 +1361,238 @@ def hide_plot_traces(self, items_to_hide: set):
...
+            Detections which will be plotted. If measurements is a set of lists it is flattened.
+        mapping: list
+            List of items specifying the mapping of the position components of the state space.
+        measurement_model : :class:`~.Model`, optional
+            User-defined measurement model to be used in finding measurement state inverses if
+            they cannot be found from the measurements themselves.
+        measurements_label : str
+            Label for the measurements.  Default is "Measurements".
+        convert_measurements: bool
+            Should the measurements be converted before being plotted. Default is True
```
mharris-dstl, line 1501:
```suggestion
            Should the measurements be converted before being plotted. Default is True.
```

stonesoup/plotter.py
```diff
@@ -1361,6 +1361,238 @@ def hide_plot_traces(self, items_to_hide: set):
...
+            components of the state space.
+        uncertainty : bool
+            If True, function plots uncertainty ellipses.
+        particle : bool
+            If True, function plots particles.
+        track_label: str
+            Label to apply to all tracks for legend.
+        \\*\\*kwargs: dict
+            Additional arguments to be passed to scatter function. Defaults are
+            ``mode='markers+lines'``
```
mharris-dstl, line 1577:
```suggestion
            ``mode='markers+lines'``.
```

mharris-dstl (changes requested):

mharris-dstl (approved):

docs/examples/Polar_Plotting.py
```diff
@@ -0,0 +1,197 @@
...
+mapping = [0, 2]
+plotter_xy.plot_ground_truths(target_1, mapping=[0, 2], truths_label="Target 1")
+plotter_xy.plot_ground_truths(target_2, mapping=[0, 2], truths_label="Target 2")
+plotter_xy.fig
+
+# %%
+# Generate Angular State Space Data
+# ----------------------------------
+# Measure the previously generated cartesian state space using a
+# :class:`~.RadarElevationBearingRange` sensor:
```
hpritchett-dstl, line 96: The sensor used in this example is the RadarBearingRange sensor, rather than RadarElevationBearingRange.
```suggestion
# :class:`~.RadarBearingRange` sensor:
```

hpritchett-dstl (approved):

stonesoup/types/angle.py
```diff
@@ -194,6 +194,9 @@ def average(cls, angles, weights=None):
 
         return cls(result)
 
+    def to_plotly_json(self):
+        return float(self)
+
```
sdhiscocks, line 199: Is this used?

gawebb-dstl: Yes, I believe it is needed to plot angles in using Plotly

sdhiscocks: Would it be possible to use `float()` or `.dtype(np.float_)` in plotting calls instead?

gawebb-dstl: You could but you would have to repeat that code in every `plot_###` function in the `Plottery` and `PolarPlotterly` classes. Plotting `Angle` types using Plotly fails on my machine without it. To clarify, the function isn't called directly anywhere in stone-soup but it is called by the plotly package.

gawebb-dstl: The `default` function in plotly/_plotly_utils/utils.py tries to convert `Angle` types but fails currently. We could try and edit the `Angle` class so it is picked up by one of the other encode methods but this would probably raise other implications.

[Link to default function code](https://github.com/plotly/plotly.py/blob/57e4d1d33c67c5cc715bec5c3c240dd6f4c3b10d/packages/python/plotly/_plotly_utils/utils.py#L85 )

sdhiscocks: Ah, I see. I'll merge now.

Thanks for link to the code. Looks like Plotly does have encode for Decimal type, but not for Real or other numbers, which would be useful to add. I may raise an issue with them.

PR #799, opened 2023-05-03
Kafka readers and writers

sglvladi:
This PR contributes the following classes:
- `KafkaDetectionReader`: A detection reader that reads detections from a Kafka broker
- `KafkaGroundTruthReader`: A ground truth reader that reads ground truths from a Kafka broker
- `KafkaWriter`: A simple Kafka writer that writes data to a Kafka topic.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,10 +27,10 @@ install_requires =
     rtree
     scipy
     utm
-    confluent-kafka
 
 [options.extras_require]
 dev =
+    confluent-kafka
     flake8<5
     folium
     h5py
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,6 +27,7 @@ install_requires =
     rtree
     scipy
     utm
+    confluent-kafka
 
 [options.extras_require]
 dev =
--- a/stonesoup/reader/kafka.py
+++ b/stonesoup/reader/kafka.py
@@ -6,15 +6,21 @@ from queue import Empty, Queue
 from threading import Thread
 from typing import Dict, List, Collection
 
+try:
+    from confluent_kafka import Consumer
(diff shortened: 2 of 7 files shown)


jswright-dstl (approved):

setup.cfg
```diff
@@ -27,6 +27,7 @@ install_requires =
     rtree
     scipy
     utm
+    confluent-kafka
```
sdhiscocks, line 30: I think this should be optional extra.

stonesoup/reader/kafka.py
```diff
@@ -0,0 +1,211 @@
...
+from datetime import datetime, timedelta
+from math import modf
+from queue import Empty, Queue
+from threading import Thread
+from typing import Dict, List, Collection
+
+import numpy as np
+from confluent_kafka import Consumer
+from dateutil.parser import parse
+from stonesoup.base import Property
```
sdhiscocks, line 12:
```suggestion
from dateutil.parser import parse

from stonesoup.base import Property
```

stonesoup/writer/kafka.py
```diff
@@ -0,0 +1,28 @@
+import json
+from typing import Dict
+
+from confluent_kafka import Producer
+from stonesoup.base import Property
```
sdhiscocks, line 5:
```suggestion
from confluent_kafka import Producer

from stonesoup.base import Property
```

sdhiscocks (changes requested):
Looks good, but I think `confluent_kafka` should be optional.

Future aspiration would be to use JSON Path syntax to allow selection of fields, which would enable this to be used more easily with GeoJSON for example.

jswright-dstl (approved):

PR #802, opened 2023-05-09
Expected Likelihood Particle Filters

sglvladi:
This PR adds the following Tracker classes:
 - `SingleTargetExpectedLikelihoodParticleFilter`
 - `MultiTargetExpectedLikelihoodParticleFilter`

The above are single/multi target implementations of the approach described in [1]. Essentially, these implement (J)PDA for particle filtering, but can also work with (G)NN data associators. 

A running example can be found [here](https://gist.github.com/sglvladi/348ea15a7941a68946d44a104c5463f3).

[1] Marrs, A., Maskell, S., and Bar-Shalom, Y., “Expected likelihood for tracking in clutter with particle filters”, in <i>Signal and Data Processing of Small Targets 2002</i>, 2002, vol. 4728, pp. 230–239. doi:10.1117/12.478507.


A-acuto:
Hey, I don't know if it is my implementation that it is not working fine or something more complex, but if I use a JPDA data associator in a simple Particle filter (so no ELPF) I get this error:
`ValueError: Length of weights not compatible with specified axis.`
A "not working" example can be found [here](https://gist.github.com/A-acuto/78f56b561414599435a0ada52be4a9c8), if in the example you switch to a distance based data associator it is working fine.

docs/examples/ELPF_multitarget_example.py
```diff
@@ -0,0 +1,280 @@
...
+    updater=updater,
+    resampler=resampler)  # if the resampler is not specified, ~SystematicResampler is used
+
+# loop over the tracker generate the final tracks
+tracks = set()
+for time, current_tracks in tracker:
+    tracks.update(current_tracks)
+
+# Visualise the final tracks obtained
+plotter.plot_tracks(tracks, [0, 2], track_label="ELPF tracks", line=dict(color="brown"))
```
sdhiscocks, line 263:
```suggestion
plotter.plot_tracks(tracks, [0, 2], track_label="ELPF tracks")
```

docs/examples/ELPF_multitarget_example.py
```diff
@@ -0,0 +1,280 @@
...
+
+# Simulation parameters
+np.random.seed(1908)  # Set a random seed
+simulation_start = datetime.now().replace(microsecond=0)
+prob_detect = 0.95  # detection probability
+simulation_timesteps = 100
+
+# Clutter parameters
+clutter_rate = 1
+clutter_area = np.array([[-1, 1], [-1, 1]]) * 25
```
sdhiscocks, line 48: This doesn't match area of clutter below.

sdhiscocks: And also noticed that one target starts outside area of clutter.

jswright-dstl (approved):

sdhiscocks (approved):

PR #805, opened 2023-05-17
Fixed neighbour.py GNNWith2DAssignment class.

jrai51:
scipy.optimize.linear_sum_assignment no longer accepts a flag to accept maximization problems. Additionally, it does not accept infinity values. This issue became prevalent in running StoneSoup's initiator/deleter tutorial, which would not work due to this change. Problem was fixed by replacing infinity values with sys.maxsize, and simply making an if statement to check if probability_flag is True, indicating a maximization problem, then just negating the matrix.

Now passes all test cases.


jrai51:
Turns out I was using an older version of scipy rather than a newer one... Occam's Razor?

PR #806, opened 2023-05-19
Enable Forward Declaration of Type Hints for Properties

orosoman-dstl:
Forward declaration of type hints is a method to allow type checking in instances where a full import would cause a problem, for example commonly an error due to a circular import. See `typing.TYPE_CHECKING` on the [python docs](https://docs.python.org/3/library/typing.html) for more information. 

This may not be the neatest solution, as it allows any type hint declared as a string. Ideas for improvement are welcome, but this is functional at least.

### DIFF
--- a/stonesoup/base.py
+++ b/stonesoup/base.py
@@ -60,6 +60,7 @@ from abc import ABCMeta
 from collections import OrderedDict
 from copy import copy
 from types import MappingProxyType
+from typing import TYPE_CHECKING
 
 from ._util import cached_property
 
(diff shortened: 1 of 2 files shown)


orosoman-dstl:
This breaks an existing type hint test. Is there a way to check the condition: `isinstance(value.cls, str) and we_are_type_checking`? Seems difficult to do at all neatly.

sdhiscocks:
> This breaks an existing type hint test. Is there a way to check the condition: `isinstance(value.cls, str) and we_are_type_checking`? Seems difficult to do at all neatly.

I think the test probably just needs to be removed, as this change makes it acceptable to be a string?

sdhiscocks (approved):

PR #809, opened 2023-05-30
Added support for a 4D state vector when using a range, bearing, range rate radar model

PACarniglia:
The RadarBearingRangeRate2D sensor and associated CartesianToBearingRangeRate2D measurement model were created to support tracking from measurements of bearing, range, and range rate to a 4D state space [x, vx, y, vy]. The current RadarBearingRangeRate/CartesianToBearingRangeRate class only supports a 6D state vector [x, vx, y, vy, z, vz] as a requirement for the creation of the rotation matrix.

Additionally, both measurement models were given inverse functions to support plotting measurements to a Cartesian plot space.

### DIFF
--- a/stonesoup/models/measurement/nonlinear.py
+++ b/stonesoup/models/measurement/nonlinear.py
@@ -863,7 +863,7 @@ class CartesianToBearingRangeRate2D(_AngleNonLinearGaussianMeasurement):
         xy_rot = self.rotation_matrix[:len(self.mapping), :len(self.mapping)] @ xy_pos
 
         # Convert to Polar
-        rho, phi= cart2pol(xy_rot[0, :], xy_rot[1, :])
+        rho, phi = cart2pol(xy_rot[0, :], xy_rot[1, :])
 
         # Determine the net velocity component in the engagement
         xy_vel = state.state_vector[self.velocity_mapping, :] - self.velocity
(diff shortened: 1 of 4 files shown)


stonesoup/models/measurement/nonlinear.py
```diff
@@ -782,12 +782,200 @@ def function(self, state, noise=False, **kwargs) -> StateVector:
...
+        x, y = pol2cart(rho, phi)
+        # because only rho_rate is known, only the components in
+        # x,y of the range rate can be found.
+        x_rate = np.cos(phi) * rho_rate
+        y_rate = np.sin(phi) * rho_rate
+
+        inv_rotation_matrix = inv(self.rotation_matrix)
+        out_vector = StateVector([[0.], [0.], [0.], [0.], [0.], [0.]])
+        out_vector[self.mapping, 0] = x, y
+        out_vector[self.velocity_mapping, 0] = x_rate, y_rate
```
sdhiscocks, line 798: From test failures, these mappings are 3D, but this is only assigning 2D values.

PACarniglia: This could be fixed by setting z and z_rate to 0 when mapping out_vector - is this supported by the theory, or is there a way to calculate the z parameters from measurements of range/bearing/range rate?

csherman-dstl (approved):

sdhiscocks (approved):

PR #811, opened 2023-06-01
Add metric MultiManager and MetricPlotter

rcgorman-dstl:
- Have replaced SimpleManager metric manager with new MultiManager class so this will be a breaking change. 
- MultiManager can take multiple metric generators of the same type and multiple sets of the same type to generate metrics. 
- All metrics that take in track and truth can now also take in 2 tracks.
- Additional functionality added to MultiManager through new methods to help easier extraction of metrics and insights.
- New MetricPlotter class created to make plotting metrics easier.

Update:
- Have added SimpleManager metric manager back in to avert breaking change. It should operate in the same way as before.

### DIFF
--- a/docs/examples/Metrics.py
+++ b/docs/examples/Metrics.py
@@ -425,9 +425,13 @@ siap_table = SIAPTableGenerator(siap_averages_PF).compute_metric()
 #
 # To use :class:`MetricPlotter` we first create an instance of the class. We can then use the .plot_metrics() method to
 # plot the metrics of our choice. Key features of :class:`MetricPlotter`:
-# - You can specify which metrics you want to plot using the generator_names argument. If you don't pass the
-#   generator_names argument, .plot_metrics() will automatically extract all plottable metrics from those you have
-#   generated and plot them.
+# - You can specify exactly which metrics you want to plot by specifying the optional generator_names and metric_names
+#   parameters.
+# - generator_names allows you to specify which generators you want to plot metrics from. If you don't pass the
+#   generator_names argument, .plot_metrics() will automatically extract all plottable metrics from all metric
+#   generators in the :class:`MultiManager` and plot them.
+# - metric_names allows you to specify which specific metrics types you want to plot from the metric generators. This is
+#   mostly relevant for the SIAP metrics. You can view which metrics are plottable with MetricPlotter.plottable_metrics.
 # - .plot_metrics has a combine_plots argument which defaults to True. This means that all metrics of the same type will
 #   be plotted together on a single subplot.
 # - You can specify additional formatting, like colour and linestyle, for the plots produced by using keyword arguments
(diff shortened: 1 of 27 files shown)


nperree-dstl (commented):
Seems to work really well, comments are mostly documentation issues. I think you also need to remove the old `SimpleManager` from test_pcrbmetric and test_uncertaintymetric to stop tests failing.

docs/examples/Metrics.py
```diff
@@ -1,212 +1,485 @@
...
 # %%
-# Once we've created a set of metrics, these are added to a Metric Manager, along with the
-# associator. The associator can be used by multiple metric generators, only being run once as this
-# can be a computationally expensive process; in this case, only SIAP Metrics requires it.
-from stonesoup.metricgenerator.manager import SimpleManager
+# Add data to metric manager and generate metrics
+# -----------------------------------------------
+# Now that we have all of our ground truth, detections, and tracks data, we can add it to the metric manager.
+#
+# The MultiManager .add_data() method takes a dictionary of lists/sets of ground truth, detections, and/or tracks data.
```
jswright-dstl, line 346:
```suggestion
# The MultiManager :meth:`.add_data()` method takes a dictionary of lists/sets of ground truth, detections, and/or tracks data.
```

jswright-dstl (approved):

jswright-dstl (approved):

sdhiscocks:
The breaking change makes this difficult to merge at this point. However, it may be possible to make it backwards compatible fairly easy:
 - Add back in the `SimpleManager`, inheriting from MultiManager, with "tracks", "groundtruth_paths" and "detections" as default keys in `states_set`.
 - Have property on `SimpleManager` for `tracks`, `groundtruth_paths` and `detections` that returns values from those dictionary keys mentioned above.
 - Change metrics so default for `tracks_key` and `truths_key` are "tracks" and "groundtruth_paths".
 - Also provide sensible default `generator_name` for each metric.

I've not tried this, but something like this should work or at least be a good start.

stonesoup/metricgenerator/manager.py
```diff
@@ -74,30 +67,65 @@ def generate_metrics(self):
...
-        if self.associator is not None and self.association_set is None:
-            self.associate_tracks()
-
         metrics = {}
-        for generator in self.generators:
+
+        generators = self.generators if isinstance(self.generators, list) else [self.generators]
+
+        for generator in generators:
+            if isinstance(generator, SIAPMetrics):
```
sdhiscocks, line 75: Referencing to specific metric here doesn't seem appropriate. Why does this need to be done?

stonesoup/metricgenerator/manager.py
```diff
@@ -74,30 +67,65 @@ def generate_metrics(self):
...
+
+        Returns
+        ----------
+        : dict of :class`SIAPMetrics` averages
+        """
+        siap_metrics = self.metrics[generator_name]
+        siap_averages = {siap_metrics.get(metric) for metric in siap_metrics
+                         if metric.startswith("SIAP") and not metric.endswith(" at times")}
+
+        return siap_averages
```
sdhiscocks, line 131: These methods also seem specific to the name or specific metric. A different approach is needed here.

PR #816, opened 2023-06-09
Plot FOV on Sensor Management tutorials

nperree-dstl:
Add sensor field of view to plot animations


mharris-dstl (commented):
Looks good, it's great to be able to see sensor management dynamically - I think it'll help new people understand what's going on. 

I don't really mind either way but it could be worth putting the sensor fov into a function to save the code being written out twice, with a note saying not to worry about the contents of the function. I know that when I went through tutorials to start with, I tried to understand what each piece of code did, and would probably get bogged down by the code.

Having said that, it does the job well so I'm happy either way.

hpritchett-dstl (approved):

mharris-dstl (approved):

hpritchett-dstl (approved):

PR #830, opened 2023-07-20
Regulariser amendment and implementation in ParticleUpdater

timothy-glover:
The prior used for the `MCMCRegulariser` should be projected through time without the inclusion of noise. To do this, `transition_model` has been added as a property of the MCMCRegulariser. The transition model is used to propagate the prior particles at the beginning of the `regularise()` method. The support for lists of type `Particle` has also been dropped to allow for the calculation of the time interval in the `regularise()` method.

### DIFF
--- a/stonesoup/regulariser/particle.py
+++ b/stonesoup/regulariser/particle.py
@@ -5,6 +5,8 @@ from scipy.stats import multivariate_normal, uniform
 from .base import Regulariser
 from ..functions import cholesky_eps
 from ..types.state import ParticleState
+from ..models.transition import TransitionModel
+from ..base import Property
 
 
 class MCMCRegulariser(Regulariser):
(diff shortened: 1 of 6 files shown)


stonesoup/regulariser/particle.py
```diff
@@ -24,17 +26,19 @@ class MCMCRegulariser(Regulariser):
...
     .. [1] Robert, Christian P. & Casella, George, Monte Carlo Statistical Methods, Springer, 1999.
 
-    .. [2] Ristic, Branco & Arulampalam, Sanjeev & Gordon, Neil, Beyond the Kalman Filter:
+    .. [2] Ristic, Branko & Arulampalam, Sanjeev & Gordon, Neil, Beyond the Kalman Filter:
         Particle Filters for Target Tracking Applications, Artech House, 2004. """
 
+    transition_model: TransitionModel = Property(doc="Transition model used for prediction")
+
     def regularise(self, prior, posterior, detections):
         """Regularise the particles
```
sdhiscocks, line 35: Could you also pass an optional transition model, using the `hypothesis.prediction.transition_model`, from the updater, and falling back to provide one (on `self`) if not provided. Possibly via `**kwargs` or even passing `hypothesis` instead?

PR #835, opened 2023-08-11
Minor Addition and Changes to MultiMeasurementInitiator

gawebb-dstl:
`NoHistoryMultiMeasurementInitiator` was created so track metrics wouldn’t take into account the part of the track that was in ‘holding’. ‘Holding’ tracks aren’t available to the tracker. When a holding track is upgraded to a full/sure track it takes its history with it. When assessing tracks post-simulation with metrics the full history is used which can be misrepresentative.

An option was added to skip measurements that don’t have a `ReversibleModel` measurement model in `MultiMeasurementInitiator`

### DIFF
--- a/stonesoup/initiator/simple.py
+++ b/stonesoup/initiator/simple.py
@@ -184,6 +184,10 @@ class MultiMeasurementInitiator(GaussianInitiator):
         doc="Initiator used to create tracks. If None, a :class:`SimpleMeasurementInitiator` will "
             "be created using :attr:`prior_state` and :attr:`measurement_model`. Otherwise, these "
             "attributes are ignored.")
+    skip_non_reversible: bool = Property(
+        default=False, doc="Skip measurements that do not have a reversible measurement model. "
+                           "Only allow measurements with a measurement model that is an instance "
+                           "of a :class:`~.LinearModel` or a :class:`~.ReversibleModel`.")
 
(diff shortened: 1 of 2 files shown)


gawebb-dstl:
The docs are failing to build docs/demos/Video_Processing.py ModuleNotFoundError: No module named 'pytube'

I don't think this is related to any of my changes

stonesoup/initiator/simple.py
```diff
@@ -165,7 +165,7 @@ class MultiMeasurementInitiator(GaussianInitiator):
     initiated only to then be removed shortly after.
     Does cause slight delay in initiation to tracker."""
 
-    prior_state: GaussianState = Property(doc="Prior state information")
+    prior_state: GaussianState = Property(default=None, doc="Prior state information")
```
sdhiscocks, line 168: This change makes this a breaking change due to change in argument order, something ideally we'd look to avoid now at v1 release.

gawebb-dstl: I think the default value for ``prior_state`` could be omitted for this pull request. However I think it should be added in the future. Is there a plan on how to integrate breaking changes, maybe a development branch or v2 branch?

sdhiscocks (approved):

nperree-dstl (approved):

PR #842, opened 2023-08-18
check positive definetness of CovarianceMatrix and PrecisionMatrix

aeblc:


aeblc:
I expected that a covariance matrix is positive definite (or at least positive semi-definite) and so tests are created accordingly. Since that is not case, and there are too many failing tests the issue is more complicated than I thought.

Issue #841, opened 2023-08-18
Documentation of Q_t does not match implementation


jswright-dstl:
https://github.com/dstl/Stone-Soup/blob/8c4fcf2c7a5fc2c3b777d1e9ff4c1f62bc8fb4e0/stonesoup/models/transition/nonlinear.py#L69C13-L80C35

The documentation for Q_t in the ConstantTurn does not match the implementation.

PR #844, opened 2023-08-21
Fix non linear constant turn transition model

orosoman-dstl:
I found a few bugs in the `ConstantTurn` transition model, which this PR aims to fix. 

1.  As detailed in #841, the documentation for `Q_t` did not match the implementation. 
2.  The bottom right term in the `Q_t` array was erroneously divided by the change in time, which would result in the noise of the turn rate being smaller for a larger change in time. I'm not really sure personally why this value is not instead **multiplied** by _dt_, but this seems to be a misunderstanding on my part and not a mistake. 
3. There was an issue where if you put in an integer turn rate of zero, you would get an error for division by zero. This is a little bit academic as it's not really the intended use for this transition model, but I thought it was worth fixing for consistency. 

I have also raised up a misbehaving comma in the first tutorial as it looked slightly messy.

### DIFF
--- a/stonesoup/models/transition/nonlinear.py
+++ b/stonesoup/models/transition/nonlinear.py
@@ -67,16 +67,11 @@ class ConstantTurn(GaussianTransitionModel, TimeVariantModel):
 
         .. math::
              Q_t & = & \begin{bmatrix}
-                          \frac{dt^4q_x^2}{4} & \frac{dt^3q_x^2}{2} & \frac{dt^4q_xq_y}{4} &
-                              \frac{dt^3q_xq_y}{2} & \frac{dt^2q_xq_\omega}{2} \\
-                          \frac{dt^3q_x^2}{2} & dt^2q_x^2 & \frac{dt^3q_xq_y}{2} & dt^2q_xq_y &
-                              dt q_x q_\omega \\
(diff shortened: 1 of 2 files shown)


jswright-dstl (approved):

stonesoup/models/transition/nonlinear.py
```diff
@@ -100,7 +95,8 @@ def function(self, state, noise=False, **kwargs) -> StateVector:
         sv1 = state.state_vector
         turn_rate = sv1[4, :]
         # Avoid divide by zero in the function evaluation
-        turn_rate[turn_rate == 0.] = np.finfo(float).eps
+        if turn_rate[0] == 0:
+            turn_rate = Matrix([np.finfo(float).eps])
```
sdhiscocks, line 99: I think, like before, this should check the entire array and set any zero elements to eps.

sdhiscocks (changes requested):

stonesoup/models/transition/nonlinear.py
orosoman-dstl: OK, I've reverted to the original code, but now with a check to ensure the Matrix is `float` before trying to fix zero elements.

mharris-dstl (approved):

sdhiscocks (approved):

PR #850, opened 2023-09-04
Create IsotropicPlume model and gas sensor

timothy-glover:
This PR introduces the `IsotropicPlume` model for calculating gas concentration at a location based on a source term. The model implements the isotropic plume method from the literature (reference in docs) as well as specific likelihood function for evaluating measurement likelihood. This is also based on work presented in literature, with some modifications to account for the fact that this is a simulated sensor with more ideal characteristics than a real sensor. Documentation has been created including the main functions behind the model.

A gas sensor has also been created to carry the `IsotropicPlume` model that can (and will) be used for a later sensor management example.

A repository in MATLAB on which this PR is based can be seen at: https://github.com/Cunjia-Liu/AutoSTE


timothy-glover:
I am unsure why the docs are not building correctly. It seems to be referring to the GMPHD tutorial plotting style in the error. The docs build without any warnings or errors on my PC.

stonesoup/models/measurement/nonlinear.py
```diff
@@ -1248,3 +1250,294 @@ def pdf(self, state1, state2, **kwargs):
     def logpdf(self, *args, **kwargs):
         # As pdf replaced, need to go to first non GaussianModel parent
         return super(ReversibleModel, self).logpdf(*args, **kwargs)
+
+
+class IsotropicPlume(GaussianModel, MeasurementModel):
```
sdhiscocks, line 1255: I think this should be in a separate module, as it's very specific. `stonesoup.models.measurement.gas` possibly, which matches name used for *sensor*?

stonesoup/models/measurement/nonlinear.py
```diff
@@ -1248,3 +1250,294 @@ def pdf(self, state1, state2, **kwargs):
...
+           Robotics, Vol. 36, 797-917, 2019
+    """
+
+    ndim_state: int = Property(
+        default=8,
+        doc="Number of state dimensions"
+    )
+
+    mapping: Sequence[int] = Property(
+        default=range(0, 8),
```
sdhiscocks, line 1319:
```suggestion
        default=tuple(range(0, 8)),
```

stonesoup/models/measurement/nonlinear.py
```diff
@@ -1248,3 +1250,294 @@ def pdf(self, state1, state2, **kwargs):
...
+            then the model also does not consider the :attr:`sensing_threshold`
+            and :attr:`missed_detection_probability`
+
+        Returns
+        -------
+        : :class:`numpy.ndarray` of shape (1, 1)
+            The model function evaluated with the provided source term
+        """
+
+        x, y, z, Q, u, phi, ci, cii = state.state_vector[self.mapping].view(np.ndarray)
```
sdhiscocks, line 1387:
```suggestion
        x, y, z, Q, u, phi, ci, cii = state.state_vector[self.mapping, :].view(np.ndarray)
```

stonesoup/models/measurement/nonlinear.py
```diff
@@ -1248,3 +1250,294 @@ def pdf(self, state1, state2, **kwargs):
...
+        ----------
+        state1 : :class:`~.State`
+        state2 : :class:`~.State`
+
+        Returns
+        -------
+        : :class:`~.Probability`
+            The likelihood of ``state1``, given ``state2``
+        """
+        return Probability.from_log_ufunc(self.logpdf(state1, state2, **kwargs))
```
sdhiscocks, line 1504: Just to make it clear that this is doing same as parent, could use this:
```suggestion
        return super().pdf(state1, state2, **kwargs)
```

stonesoup/models/transition/base.py
```diff
@@ -7,6 +7,7 @@
 
 from ..base import Model, GaussianModel
 from ...base import Property
+from ...types.array import StateVectors
 from ...types.state import StateVector
```
sdhiscocks, line 11: Odd, `StateVector` should have been from `array` module as well. May as well fix that in this PR.
```suggestion
from ...types.array import StateVector, StateVectors
```

sdhiscocks (approved):

stonesoup/models/measurement/nonlinear.py
timothy-glover: This has been done in the latest commit.

nperree-dstl (approved):

PR #855, opened 2023-09-13
Introduce ability to constrain particle states in ParticleUpdater and MCMCRegulariser

timothy-glover:
This PR has introduced the optional property `constraint_func` to `ParticleUpdater` and `MCMCRegulariser`. If defined, this is intended to enforce that particle states should not exceed certain values by modifying the associated weights, increasing the likelihood of resampling. Here the constraint function should accept a `ParticleState` type and return a logical index of particles that exceed the defined constraints. An example of this can be seen here:
https://github.com/timothy-glover/Stone-Soup/blob/7ae343e147176107a35fafa81c00c2a7d879561e/stonesoup/regulariser/tests/test_particle.py#L16-L18

A bug was also spotted in the test for the `ParticleUpdater`. The original line, seen below, was passing regardless of what the weights are. This has been replaced with a check to ensure the sum of the weights is correct. 
https://github.com/dstl/Stone-Soup/blob/f24090cc919b3b590b84f965a3884ed1293d181d/stonesoup/updater/tests/test_particle.py#L74

### DIFF
--- a/stonesoup/regulariser/particle.py
+++ b/stonesoup/regulariser/particle.py
@@ -1,7 +1,7 @@
 import copy
 import numpy as np
 from scipy.stats import multivariate_normal, uniform
-from typing import Sequence
+from typing import Sequence, Callable
 
 from .base import Regulariser
 from ..functions import cholesky_eps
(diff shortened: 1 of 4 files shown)


stonesoup/regulariser/particle.py
```diff
@@ -32,6 +32,14 @@ class MCMCRegulariser(Regulariser):
 
     transition_model: TransitionModel = Property(doc="Transition model used for prediction",
                                                  default=None)
+    constraint_func: Callable = Property(
+        default=None,
+        doc="Callable, user defined function for applying "
+            "constraints to particle states. This is done by reverting "
+            "that are moved to a state outside of the defined constraints "
```
jswright-dstl, line 39:
```suggestion
            "constraints to particle states. This is done by reverting particles "
            "that are moved to a state outside of the defined constraints "
```

timothy-glover: Well spotted. Fixed and improved clarity in latest commit.

jswright-dstl (approved):

PR #865, opened 2023-09-26
Function to linearly interpolate state mutable sequences

gawebb-dstl:
The function will take a `StateMutableSequence` and a `times: List[datetime.datetime]`. The function will create a new `StateMutableSequence` with states only at the times specified by `times`

### DIFF
--- a/stonesoup/functions/interpolate.py
+++ b/stonesoup/functions/interpolate.py
@@ -1,0 +1,261 @@
+import copy
+import datetime
+import timeit
+import warnings
+from typing import Union, List, Iterable
+
+import numpy as np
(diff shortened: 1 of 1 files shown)


gawebb-dstl:
I've looked at 4 options (see code). All 4 options run function correctly. Option 4 is the quickest* so will be taken forward

Output when running the file:
```
Option 1 took 3.9378470000000005
Option 2 took 79.7136048
Option 3 took 1.762432000000004
Option 4 took 1.326851700000006
Option 1 worked
Option 2 worked
Option 3 worked
Option 4 worked
```

*Option 3 is quicker if all of the states need to be produced by interpolation

PR #868, opened 2023-10-02
Introduce Kullback-Leibler Divegence measure and reward function

timothy-glover:
This PR is to introduce the Kullback-Leibler Divergence (KLD) measure and reward function for calculating the relative entropy between two distributions. This implementation uses the discrete formulation for KLD and is currently only compatible with particle state types. The mathematical form of the equation is provided in the measure, `KLDivergence`. The reward function, `ExpectedKLDivergence`, implements the KLD measure in order to decide which candidate sensing (or movement) action leads to the most informative measurement based on the provided predictor and updater. A second KLD based reward function, `MultiUpdateExpectedKLDivergence`, which generates expected detections by resampling from the predicted particle state to create a subsample of possible measurements based on the target distribution. The KLD is then calculated based on each of the measurements. This comes at a computational burden. I am open to name suggestions if the general consensus on these class names is that these are confusing.

Tests for each new class have been introduced or incorporated into existing test frameworks. The `MultiUpdateExpectedKLDivergence` results in a significant increate in test time for the sensor management module. The sensor management test takes around 8 minutes 30 seconds on my fairly standard PC and takes total testing time to 9 minutes 24 seconds.

### DIFF
--- a/stonesoup/measures.py
+++ b/stonesoup/measures.py
@@ -1,4 +1,5 @@
 import copy
+import math
 from abc import abstractmethod
 from functools import lru_cache
 
@@ -6,7 +7,7 @@ import numpy as np
 from scipy.spatial import distance
 
 from .base import Base, Property
-from .types.state import State
+from .types.state import State, ParticleState, GaussianState
 
 
 class Measure(Base):
@@ -401,3 +402,33 @@ class ObservationAccuracy(Measure):
         mins = [min(s1, s2) for s1, s2 in zip(s1, s2)]
         maxs = [max(s1, s2) for s1, s2 in zip(s1, s2)]
         return np.sum(mins)/np.sum(maxs)
+
+class KLDivergence(Measure):
+    """  """
+
+    def __call__(self, state1, state2):
+        r"""
+        Computes the Kullback–Leibler divergence from :class:`~.State` object1 to :class: '~.State object2
+
+        Parameters
+        ----------
+        state1 : :class:`~.State`
+        state2 : :class:`~.State`
+
+        Returns
+        -------
+        float
+            Kullback–Leibler divergence from :class:`~.State` object1 to :class: '~.State object2
+
+        """
+        if isinstance(state1, ParticleState) and isinstance(state2, ParticleState):
+            if len(state1.particles) == len(state2.particles):
+                kld = 0
+                for ii in range(0, len(state1.particles)):
+                    kld += math.exp(state1.log_weight[ii])*(state1.log_weight[ii] - state2.log_weight[ii])
+            else:
+                raise ValueError('The input objects different number of particles.')
+        else:
+            raise TypeError('The inputs are possible states')
+
+        return kld
--- a/stonesoup/measures.py
+++ b/stonesoup/measures.py
@@ -1,5 +1,4 @@
 import copy
-import math
 from abc import abstractmethod
 from functools import lru_cache
 
@@ -7,7 +6,7 @@ import numpy as np
 from scipy.spatial import distance
 
 from .base import Base, Property
-from .types.state import State, ParticleState, GaussianState
+from .types.state import State, ParticleState
 
 
 class Measure(Base):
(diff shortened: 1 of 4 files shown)


stonesoup/measures.py
```diff
@@ -401,3 +401,63 @@ def __call__(self, state1, state2):
...
+                # Do not consider NANs and inf in the subtraction
+                log_term[~invalid_indx] = state1.log_weight[~invalid_indx] \
+                    - state2.log_weight[~invalid_indx]
+
+                kld = np.sum(np.exp(state1.log_weight)*log_term)
+            else:
+                raise ValueError(f'The input sizes are not compatible '
+                                 f'({len(state1.particles)} != {len(state2.particles)})')
+        else:
+            raise TypeError('state1 or state2 is not a ParticleState')
```
sdhiscocks, line 461: Do you think this can be extended to other states in the future? Maybe this could raise a `NotImplementedError`, stating that only `ParticleState` is currently supported?

sdhiscocks (commented):
On the slow tests, we currently have something similar with test for remote data where they are skipped by default.

How about using something like `pytest-skip-slow` to mark the slow tests? I think in the CI we should have the slow tests enabled, but makes sense you may be default want to skip them locally when not working on anything related.

stonesoup/measures.py
timothy-glover: The only reason it only supports particle states currently is due to this being the discretised version of KLD. You could use it with any continuous distribution if we implemented a sampling technique to discretise the provided continuous distribution, perhaps using the sampler. I am happy to change this to a `NotImplementedError` for now.

timothy-glover: I will also add this to the `ExpectedKLDivergence` and `MultiUpdateExpectedKLDivergence` `__init__` methods as a similar particle constraint is enforced there and they can the problems relating to continuous states there can also be overcome in this way.

stonesoup/sensormanager/reward.py
```diff
@@ -133,3 +141,182 @@ def __call__(self, config: Mapping[Sensor, Sequence[Action]], tracks: Set[Track]
...
+                predicted_track.append(Prediction.from_state(track[-1],
+                                                             timestamp=metric_time))
+
+            predicted_tracks.add(predicted_track)
+
+        for predicted_actionable_ in predicted_actionables:
+            if isinstance(predicted_actionable_, Sensor):
+                sensors = [predicted_actionable_]
+            elif isinstance(predicted_actionable_, Platform):
+                sensors = predicted_actionable_.sensors
```
sdhiscocks, line 235: I can see why you've done this, to include sensors from taskable platform, but I'm thinking this may risk accessing a sensor multiple times where the sensor is taskable, and the platform it is on is also.

I can see same bug in other other reward function as well, with skipping sensors.

Maybe something like this:
```diff
diff --git a/stonesoup/sensormanager/reward.py b/stonesoup/sensormanager/reward.py
index a122af7f..5d0d5444 100644
--- a/stonesoup/sensormanager/reward.py
+++ b/stonesoup/sensormanager/reward.py
@@ -93,7 +93,7 @@ class UncertaintyRewardFunction(RewardFunction):
         # Reward value
         config_metric = 0
 
-        predicted_sensors = list()
+        predicted_sensors = set()
         memo = {}
         # For each sensor in the configuration
         for sensor, actions in config.items():
@@ -101,7 +101,9 @@ class UncertaintyRewardFunction(RewardFunction):
             predicted_sensor.add_actions(actions)
             predicted_sensor.act(metric_time)
             if isinstance(sensor, Sensor):
-                predicted_sensors.append(predicted_sensor)  # checks if its a sensor
+                predicted_sensors.add(predicted_sensor)  # checks if its a sensor
+            elif isinstance(sensor, Platform):
+                predicted_sensors.update(predicted_sensor.sensors)
 
         # Create dictionary of predictions for the tracks in the configuration
         predicted_tracks = set()
@@ -204,7 +206,7 @@ class ExpectedKLDivergence(RewardFunction):
         kld = 0.
 
         memo = {}
-        predicted_actionables = []
+        predicted_sensors = set()
         # For each actionable in the configuration
         for actionable, actions in config.items():
             # Don't currently have an Actionable base for platforms hence either Platform or Sensor
@@ -212,7 +214,10 @@ class ExpectedKLDivergence(RewardFunction):
                 predicted_actionable = copy.deepcopy(actionable, memo)
                 predicted_actionable.add_actions(actions)
                 predicted_actionable.act(metric_time)
-                predicted_actionables.append(predicted_actionable)
+                if isinstance(actionable, Sensor):
+                    predicted_sensors.add(predicted_actionable)  # checks if its a sensor
+                elif isinstance(actionable, Platform):
+                    predicted_sensors.update(predicted_actionable.sensors)
 
         # Create dictionary of predictions for the tracks in the configuration
         predicted_tracks = set()
@@ -228,36 +233,30 @@ class ExpectedKLDivergence(RewardFunction):
 
             predicted_tracks.add(predicted_track)
 
-        for predicted_actionable_ in predicted_actionables:
-            if isinstance(predicted_actionable_, Sensor):
-                sensors = [predicted_actionable_]
-            elif isinstance(predicted_actionable_, Platform):
-                sensors = predicted_actionable_.sensors
-            # sensors = predicted_actionable_.sensors
-            for sensor in sensors:
-                # Assumes one detection per track
+        for sensor in predicted_sensors:
+            # Assumes one detection per track
 
-                detections = self._generate_detections(predicted_tracks, sensor)
+            detections = self._generate_detections(predicted_tracks, sensor)
 
-                for predicted_track, detection_set in detections.items():
+            for predicted_track, detection_set in detections.items():
 
-                    for n, detection in enumerate(detection_set):
+                for n, detection in enumerate(detection_set):
 
-                        # if detection:
-                        # Generate hypothesis based on prediction/previous update and detection
-                        hypothesis = SingleHypothesis(predicted_track, detection)
+                    # if detection:
+                    # Generate hypothesis based on prediction/previous update and detection
+                    hypothesis = SingleHypothesis(predicted_track, detection)
 
-                        # Do the update based on this hypothesis and store covariance matrix
-                        update = self.updater.update(hypothesis)
+                    # Do the update based on this hypothesis and store covariance matrix
+                    update = self.updater.update(hypothesis)
 
-                        # else:
-                        #     update = copy.copy(predicted_track[-1])
+                    # else:
+                    #     update = copy.copy(predicted_track[-1])
 
-                        kld += self.KLD(predicted_track[-1], update)
+                    kld += self.KLD(predicted_track[-1], update)
 
-                if self.method_sum is False and len(detections) != 0:
+            if self.method_sum is False and len(detections) != 0:
 
-                    kld /= len(detections)
+                kld /= len(detections)
 
         # Return value of configuration metric
         return kld
```

timothy-glover:
> On the slow tests, we currently have something similar with test for remote data where they are skipped by default.
> 
> How about using something like [`pytest-skip-slow`](https://pypi.org/project/pytest-skip-slow/) to mark the slow tests? I think in the CI we should have the slow tests enabled, but makes sense you may be default want to skip them locally when not working on anything related.

That solution seems to be ideal. The package is not installed by default however so is there a procedure for incorporating this as a required package?

jswright-dstl (approved):

sdhiscocks (approved):

PR #869, opened 2023-10-02
Resampling dependant regularisation in ParticleUpdater

timothy-glover:
This PR adds the functionality that prevents regularisation from happening if resampling has not been performed. Since the regularisation process is intended to prevent particle impoverishment as a result of resampling. Performing regularisation when resampling has not been performed may lead to unnecessary particle spreading. The changes in the PR check to see if the `ParticleState` object is the same before and after resampling. This check prevents negates the problem where a `Resampler` is defined but the resampling process has not been performed due to the effective sample size threshold not being met. If changes have occurred then the `Regulariser` will be used if it has been defined. 

The changes have been made here in such a way that will allow regularisation to occur if a `Resampler` has not been provided. The reason for this is if the author has not defined a `Resampler` but has defined a `Regulariser` then there must be an exceptional reason for doing so. If this is the case, the `ParticleUpdater` will warn the user on initialisation.

### DIFF
--- a/stonesoup/updater/particle.py
+++ b/stonesoup/updater/particle.py
@@ -1,6 +1,7 @@
 import copy
 from functools import lru_cache
 from typing import Callable
+import warnings
 
 import numpy as np
 from scipy.linalg import inv
@@ -30,8 +31,14 @@ class ParticleUpdater(Updater):
     """
 
     resampler: Resampler = Property(default=None, doc='Resampler to prevent particle degeneracy')
-    regulariser: Regulariser = Property(default=None, doc="Regulariser to prevent particle "
-                                                          "impoverishment")
+    regulariser: Regulariser = Property(
+        default=None,
+        doc='Regulariser to prevent particle impoverishment. The regulariser '
+            'is normally used after resampling. If a class:`~.Resampler` is defined,'
+            ' then regularisation will only take place if the particles have been '
+            'resampled. If the class:`~.Resampler` is not defined but a '
+            'class:`~.Regulariser` is, then regularisation will be conducted under the '
+            'assumption that the user intends for this to occur.')
 
     constraint_func: Callable = Property(
         default=None,
@@ -42,6 +49,13 @@ class ParticleUpdater(Updater):
             "should accept a :class:`~.ParticleState` object and return an array-like "
             "object of logical indices. "
     )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if self.resampler is None and self.regulariser is not None:
+            warnings.warn('`regulariser` has been defined but a `resampler` has not. This'
+                          ' is not normal procedure.')
 
     def update(self, hypothesis, **kwargs):
         """Particle Filter update step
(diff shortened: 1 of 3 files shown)


stonesoup/updater/particle.py
```diff
@@ -83,13 +97,20 @@ def update(self, hypothesis, **kwargs):
         predicted_state.log_weight = new_weight
 
         # Resample
+        resample_flag = True
         if self.resampler is not None:
-            predicted_state = self.resampler.resample(predicted_state)
+            resampled_state = self.resampler.resample(predicted_state)
+            if resampled_state == predicted_state:
+                resample_flag = False
```
sdhiscocks, line 104: There is a bug here if regulariser isn't defined. 

I think you should just set predicated state per suggestion, and then also remove `if` statement below.

```suggestion
                resample_flag = False
            predicted_state = resampled_state
```

timothy-glover: That is a problem. I will get that sorted in a new commit.

timothy-glover:
The test error seems to be a problem with the `test_gaussian_particle` test in the initiator module. The chance of the assertion error happening is <1%. Changing the tolerance on this test assertion should prevent the error.

stonesoup/updater/particle.py
```diff
@@ -30,8 +31,14 @@ class ParticleUpdater(Updater):
     """
 
     resampler: Resampler = Property(default=None, doc='Resampler to prevent particle degeneracy')
-    regulariser: Regulariser = Property(default=None, doc="Regulariser to prevent particle "
-                                                          "impoverishment")
+    regulariser: Regulariser = Property(
+        default=None,
+        doc='Regulariser to prevent particle impoverishment. The regulariser '
+            'is normally used after resampling. If a class:`~.Resampler` is defined,'
+            ' then regularisation will only take place if the particles have been '
```
hpritchett-dstl, line 38: minor documentation change

```suggestion
            'is normally used after resampling. If a class:`~.Resampler` is defined, '
            'then regularisation will only take place if the particles have been '
```

hpritchett-dstl:
Happy to approve this once the test_gaussian_particle test has been addressed.

timothy-glover:
> Happy to approve this once the test_gaussian_particle test has been addressed.

Thank you. I have implemented your suggested change above in a local commit that I will push shortly. It actually pointed me towards a documentation error as I missed the preceding colon before two class type definitions in that doc string. 

Regarding the test_gaussian_particle test, I am not sure if this is the right place to fix that. I haven't made any changes to the initiator classes in this PR or caused the test fail. Happy to fix it here though if you think that is the right thing to do.

hpritchett-dstl:
> Regarding the test_gaussian_particle test, I am not sure if this is the right place to fix that. I haven't made any changes to the initiator classes in this PR or caused the test fail. Happy to fix it here though if you think that is the right thing to do.

Yes you're right, should be addressed elsewhere as this PR did not influence it!

hpritchett-dstl (approved):

sdhiscocks (approved):

PR #874, opened 2023-10-10
adding greedy sensor manager

akenyon:


stonesoup/sensormanager/tests/test_sensormanager.py
```diff
@@ -177,3 +185,105 @@ def test_uncertainty_based_managers():
...
+    sensor.timestamp = start_time
+    sensor_set.add(sensor)
+
+    predictor = KalmanPredictor(transition_model)
+    updater = ExtendedKalmanUpdater(measurement_model=None)
+
+    prior = GaussianState([[-10], [1], [0], [1]],
+                          np.diag([0.5, 0.5, 0.5, 0.5] + np.random.normal(0, 5e-4, 4)),
+                          timestamp=start_time)
+    tracks = {Track([prior])}
```
spike-dstl, line 242: A lot of this set-up code could be moved to a conftest.py file and imported. Happy to go through it with you if that's any use.

stonesoup/sensormanager/base.py
```diff
@@ -150,3 +150,62 @@ def choose_actions(self, tracks, timestamp, nchoose=1, return_reward=False, **kw
...
+            # save nchoose best actions for the sensor
+            all_sensor_actions[sensor] = selected_actions
+
+        # convert from single dict of sensor: list(actions) to list of dicts of sensors: actions
+        all_selected_configs = [{sensor: all_sensor_actions[sensor][i]
+                                 for sensor in all_sensor_actions}
+                                for i in range(nchoose)]
+
+        # Return mapping of sensors and chosen actions for sensors
+        return all_selected_configs
```
spike-dstl, line 211: Some variable names you use are different to names that appear to do the same job in other sensor manager algorithms. Might be worth changing the variable names here or in the others so that it's consistent.

spike-dstl (commented):
Code looks good. Was able to get an example running using `GreedySensorManager` that gave results comparable to `BruteForceSensorManager`. Left a couple picky comments that could be changed.

stonesoup/sensormanager/base.py
nperree-dstl: Good point, although some of the variables here are different to those in other sensor managers - perhaps "chosen_sensor_actions" rather than "all_sensor_actions" to make it clear that this is the dictionary of actions that have been selected for that sensor? You could also probably drop the "all_" from "all_selected_configs" to make it consistent with the Brute Force one.

nperree-dstl (approved):
Looks good. We should probably use a conftest.py file for the other sensor manager tests too but can change that separately.

stonesoup/sensormanager/base.py
```diff
@@ -150,3 +150,62 @@ def choose_actions(self, tracks, timestamp, nchoose=1, return_reward=False, **kw
...
+class GreedySensorManager(SensorManager):
+    """A sensor manager that returns a choice of actions from those available. Calculates
+    a reward function for each sensor in isolation. Selects the action that maximises reward
+    for each sensor.
+
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
```
sdhiscocks, line 164:
```suggestion
```

Issue #883, opened 2023-11-02
When `iter` is called on an iterator (Tracker), it shouldn't reset according to Python standard


gawebb-dstl:
This issue rarely is a problem, but it is when using heapq. (See example below)
```
from stonesoup.tracker import Tracker
import heapq

class MyTracker(Tracker):
    a_list = [0, 1, 2, 3, 4, 5]

    def __iter__(self):
        self.list_iter = iter(self.a_list)

        return super().__iter__()

    def __next__(self):
        return next(self.list_iter) ** 2

    def tracks(self):
        return self.a_list

print(list(MyTracker()))
print(list(heapq.merge(MyTracker(), MyTracker())))
```
Expected output:
``[0, 1, 4, 9, 16, 25]``
``[0, 0, 1, 1, 4, 4, 9, 9, 16, 16, 25, 25]``

Actual output:
``[0, 1, 4, 9, 16, 25]``
``[0, 0, 1, 1, 4, 4, 9, 9, 16, 16, 25, 25, 0, 1, 4, 9, 16, 25]``

PR #888, opened 2023-11-07
Progress Trackers Manually: Option C

gawebb-dstl:
All current trackers require a detection feeder. It can be useful to progress a tracker manually without a detection feeder (see Multi-Sensor Fusion: Covariance Intersection Using Tracks as Measurements example). 

All current trackers and start their `__next__` function with:
```
time, detections = next(self.detector_iter)
# Some tracking logic
```

I've moved this line to the new `Tracker`  baseclass and added an `update_tracker` function which takes the time and detections from the detection feeder as an input. With this being a separate function, you can access it directly and bypass the detection feeder which may be easier in instances.

This also fixes issue #883 

This is my third iteration of trying to get this right. See #742 and #779 for previous implementations.

### DIFF
--- a/docs/examples/Track2Track_Fusion_Example.py
+++ b/docs/examples/Track2Track_Fusion_Example.py
@@ -571,7 +571,7 @@ class DummyDetector(DetectionReader):
 
 # %%
 
-sensor1_detections, sensor2_detections = [], []
+sensor1_all_detections, sensor2_all_detections = [], []
 jpda_tracks, gmlcc_tracks = set(), set()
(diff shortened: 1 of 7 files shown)


gawebb-dstl:
The changes to docs/examples/Track2Track_Fusion_Example.py can probably be removed now #882 has been merged.

nperree-dstl (approved):

mharris-dstl (approved):

sdhiscocks:
Resolved in #980

PR #890, opened 2023-11-14
Gaussian process for tracking and bayesian optimisation for sensor managment 

Lyuchenyi:
Code from the University of Sheffield.


mharris-dstl:
A couple comments:

- there are multiple hundred FLAKE8 formatting errors that need fixing (my rough estimate is close to 1000). See https://app.circleci.com/pipelines/github/dstl/Stone-Soup/3275/workflows/e8f186bb-11fe-4392-939e-1982b08aa166/jobs/17963/tests
- Some of the file names need changing. Examples such as "visualization_2" are not descriptive as to what is happening. There is also a typo in another file name. 
- The tutorial is 23,00 lines long which is too big to even render on GitHub and bigger than all the tutorials combined. It is also in the same directory as the introductory tutorials and is not in the correct python format. You have a jupyter notebook but they are .py sphinx files.

Since this PR is a long way from being ready for a formal review, I recommend getting a couple of files at a time into the correct format for Stone Soup so the docs successfully build locally and all tests pass. Ensure the files have appropriate names and are in the correct directories/sub directories. Make a PR for those few files and then repeat so there are multiple, smaller PRs which can be more easily reviewed.

PR #892, opened 2023-11-16
Introduce grid based action generators and actionable movable classes

timothy-glover:
This pull request introduces actionable features to the movable module. A grid bases actionable movable object has been introduced which works with a corresponding grid based action generator. This enables platforms to move in a grid based fashion. Currently, the `NStepDirectionalGridActionGenerator` is introduced to allow movement in selected position axes according to a number of steps and step size. This is to be used with the `NStepDirectionalGridMovable`.

This pull request also introduces a new way for `Actionables` to interact with generators, making their usage clearer but this has not been implemented for actionable sensors yet.

Documentation has been created for each of the new classes and it has been incorporated into the documentation structure. This required that the movable module, which was previously not included, would also be included in the documentation.


stonesoup/movable/action/move_position_action.py
```diff
@@ -0,0 +1,89 @@
...
+class MovePositionAction(Action):
+    """This is the base class for an action that changes the
+    position of a platform or sensor."""
+
+    def act(self, current_time, timestamp, init_value):
+        return self.target_value
+
+
+class GridActionGenerator(ActionGenerator):
+    """This is the base class for generators that generate actions in a grid like fashion."""
```
nperree-dstl, line 19: Might be worth specifying in documentation that they generate a MovePositionAction, same for the generator below

stonesoup/movable/action/tests/test_move_position_action.py
```diff
@@ -0,0 +1,139 @@
+import pytest
+import numpy as np
+from datetime import datetime, timedelta
+import itertools as it
+import copy
+
+from ....types.state import StateVector, State, StateVectors
+from ....platform import FixedPlatform
+# from ..move_position_action import GridActionGenerator, NStepDirectionalGridActionGenerator
+from ...actionable_movable import NStepDirectionalGridMovable  # GridActionableMovable
```
nperree-dstl, line 10:
```suggestion
from ...actionable_movable import NStepDirectionalGridMovable
```

stonesoup/movable/action/move_position_action.py
```diff
@@ -0,0 +1,89 @@
...
+
+class MovePositionAction(Action):
+    """This is the base class for an action that changes the
+    position of a platform or sensor."""
+
+    def act(self, current_time, timestamp, init_value):
+        return self.target_value
+
+
+class GridActionGenerator(ActionGenerator):
```
nperree-dstl, line 18: Could we add a min and max method to the grid action generator (using the action space provided?) - this would allow it to be used with some of the optimised sensor managers we have.

timothy-glover: So would the min essentially return `current x position - n_steps*step_size*resolution` and max return `current x position + n_steps*step_size*resolution` ?

timothy-glover: I meant to remove the x, so it would be `current position +/- n_steps*step_size*resolution`

nperree-dstl (approved):
I'm happy that this all works as I'd expect and it's structured well. 

During testing I've noticed that the reward functions we have can either handle actionable sensors OR actionable platforms/movables. When you have an actionable platform with actionable sensors attached it doesn't work properly. I've raised this as an issue (#899) as it's probably best addressed separately (I'll probably take a look at it myself when I get a chance).

sdhiscocks (approved):

PR #907, opened 2023-12-12
Inertia navigation and landmarks tracking

A-acuto:
In this PR we present a set of measurement models regarding inertia navigation, in particular the measurement models from accelerometer and gyroscope on board of sensors. To use these measurement models a set of navigation functions have been added (including tests).
Summarising this PR consists in:

1.  Navigation functions and Euler angles calculations;
2.  Accelerometer measurement model
3.  Gyroscope Measurement model
4.  Azimuth-Elevation and Azimuth-Elevation-Range measurement model with fixed targets (for navigation calibration)
5.  an example of tracking using these measurement models.
6.  set of tests for the functions and measurement models.

### DIFF
--- a/stonesoup/functions/navigation.py
+++ b/stonesoup/functions/navigation.py
@@ -1,0 +1,469 @@
+"""
+Navigation functions
+--------------------
+"""
+
+import numpy as np
+from math import pi
+from . import localSphere2GCS
+
+def angle_wrap(angle):
+    """ Fix the angle if it gets over the threshold"""
+
+    return np.remainder(angle+pi, 2.*pi) - pi
+
+
+def earthSpeedFlatSq(dx, dy):
+    r"""Calculate the Earth speed flat vector respect to the reference
+    frame, from 2D Cartesian coordinates.
+
+    Parameters
+    ----------
+    dx : float, array like
+        :math:`dx` Earth velocity component
+
+    dy : float, array like
+        :math:`dy` Earth velocity component
+
+    Returns
+    -------
+    np.array: sum of the powers of dx and dy
+    """
+
+    return np.power(dx, 2) + np.power(dy, 2)
+
+
+def earthSpeedSq(dx, dy, dz):
+    r"""Calculate the Earth speed respect to the reference frame
+    from 3D Cartesian coordinates.
+
+    Parameters
+    ----------
+    dx : float, array like
+        :math:`dx` Earth velocity component
+
+    dy : float, array like
+        :math:`dy` Earth velocity component
+
+    dz : float, array like
+        :math:`dz` Earth velocity component
+
+    Returns
+    -------
+    np.array: sum of the powers of dx, dy and dz
+    """
+    return np.power(dx, 2) + np.power(dy, 2) + np.power(dz, 2)
+
+
+def earthSpeedFlat(dx, dy):
+    r"""Same as :class:`~.EarthSpeedFlatSq` but squared.
+
+    Parameters
+    ----------
+    dx : float, array like
+        :math:`dx` Earth velocity component
+
+    dy : float, array like
+        :math:`dy` Earth velocity component
+
+    Returns
+    -------
+    np.array: square root of the sum of the powers of dx and dy
+    """
+    return np.sqrt(np.power(dx, 2) + np.power(dy, 2))
+
+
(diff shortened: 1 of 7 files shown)


A-acuto:
In the comparison [example](docs/examples/comparison_filters_on_navigation.py) I was not able to use the metric manager (in particular the SIAP metric) because the track associator was trying to iterate the `GroundTruthState` (doing something like `next(state)` from the `GroundTruthPath` and it was failing. Therefore I used a less elegant solution to show the positional accuracy. 

Here is a gist that shows the problem:
https://gist.github.com/A-acuto/3c182d049e4df181e35c2a82307544fc

stonesoup/models/measurement/nonlinear.py
```diff
@@ -1400,3 +1401,346 @@ def rvs(self, num_samples=1, **kwargs) -> Union[StateVector, StateVectors]:
...
+            heading = np.array(np.radians(state.state_vector[9, ipoint])).reshape(-1, 1)
+            pitch = np.array(np.radians(state.state_vector[11, ipoint])).reshape(-1, 1)
+
+            # Transform the azimuths and elevation and fix for 180 degrees in case
+            azimuths[:, ipoint] = [Azimuth(angle_wrap(angle - heading[index])) for index, angle in
+                                   enumerate(absolute_azimuth)]
+            elevations[:, ipoint] = [Elevation(angle - pitch[index]) for index, angle in
+                                     enumerate(absolute_elevation)]
+
+        return StateVector([*azimuths, *elevations]) + noise
```
sdhiscocks, line 1621:
```suggestion
        # adjust the sensor location with the translation offset, if present
        sensor_location = state.state_vector[self.mapping, :] - self.translation_offset

        diff_position = self.target_location - sensor_location

        # evaluate the range
        rg = np.linalg.norm(diff_position, axis=0)

        # evaluate the absolute azimuth and elevation of the target respect to the sensor
        absolute_azimuth = np.arctan2(diff_position[1, :], diff_position[0, :])
        absolute_elevation = np.arcsin(diff_position[2, :] / rg)

        # evaluate the sensor heading and elevation
        heading = np.radians(state.state_vector[9, :])
        pitch = np.radians(state.state_vector[11, :])

        # Transform the azimuths and elevation and fix for 180 degrees in case
        azimuths = [Azimuth(angle) for angle in absolute_azimuth - heading]
        elevations = [Elevation(angle) for angle in absolute_elevation - pitch]

        return StateVectors([azimuths, elevations]) + noise
```

sdhiscocks (commented):
Had an initial look at the code. Main points: being consistent with radians, rather than degrees; and to double check if a different angle convention is being used; a few places where functions could be vectorised (included some suggestions but may not be best approach).

A-acuto:
@sdhiscocks thanks for the comments and suggestions, I'll apply them and double check the degree-radians transformations to keep the code uniform. As well, I'll check if PyMap3D does contain functions needed.

A-acuto:
In the last commits I answered the comments provided, in details:
- removed functions already present (e.g. angle_wrap -> mod_bearing)
- made extensive use of PyMap3D for distances and functions
- vectorisation of functions 
- switch from degrees to radians and alignment with stone soup rotation order
- grammar improvements and checks in the examples

The only point I am not sure about, but as presented works fine, is the use of function Build_rotation_matrix which assumes angles being 1-dimensional, which should be fine when dealing with Kalman filter tracking but not for Particle filter. However allowing it to be multi-dimensional (arrays of (3, N) dimension) is messing up the tracking with the particle filter.

stonesoup/functions/navigation.py
```diff
@@ -0,0 +1,384 @@
...
+
+    Reference
+    ---------
+    .. [#] M. Kok, J. Hol and T. Sch\"{o}n,
+           “Using Inertial Sensors for Position and Orientation Estimation”,
+           Foundations 456 and Trends in Signal Processing, Vol. 11, No. 1–2, pp 1–153, 2017.
+    """
+
+    # mapping
+    position_mapping = (0, 3, 6)
```
csherman-dstl, line 286: Should these mappings be arguments for generality?

csherman-dstl (commented):
I've left some comments generally on the wording in the examples and some PEP styling on function names. All functions and function variables in __init__.py and navigation.py should be named in snake_case.

PR #891, opened 2023-11-15
Dynamically Iterated EKF

spike-dstl:
This PR adds a new updater class: DynamicallyIteratedUpdater (DIEKF) - an implementation of algorithm 2: Dynamically iterated filter, from "Iterated Filters for Nonlinear Transition Models" by Anton Kullberg, Isaac Skog, and Gustaf Hendeby.

Additionally, this PR adds the following as required changes:

1.  Added compatibility to pre-existing Kullback-Leibler divergence measure for GaussianState types.
2. Added prior state as a property of Prediction states. This is required in order for the DIEKF to access and smooth the prior state
3. Updated all relevant Predictor classes to assign the prior state to the output Prediction state.

PR #913, opened 2023-12-20
PDA Updater

jmbarr:
Provides an updater based on probabilistic data association. See the documentation in the _Updaters_ section, or have a look at this gist: https://gist.github.com/jmbarr/92dc83e28c04026136d4f8706a1159c1

### DIFF
--- a/stonesoup/updater/probability.py
+++ b/stonesoup/updater/probability.py
@@ -47,7 +47,6 @@ class PDAUpdater(ExtendedKalmanUpdater):
         \mathbf{y}_{k,i}^T - \mathbf{y}_k \mathbf{y}_k^T \] K_k^T
 
     """
-
     def update(self, hypotheses, gm_method=False, **kwargs):
         r"""The update step.
 
--- a/stonesoup/updater/probability.py
+++ b/stonesoup/updater/probability.py
@@ -5,17 +5,20 @@ import scipy.linalg as la
 from functools import lru_cache
 
 from .kalman import ExtendedKalmanUpdater
+from ..types.update import Update
+from ..types.array import StateVectors
+from ..functions import gm_reduce_single
 
 
 class PDAUpdater(ExtendedKalmanUpdater):
     r"""An updater which undertakes probabilistic data association (PDA), as defined in [1]. It
     differs slightly from the Kalman updater it inherits from in that instead of a single
-    hypothesis object, the :meth:`update` method it takes a hypotheses object returned by a
+    hypothesis object, the :meth:`update` method takes a hypotheses object returned by a
     :class:`~.PDA` (or similar) data associator. Functionally this is a list of single hypothesis
     objects which group tracks together with associated measuments and probabilities.
 
     The :class:`~.ExtendedKalmanUpdater` is used in order to inherit the ability to cope with
-    (slight) non-linearities.
+    (slight) non-linearities. Other inheritance structures should be trivial to implement.
 
     The update step proceeds as:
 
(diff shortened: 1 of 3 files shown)


stonesoup/updater/probability.py
```diff
@@ -0,0 +1,192 @@
...
+                sum_of_innovations = float(hypothesis.probability) * innovation
+                sum_of_weighted_cov = float(hypothesis.probability) * (innovation @ innovation.T)
+            else:
+                sum_of_innovations += float(hypothesis.probability) * innovation
+                sum_of_weighted_cov += float(hypothesis.probability) * (innovation @ innovation.T)
+
+        posterior_mean += kalman_gain @ sum_of_innovations
+        posterior_covariance += \
+            kalman_gain @ (sum_of_weighted_cov - sum_of_innovations @ sum_of_innovations.T) \
+            @ kalman_gain.T
```
nperree-dstl, line 190: Do these lines work if line 162 is False? I can't work out where posterior_mean, posterior_covariance or kalman_gain would come from in that case.

nperree-dstl: Ignore me, I think I've got it now - the first hypothesis always a MissedDetection hypothesis right?

nperree-dstl (approved):

stonesoup/updater/probability.py
```diff
@@ -0,0 +1,192 @@
...
+            else:
+                innovation = hypothesis.measurement.state_vector - \
+                             hypothesis.measurement_prediction.state_vector
+
+            # probably exists a less clunky way of doing this using exists() or overwritten +=
+            # All these floats should be redundant if/when the bug in Probability.__mult__() is
+            # fixed.
+            if n == 0:
+                sum_of_innovations = float(hypothesis.probability) * innovation
+                sum_of_weighted_cov = float(hypothesis.probability) * (innovation @ innovation.T)
```
sdhiscocks, line 182: Set these two sum variables outside the loop to `0`?

sdhiscocks: Wasn't too sure if the proposed would work, but doing a simple test, looks like it should.
```
>>> a = 0
>>> a += np.array([[0], [0]])
>>> a
array([[0],
       [0]])
```

stonesoup/updater/probability.py
jmbarr: > the first hypothesis always a MissedDetection hypothesis right?
>
It seems to be but that's a matter for the `associator.associate()` method. I think this should work whatever the order; it exploits the fact that `not hypothesis` returns True for a missed detection (not sure why that is) and calculates the required mean and covariance, and attaches a 0-innovation.

stonesoup/updater/probability.py
jmbarr: So, 
`sum_of_innovations = 0
sum_of_innovations += hypothesis.probability * innovation`
?

jmbarr: > So, `sum_of_innovations = 0` `sum_of_innovations += hypothesis.probability * innovation` ?

But won't matter if :class:`Probability` is corrected.

sdhiscocks:
```diff
diff --git a/stonesoup/updater/probability.py b/stonesoup/updater/probability.py
index c471e514..c6495111 100644
--- a/stonesoup/updater/probability.py
+++ b/stonesoup/updater/probability.py
@@ -156,6 +156,8 @@ class PDAUpdater(ExtendedKalmanUpdater):
             The covariance of the reduced/single Gaussian
         """
 
+        sum_of_innovations = 0
+        sum_of_weighted_cov = 0
         for n, hypothesis in enumerate(hypotheses):
             # Check for the existence of an associated measurement. Because of the way the
             # hypothesis is constructed, you can do this:
@@ -174,15 +176,10 @@ class PDAUpdater(ExtendedKalmanUpdater):
                 innovation = hypothesis.measurement.state_vector - \
                              hypothesis.measurement_prediction.state_vector
 
-            # probably exists a less clunky way of doing this using exists() or overwritten +=
             # All these floats should be redundant if/when the bug in Probability.__mult__() is
             # fixed.
-            if n == 0:
-                sum_of_innovations = float(hypothesis.probability) * innovation
-                sum_of_weighted_cov = float(hypothesis.probability) * (innovation @ innovation.T)
-            else:
-                sum_of_innovations += float(hypothesis.probability) * innovation
-                sum_of_weighted_cov += float(hypothesis.probability) * (innovation @ innovation.T)
+            sum_of_innovations += float(hypothesis.probability) * innovation
+            sum_of_weighted_cov += float(hypothesis.probability) * (innovation @ innovation.T)
 
         posterior_mean += kalman_gain @ sum_of_innovations
         posterior_covariance += \
```

sglvladi:
Looking at this made me think whether we could make this more modular by using a `KalmanUpdater` (or subclass) via composition, rather than inheritance:
- `PDAUpdater` inherits from `Updater` and accepts a `KalmanUpdater` as an `updater` property, which it then uses to either `update()` or `predict_measurement()`, as per lines [127](https://github.com/dstl/Stone-Soup/blob/e2edd269f89c669e90af0a8008409bffaa7525b0/stonesoup/updater/probability.py#L127) and [164](https://github.com/dstl/Stone-Soup/blob/e2edd269f89c669e90af0a8008409bffaa7525b0/stonesoup/updater/probability.py#L163)
- The class could also be renamed to `KalmanPDAUpdater`/`GaussianPDAUpdater` to reflect the fact that it applies specifically to Kalman/Gaussian cases.

The above would pave the way for a `ParticlePDAUpdater`/`ELPFUpdater` that could be structured similarly to perform the PDA update for particle filters.

sdhiscocks:
> Looking at this made me think whether we could make this more modular by using a `KalmanUpdater` (or subclass) via composition, rather than inheritance:
> 
> * `PDAUpdater` inherits from `Updater` and accepts a `KalmanUpdater` as an `updater` property, which it then uses to either `update()` or `predict_measurement()`, as per lines [127](https://github.com/dstl/Stone-Soup/blob/e2edd269f89c669e90af0a8008409bffaa7525b0/stonesoup/updater/probability.py#L127) and [164](https://github.com/dstl/Stone-Soup/blob/e2edd269f89c669e90af0a8008409bffaa7525b0/stonesoup/updater/probability.py#L163)
> * The class could also be renamed to `KalmanPDAUpdater`/`GaussianPDAUpdater` to reflect the fact that it applies specifically to Kalman/Gaussian cases.
> 
> The above would pave the way for a `ParticlePDAUpdater`/`ELPFUpdater` that could be structured similarly to perform the PDA update for particle filters.

Something similar was done for the DIEKF in #891: https://github.com/dstl/Stone-Soup/blob/e27fc3639d791d2441b6c4b8572c57932651deb3/stonesoup/updater/iterated.py

jswright-dstl (approved):

sdhiscocks:
Per @sglvladi comment, I think it would be good to change structure to support more flexible structure, but I'll merge this for now.

PR #927, opened 2024-01-09
Simplify (or at least make more standard) control models

jmbarr:
Problem: current control models, both in their stand-alone state and as part of the predictor, are somewhat peculiar (that's my fault). This change is designed to make them look a bit more like other models, and so hopefully make their use more intuitive. 

See here for an example of use: https://gist.github.com/jmbarr/1f9cfb06eea9dd051bb7adea9f75d3a8

Should have implications for sensor and platform management.

### DIFF
--- a/stonesoup/models/control/linear.py
+++ b/stonesoup/models/control/linear.py
@@ -20,7 +20,6 @@ class LinearControlModel(ControlModel, LinearModel, GaussianModel):
 
     """
 
-    # control_vector: np.ndarray = Property(doc="Control vector at time :math:`k`")
     control_matrix: np.ndarray = Property(
         doc="Control input model matrix at time :math:`k`, :math:`B_k`")
     control_noise: np.ndarray = Property(
--- a/stonesoup/models/control/linear.py
+++ b/stonesoup/models/control/linear.py
@@ -2,16 +2,17 @@ import numpy as np
 from scipy.stats import multivariate_normal
 
 from .base import ControlModel
-from ..base import LinearModel
+from ..base import LinearModel, GaussianModel
 from ...base import Property
+from ...types.array import StateVector
 
 
-class LinearControlModel(ControlModel, LinearModel):
+class LinearControlModel(ControlModel, LinearModel, GaussianModel):
     r"""Implements a linear effect to the state vector via,
 
     .. math::
 
-        \hat{x}_k = B_k \mathbf{u}_k + \gamma_k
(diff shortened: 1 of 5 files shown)


stonesoup/predictor/kalman.py
```diff
@@ -167,9 +167,11 @@ def predict(self, prior, timestamp=None, **kwargs):
             :math:`\mathbf{x}_{k-1}`
         timestamp : :class:`datetime.datetime`, optional
             :math:`k`
+        control_input : :class:`StateVector`, optional
+            :math:`\mathbf{u}_k`
         **kwargs :
             These are passed, via :meth:`~.KalmanFilter.transition_function` to
-            :meth:`~.LinearGaussianTransitionModel.matrix`
+            :meth:`~.LinearGaussianTransitionModel.matrix` and
```
jswright-dstl, line 174: and what?

stonesoup/models/control/linear.py
```diff
@@ -20,68 +20,57 @@ class LinearControlModel(ControlModel, LinearModel):
 
     """
 
-    control_vector: np.ndarray = Property(doc="Control vector at time :math:`k`")
+    # control_vector: np.ndarray = Property(doc="Control vector at time :math:`k`")
```
jswright-dstl, line 23: Is this commented line needed? If not, can it be removed?

jswright-dstl (approved):

nperree-dstl (approved):

stonesoup/predictor/kalman.py
```diff
@@ -326,17 +329,19 @@ def _transition_and_control_function(self, prior_state, **kwargs):
...
         r"""The unscented version of the predict step
 
         Parameters
         ----------
         prior : :class:`~.State`
             Prior state, :math:`\mathbf{x}_{k-1}`
+        control_input: :class:`~.State`
+            Control input vector, :math:`\mathbf{u}_k`
         timestamp : :class:`datetime.datetime`
             Time to transit to (:math:`k`)
```
sdhiscocks, line 346: Minor change to be consistent with other `predict` methods argument orders.

```suggestion
    def predict(self, prior, timestamp=None, control_input=None, **kwargs):
        r"""The unscented version of the predict step

        Parameters
        ----------
        prior : :class:`~.State`
            Prior state, :math:`\mathbf{x}_{k-1}`
        timestamp : :class:`datetime.datetime`
            Time to transit to (:math:`k`)
        control_input: :class:`~.State`
            Control input vector, :math:`\mathbf{u}_k`
```

PR #936, opened 2024-01-18
remove unnecessary typo

avcuenes:


mharris-dstl:
https://stackoverflow.com/questions/33729045/what-does-an-r-represent-before-a-string-in-python

r before string means string is treated as a raw string, affecting the use of \. The tests show a formatting error kicks up in stonesoup/dataassociator/probability.py, suggesting that this is not a typo and comes into use. Two options as I see it:

1. Update stonesoup/dataassociator/probability.py to fix this formatting error
2. Close the PR (less effort)

If you disagree, I'm happy to hear your reasoning.

PR #939, opened 2024-01-19
Adding an example showing how to deal with OOSM using a buffer of measurements

A-acuto:
This PR adds an example that shows how to use Algorithm A [1] to deal with Out of sequence measurements (OOSM). 
This example together with other PRs  (#915) aims to explain various techniques to deal with OOSM, in this case we consider a buffer of measurements and we re-order them to fix for any delay in their arrival. 

[1]  Y. Bar-Shalom, M. Mallick, H. Chen, R. Washburn, 2002, One-step solution for the general out-of-sequence measurement problem in tracking, Proceedings of the 2002 IEEE Aerospace Conference.

PR #915, opened 2023-12-22
Example of Ensemble Kalman filter with Out-Of-Sequence measurements

A-acuto:
This PR adds a simple example of using a Kalman Filter with measurements with a fixed delay from a sensor.

As referenced by #940 and #939 , this is a series of examples that shows different methods to deal with OOSM. 
The example has been modified by adding a EK filter for comparison.

PR #940, opened 2024-01-19
Adding an example showing inverse-time dynamics method to deal with OOSM 

A-acuto:
Thir PR adds an example that shows how to implement Algorithm C [1, 2] to deal with out of sequence measurements. 
As other examples (see PR #939 and PR #915 ) we explain how to treat OOSM, in this specific case we employ a time-inverse dynamics method to create pseudo-measurements at "correct timestamp" from the delayed detections.

[1]  Y. Bar-Shalom, M. Mallick, H. Chen, R. Washburn, 2002, One-step solution for the general out-of-sequence measurement  problem in tracking, Proceedings of the 2002 IEEE Aerospace Conference.
[2] S. R. Maskell, R. G. Everitt, R. Wright, M. Briers, 2005, Multi-target out-of-sequence data association: Tracking using graphical models, Information Fusion.

### DIFF
--- a/docs/examples/example_oosm_algorithm.py
+++ b/docs/examples/example_oosm_algorithm.py
@@ -54,8 +54,6 @@ Algorithm dealing with Out-of-Sequence measurements
 # ^^^^^^^^^^^^^^^
 from datetime import datetime, timedelta
 import numpy as np
-import matplotlib.pyplot as plt
-from ordered_set import OrderedSet
 from copy import deepcopy
 
 # Simulation parameters
@@ -350,7 +348,7 @@ plotter.plot_tracks(oosm_tracks, [0, 2], track_label='OOSM Tracks',
                     line= dict(color='orange'))
 plotter.plot_tracks(noOsm_tracks, [0, 2], track_label='no-OOSM Tracks',
                     line= dict(color='red'))
-plotter.fig
+plotter.show()
 
 # %%
 # Evaluate the track-to-truth accuracy
@@ -409,4 +407,3 @@ graph.fig
 # .. [#] S. R. Maskell, R. G. Everitt, R. Wright, M. Briers, 2005,
 #        Multi-target out-of-sequence data association: Tracking using
 #        graphical models, Information Fusion.
-#
--- a/docs/examples/example_oosm_algorithm.py
+++ b/docs/examples/example_oosm_algorithm.py
@@ -15,20 +15,20 @@ Algorithm dealing with Out-of-Sequence measurements
 # As shown in literature (e.g., [#]_) there are
 # multiple approaches on how to deal with OOS measurements,
 # spanning from simply ignoring them, assuming that the fraction
-# of these is small and it will not significally impact the quality of the
+# of these is small and it will not, significally, impact the quality of the
 # tracking, iterate over last $\math{\ell}$ measurements with
 # a fixed lag (see what it is called Algorithm A and B in [#]_ and
-# [#]_, see also the previous examples)
-# or, like in this example, you can include any OOSM
-# and re-process the measurement using inverse-time dynamics creating
-# pseudo-measurements. This approach is called in literature as Algorithm C
+# [#]_, see also the previous examples) or, like in this example,
+# you can include any OOSM and re-process the measurement using
+# inverse-time dynamics creating pseudo-measurements.
+# This approach is called in literature as Algorithm C
 # (from [#]_, [#]_) and it will be refered as such later on.
 # To explain how does it work let's consider a single target scenario,
 # this algorithm deals with the presence of delayed measurements by
 # using the predicted dynamics of the target to go back in time from
 # the actual arrival time and obtain a pseudo-measurement obtained from
-# simulating the measurement at the "expected" scan time.
-# In this manner we can reconstruct the measurement chain and not
+# simulating the detection at the "expected" scan time.
+# In this manner we can reconstruct the measurements chain and not
 # discard any information, as well, we don't need to store a large fix-lag
 # distribution of data (as can happen with other algorithms).
 # In this example, we focus our efforts in showing how to use the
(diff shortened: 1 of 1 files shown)


docs/examples/example_oosm_algorithm.py
```diff
@@ -0,0 +1,409 @@
...
+# %%
+# 3) Run the tracker and apply the algorithm on delayed measurements;
+# -------------------------------------------------------------------
+# We can now run the tracker and generate the tracks.
+# By looping over the scans we can spot any OOS measurement and apply the algorithm.
+# When we encounter a delayed measurement, at time $\math{\tau}$, we make use of both
+# the measurement model and transition model: first we use the measurement model
+# inverse function at $\math{\tau}$ to obtain a predicted Cartesian location of the
+# target, then we use the transition model with time-inverse dynamics ($t_{k} - \math{\tau}$)
+# to trace back where the target was in the scan-timestamp ($t$). With this pseudo-location, which is an
```
spike-dstl, line 252:
```suggestion
# target, then we use the transition model with time-inverse dynamics (:math:`t_{k} - \tau`)
# to trace back where the target was in the scan-timestamp :math:`t`. With this pseudo-location, which is an
```

docs/examples/example_oosm_algorithm.py
```diff
@@ -0,0 +1,409 @@
...
+# approximation, we can compute the pseudo-measurement usign the measurement model.
```
spike-dstl, line 253:
```suggestion
# approximation, we can compute the pseudo-measurement using the measurement model.
```

docs/examples/example_oosm_algorithm.py
```diff
@@ -0,0 +1,409 @@
...
+# In this way now we have a pseudo-measurement at $t$ and we can keep the time-chain and
```
spike-dstl, line 254:
```suggestion
# In this way now we have a pseudo-measurement at :math:`t` and we can keep the time-chain and
```

docs/examples/example_oosm_algorithm.py
```diff
@@ -0,0 +1,409 @@
...
+                        post = updater.update(hyp)
+                        track.append(post)
+
+
+# %%
+# 4) Run the comparison between the resulting tracks and plot the results.
+# ------------------------------------------------------------------------
+# We have obtained the final tracks from the detections and we have two sets of
+# tracks, one with OOSM and one without. We now can visualise the results
+# and evaluate the track-to-truh accuracy using the OSPA metric tool.
```
spike-dstl, line 339:
```suggestion
# and evaluate the track-to-track accuracy using the OSPA metric tool.
```

spike-dstl (changes requested):
Mostly grammatical changes. Please also see comment on line 352 for a suggested change to help present the results more effectively.

spike-dstl (approved):

docs/examples/example_oosm_algorithm.py
```diff
@@ -0,0 +1,405 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+===================================================
+Algorithm dealing with Out-of-Sequence measurements
```
nperree-dstl, line 6: Could you make this title more specific to distinguish from other OOSM examples? E.g. "Handling OOSM using inverse time dynamics"

docs/examples/example_oosm_algorithm.py
```diff
@@ -0,0 +1,405 @@
...
+# to go back in time from the actual arrival time and obtain a pseudo-measurement obtained from
+# simulating the detection at the "expected" scan time.
+# In this manner we can reconstruct the measurements chain and not discard any information, as well,
+# we don't need to store a large fix-lag distribution of data (as can happen with other algorithms).
+#
+# In this example, we focus our efforts in showing how to use the algorithm C in a multi-target scenario with clutter.
+# We simulate two sensors obtaining scans of two objects travelling with a nearly constant velocity transition model
+# and with some level of clutter, at specific timesteps we insert a delay in one of the scan from the sensor and we
+# employ Algorithm C to process in the correct way the chain of measurements.
+#
```
nperree-dstl, line 35:
```suggestion
# As shown in literature [1]_ there are multiple approaches on how to deal with OOS measurements.
# Simply ignoring them, assuming that the fraction of these is small and it will not
# significantly impact the quality of the tracking; iterating over last :math:`\ell` measurements with
# a fixed lag (see what it is called Algorithm A and B in [2]_ and [3]_, and the other OOSM examples);
# or, like in this example, you can include any OOSM and re-process the measurement using
# inverse-time dynamics creating pseudo-measurements.
#
# This approach is known in literature as Algorithm C [2]_ [3]_ and it will be referred to as such later on.
#
# To explain how it works, we consider a single target scenario.
# This algorithm deals with the presence of delayed measurements by using the predicted dynamics of the target
# to go back in time from the actual arrival time and obtain a pseudo-measurement from
# simulating the detection at the "expected" scan time.
# In this manner we can reconstruct the measurements chain without discarding any information, and
# we don't need to store a large fix-lag distribution of data (as can happen with other algorithms).
#
# In this example, we focus our efforts in showing how to use the algorithm C in a multi-target scenario with clutter.
# We simulate two sensors obtaining scans of two objects travelling with a nearly constant velocity transition model
# and with some level of clutter. At specific timesteps we insert a delay in one of the scans from the sensor and we
# employ Algorithm C to process in the correct way the chain of measurements.
#
```

docs/examples/example_oosm_algorithm.py
```diff
@@ -0,0 +1,405 @@
...
+from stonesoup.plotter import AnimatedPlotterly
+plotter = AnimatedPlotterly(timesteps=timestamps)
+plotter.plot_ground_truths(truths, [0, 2])
+plotter.plot_measurements(scan_s1, [0, 2], measurements_label='scan1', measurement_model=sensor_1_mm)
+plotter.plot_measurements(scan_s2, [0, 2], measurements_label='scan2', measurement_model=sensor_1_mm)
+plotter.plot_tracks(oosm_tracks, [0, 2], track_label='OOSM Tracks',
+                    line= dict(color='orange'))
+plotter.plot_tracks(noOsm_tracks, [0, 2], track_label='no-OOSM Tracks',
+                    line= dict(color='red'))
+plotter.fig
```
nperree-dstl, line 347: The `TrueDetection`s plotted in your output (blue points) look pretty noisy and a long way from the ground truth towards the end of the simulation. Ignore me if that's intentional, but might be worth adjusting the noise in your measurement model if not.

nperree-dstl (changes requested):
Some suggestions to improve grammar and readability but otherwise looks good!

nperree-dstl (approved):

Issue #899, opened 2023-11-30
RewardFunction classes need adapting to handle actionable platforms and sensors


nperree-dstl:
The RewardFunction classes we have can only handle either actionable sensors, or actionable platforms - they will not work properly when an actionable platform has an actionable sensor attached.

PR #943, opened 2024-01-30
Sensor management with actionable movables

akenyon:
Making some changes to sensor management to be compatible with new actionable movable platforms.
Key changes:
- base actionable class moved from sensor to sensormanager directory
- platforms now also inherit from actionable
- sensor manager changed to take separate sensor and platform sets and work with any combination of actionable/non-actionable sensors/platforms


stonesoup/movable/actionable_movable.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
```
nperree-dstl, line 2: We might be able to remove the Actionable.register() line in this file anyway as FixedMovable will inherit from MovableActionable (and therefore Actionable).

stonesoup/sensormanager/base.py
```diff
@@ -34,6 +41,24 @@ class SensorManager(Base, ABC):
                           "vector in the case of multi-objective optimisation. Metrics may be of "
                           "any type and in any units.")
 
+    take_sensors_from_platforms: bool = Property(
```
nperree-dstl, line 44: Quite a long variable name, could we call in "platform_sensors" and explain it's functionality in the doc string?

stonesoup/sensormanager/base.py
```diff
@@ -34,6 +41,24 @@ class SensorManager(Base, ABC):
...
+        default=True, doc="Whether to update the sensor set with any sensors that are on the "
+                          "platform(s) but not already in the sensor set. Any sensors not added "
+                          "to the sensor set will not be considered by the sensor manager or "
+                          "reward function.")
```
nperree-dstl, line 48:
```suggestion
        default=True, doc="Whether to include sensors that are on the "
                          "platform(s) but not explicitly passed to the sensor manager. "
                          "Any sensors not added "
                          "will not be considered by the sensor manager or "
                          "reward function.")
```

stonesoup/movable/actionable.py
```diff
@@ -0,0 +1,9 @@
+from ..sensormanager.actionable import Actionable
+
+
+class MovableActionable(Actionable):
+    """Movable Actionable base class
+
+    Base class for actionable sensors which has all the functionality of base
+    :class:`~.Actionable` class, with additional platform-specific features.
+    """
```
nperree-dstl, line 9: I don't love that we now have `movable.actionable` and `movable.actionable_movable` modules... 

I think it makes sense in terms of the code because this class is the `MovableActionable` base class whereas the other one is a `Movable` which has some of that actionable functionality but the naming is a bit confusing. Would renaming the `actionable_movable` module be too much of a headache @sdhiscocks?

stonesoup/sensormanager/actionable.py
```diff
@@ -0,0 +1,277 @@
...
+    @property
+    @abstractmethod
+    def min(self):
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def max(self):
+        raise NotImplementedError
+
```
nperree-dstl, line 94: Might be worth moving these into a separate `sensormanager.action` file, in line with file structure under `sensor` and `movable` sections?

nperree-dstl: I'm probably just being picky now though...

stonesoup/movable/actionable.py
nperree-dstl: It also needs adding to stonesoup.movable.rst so it appears in the docs

stonesoup/sensormanager/actionable.py
```diff
@@ -0,0 +1,277 @@
+import datetime
```
nperree-dstl, line 1: File needs adding to the sensormanager.rst file so it appears in the docs

stonesoup/sensormanager/base.py
```diff
@@ -23,8 +27,11 @@ class SensorManager(Base, ABC):
     which communicate with other sensor managers in a networked fashion.
 
     """
-    sensors: Set[Sensor] = Property(doc="The sensor(s) which the sensor manager is managing. "
-                                        "These must be capable of returning available actions.")
+    sensors: Set['Sensor'] = Property(
+        default=set(), doc="The sensor(s) which the sensor manager is managing.")
+
+    platforms: Set['Platform'] = Property(
+        default=set(), doc="The platform(s) which the sensor manager is managing.")
```
nperree-dstl, line 34:
```suggestion
    sensors: Set['Sensor'] = Property(
        default=None, doc="The sensor(s) which the sensor manager is managing.")

    platforms: Set['Platform'] = Property(
        default=None, doc="The platform(s) which the sensor manager is managing.")
```

These need to be default=None and then can create empty sets in the init function

stonesoup/sensor/tests/test_actionable.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 
 from ..action.dwell_action import DwellActionsGenerator, ChangeDwellAction
-from ..actionable import Actionable, ActionableProperty
+from stonesoup.sensormanager.action import Actionable, ActionableProperty
```
sdhiscocks, line 7: Ideally should be relative like others.

stonesoup/sensormanager/base.py
```diff
@@ -34,6 +41,24 @@ class SensorManager(Base, ABC):
...
+                          "platform(s) but not already in the sensor set. Any sensors not added "
+                          "to the sensor set will not be considered by the sensor manager or "
+                          "reward function.")
+
+    @sensors.getter
+    def sensors(self):
+        sensors = self._property_sensors
+        if self.take_sensors_from_platforms:
+            for platform in self.platforms:
+                sensors.update(platform.sensors)
```
sdhiscocks, line 55: Missing on coverage. Test to cover this would be good.

sdhiscocks (approved):

stonesoup/sensormanager/base.py
sdhiscocks: I've approved PR, but based on assumption this will be fixed 👍

stonesoup/sensormanager/base.py
```diff
@@ -34,6 +41,33 @@ class SensorManager(Base, ABC):
...
+                          "will not be considered by the sensor manager or "
+                          "reward function.")
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if self.platforms is None:
+            self.platforms = set()
+        if self.sensors is None:
+            self.sensors = set()
```
nperree-dstl, line 57:
```suggestion
        if self._property_sensors is None:
            self._property_sensors = set()
```

Otherwise it tries to call the sensors.getter and tries to update a NoneType object

nperree-dstl (changes requested):

nperree-dstl (approved):
Closes #899

stonesoup/sensormanager/base.py
```diff
@@ -34,6 +41,33 @@ class SensorManager(Base, ABC):
...
+        super().__init__(*args, **kwargs)
+
+        if self.platforms is None:
+            self.platforms = set()
+        if self._property_sensors is None:
+            self._property_sensors = set()
+
+    @sensors.getter
+    def sensors(self):
+        sensors = self._property_sensors
```
sdhiscocks, line 61: Just thinking now, you probably shouldn't modify the original set, in case you remove a platform from the sensor manager for example.
```suggestion
        sensors = self._property_sensors.copy()
```

PR #945, opened 2024-01-31
Fix incorrect transition model being used in Kalman Smoothers

### LINKED ISSUE #941: Smoothing gain in KF smoother
I believe there is a typo in how the smoothing gain is implemented for the KF smoother:
https://github.com/dstl/Stone-Soup/blob/0a9ebe2584eb6cd3d77756fd9dfa7b98649799fd/stonesoup/smoother/kalman.py#L131-L148

Basically, the function aims to return the definition [1]
$$G_{k} = P_{k} F_{k}^{T} (P_{k+1|k})^{-1},$$
but the implementation instead evaluates this:
$$G_{k} = P_{k} F_{k-1}^{T} (P_{k+1|k})^{-1}.$$

Specifically, the second term `self._transition_matrix(state, **kwargs)` responsible for $F_{k}$ obtains it from the current `state` by looking into `state.hypothesis.prediction`, i.e., a prediction from $k-1$ to $k$, which contains $F_{k-1}$ and not $F_{k}$. However, if the function was implemented using `prediction`, which describes the prediction from $k$ to $k+1$, like this ``self._transition_matrix(prediction, **kwargs)``, it would actually access $F_{k}$ as required by the definition.

[1] See Algorithm 1 in https://livrepository.liverpool.ac.uk/3015339/1/PL_smoothing_accepted1.pdf

sdhiscocks:
Fixes #941

### DIFF
--- a/stonesoup/predictor/kalman.py
+++ b/stonesoup/predictor/kalman.py
@@ -229,6 +229,8 @@ class ExtendedKalmanPredictor(KalmanPredictor):
         ----------
         prior : :class:`~.State`
             :math:`\mathbf{x}_{k-1}`
+        linearisation_point : :class:`~State`, optional
+            State to linearise over. Default `None` where prior will be used.
         **kwargs : various, optional
             These are passed to :meth:`~.TransitionModel.matrix` or
             :meth:`~.TransitionModel.jacobian`
--- a/stonesoup/smoother/kalman.py
+++ b/stonesoup/smoother/kalman.py
@@ -5,13 +5,13 @@ import numpy as np
 
 from .base import Smoother
 from ..base import Property
+from ..models.base import LinearModel
 from ..types.multihypothesis import MultipleHypothesis
 from ..types.prediction import GaussianStatePrediction
 from ..types.update import GaussianStateUpdate
 from ..models.transition.base import TransitionModel
 from ..models.transition.linear import LinearGaussianTransitionModel
 from ..functions import gauss2sigma, unscented_transform
-from ..predictor.kalman import ExtendedKalmanPredictor
 
 
 class KalmanSmoother(Smoother):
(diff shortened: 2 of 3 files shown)


jswright-dstl (approved):

spike-dstl (approved):

sdhiscocks:
@narykov Does this resolve #941?

narykov:
> @narykov Does this resolve #941?

Apologies I didn't get back to this earlier. I can confirm this has resolved the issue. Thank you!

narykov:
> > @narykov Does this resolve #941?
> 
> Apologies I didn't get back to this earlier. I can confirm this has resolved the issue. Thank you!

Perhaps, the only additional comment I have is that in my case each prediction in the track is accompanied by its own `transition_model`, so I find it puzzling that I am forced to specify a single transition model when calling the smoother. As a consequence, I had to specify some dummy model that the smoother end up never using.

sdhiscocks:
> Apologies I didn't get back to this earlier. I can confirm this has resolved the issue. Thank you!

Thanks for confirming.

> Perhaps, the only additional comment I have is that in my case each prediction in the track is accompanied by its own `transition_model`, so I find it puzzling that I am forced to specify a single transition model when calling the smoother. As a consequence, I had to specify some dummy model that the smoother ends up never using.

So hopefully setting it to `None` should work okay? But maybe we should make it default `None` so it's optional?

narykov:
> So hopefully setting it to `None` should work okay? But maybe we should make it default `None` so it's optional?

Setting `transition_model=None` works for me, just it hasn't crossed my mind that there is that possibility. And as for the second question, I am thinking that it is reasonable to have `transition_model` explicitly specified as optional because:

- The supplied `transition_model` is primarily used as a fallback option, which won't be used unless the track updates lack their own models. Furthermore, one might mistakenly believe that it is the supplied model that is employed for processing, whereas another model is actually implemented inside the smoother.

- I believe that one could obtain a `Track` object without providing a reference to any `transition_model` when the updates are constructed using the `from_state` method. However, this is by no means a common way of using Stone Soup.

Thanks!

narykov:
> The supplied transition_model is primarily used as a fallback option, which won't be used unless the track updates lack their own models. Furthermore, one might mistakenly believe that it is the supplied model that is employed for processing, whereas another model is actually implemented inside the smoother.

Perhaps, if `transition_model` is specified explicitly in the definition of smoother it should be enforced by the smoother? As this appears to me as the expected behaviour.

PR #886, opened 2023-11-06
Reinforcement Learning Sensor Manager Example

sgboakes:
Instead of adding a new reinforcement learning sensor manager (as in #760), this adds an example of how to interface an RL framework with Stone Soup. This reduces the repo size, and allows users more freedom to adapt to their own problems. This PR includes the example, and the associated doc results to reduce doc compilation time.

PR #952, opened 2024-02-07
Adding an example that shows linearised ODE as transition model

A-acuto:
This PR adds an example that shows how to create linearised transition models from non-linear time continuing dynamics using Van Loan methods. 

A more detailed implementation of linearisation method will be available in a future PR coming from a different work package.

### DIFF
--- a/docs/examples/ODE_example.py
+++ b/docs/examples/ODE_example.py
@@ -1,0 +1,367 @@
+#!/usr/bin/env python
+
+"""
+==================================================================
+Using linearised ODEs from non-linear dynamic models in Stone Soup
+==================================================================
+"""
+
+# %%
+# In real world applications objects are subject to much more complex
+# dynamics than the simpler approximations we have
+# considered in other examples. However, working with non-linear
+# and time-variant dynamics can be challenging and computationally
+# expensive, hence the needs of approximation methods to ease
+# the simulation and the tracking performance.
+# In this example we present a simple method to adopt when you have to linearise
+# a non-linear dynamical model and use such linearised method in a
+# Stone soup application acting as transition model.
+# Specifically in this example we will show how to use the Van Loan method [#]_
+# to linearise the gravitational forces acting on a satellite orbiting Earth
+# and create a transition model to model the target trajectory, then making use of
+# standard Stone Soup components perform the tracking.
+# This method can be used in other context such as navigation.
+#
+# To linearise the dynamical functions, OrdinaryOrdinary Differential Equations (ODEs), we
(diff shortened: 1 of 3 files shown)


docs/examples/ODE_example.py
```diff
@@ -0,0 +1,367 @@
...
+
+        # define the dimension and timedelta
+        nx = self.ndim_state
+        dt = time_interval.total_seconds()
+
+        # Get Q
+        q_xdot, q_ydot, q_zdot = self.linear_noise_coeffs
+        dQ = np.diag([0., q_xdot, 0., q_ydot, 0., q_zdot])
+
+        G = expm(dt * np.block([[-self.dA, dQ],
```
sdhiscocks, line 204: Not ideal having `self.dA` here as it may not be set. But can see issue in that you need the `state` here in order to calculate it. Not sure what solution would be for this...

docs/examples/ODE_example.py
```diff
@@ -0,0 +1,367 @@
...
+for measurement in measurements:
+    prediction = predictor.predict(prior, timestamp=measurement.timestamp)
+    hypothesis = SingleHypothesis(prediction, measurement)
+    post = updater.update(hypothesis)
+    track.append(post)
+    prior = track[-1]
+
+# Now visualise the results
+plotter.plot_tracks(track, [0, 2, 4], track_label='Track')
+plotter.fig
```
sdhiscocks, line 353: The results look like track following very closely to the measurements. Is this what you'd expect? An issue calculating Q??

image

A-acuto:
Hi @sdhiscocks , thanks for the comments. 
I checked the Q matrix and I think I got an index wrong, now the track is guided by the measurements (which are pretty noisy) but it is not following as before.

Regarding your comment on the `self.dA` : I am not convinced on how to approach this issue. I might try simpler scenarios (eg smaller scales) and see if `self.dA` creates any issue.

For the stone soup installation I followed a recent example PR #886 for the clean installation, however if you feel like something different I can amend the current example.

simonrmaskell:
Would it be useful for this PR to be augmented with another example, eg involving the model in this paper: https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=6290578. The paper has a non-trivial analytic result that we could check against: Van Loan wouldn't be expected to get the same answer (since it is unaware of the trigonometric identities used in the paper and uses a Taylor series in state to approximate the ODE and then solves the approximate ODE exactly)?

A-acuto:
Closing this PR, while developing the new ODE example suggested by @simonrmaskell

PR #966, opened 2024-03-07
Correct calculation of net orientation of PlatformMountables

edwheelhouse-dstl:
Calculation of the net orientation of a PlatformMountable on a Platform looks incorrect for 3D cases. It adds the platform's (MovementController's) orientation to the sensor's rotation_offset. Think of a sensor oriented to the left when its platform pitches up: addition would orient the sensor up too, when it should just roll. Instead, this PR rotates a vector aligned to the x-axis by the platform's orientation and the sensor's rotation_offset. This requires rotations in order: x, y, z. The current functions.init.build_rotation_matrix() does it the other way round (as required for other use cases) so a new build_rotation_matrix_xyz() has been added. Tests have been updated to match: rotation instead of addition requires orientation about x and z to be within +/-pi and in y to be within +/-pi/2.

### DIFF
--- a/stonesoup/sensor/base.py
+++ b/stonesoup/sensor/base.py
@@ -118,8 +118,8 @@ class PlatformMountable(Base, ABC):
         if self.movement_controller is None:
             return None
         x_axis = (1, 0, 0)
-        rotmat = build_rotation_matrix_xyz(-self.movement_controller.orientation) @ \
-            build_rotation_matrix_xyz(-self.rotation_offset)
+        rotmat = (build_rotation_matrix_xyz(-self.movement_controller.orientation) @ 
+            build_rotation_matrix_xyz(-self.rotation_offset))
         offset_axis = rotmat @ x_axis
         roll = mod_bearing(self.movement_controller.orientation[0] + self.rotation_offset[0])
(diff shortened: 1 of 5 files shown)


stonesoup/sensor/base.py
```diff
@@ -115,7 +117,14 @@ def orientation(self) -> Optional[StateVector]:
             """
         if self.movement_controller is None:
             return None
-        return self.movement_controller.orientation + self.rotation_offset
+        x_axis = (1, 0, 0)
+        rotmat = build_rotation_matrix_xyz(-self.movement_controller.orientation) @ \
+            build_rotation_matrix_xyz(-self.rotation_offset)
```
etfrogers-dstl, line 122:
```suggestion
        rotmat = (build_rotation_matrix_xyz(-self.movement_controller.orientation) @ 
            build_rotation_matrix_xyz(-self.rotation_offset))
```
Marginal preference for brackets instead of `\` as a line continuation char

etfrogers-dstl (approved):

mharris-dstl (changes requested):
Have checked the maths and made a couple of suggestions.

mharris-dstl (approved):

PR #1022, opened 2024-05-22
Implementation of Loopy Belief Propagation (LBP)

marcelh71:
This pull request is the LBP algorithm of Williams and Lau (2014). This is an efficient mechanism for calculating the marginal association probabilities when performing soft data association (e.g. as in the JPDA filter).

PR #1041, opened 2024-06-12
Loopy belief propagation tutorial 

Amir-Belmekki:
This is a tutorial for Loopy belief propagation and plotter changes to visualise association probabilities.

It requires #1022


jswright-dstl:
Rewriiten as #1173 containing only the tutorial.
The plotting change (extracting association probobility matrix) is a useful addition but should be included in the Plotter(ly) classes so that it can be used with any of the plotter types and any JPDA data associator.

PR #1042, opened 2024-06-13
Fix GaussianMixtureInitiator to use a MultiMeasurementInitiator

jswright-dstl:
This PR adds the ability to use the `MultiMeasurementInitiator` with the `GaussianMixtureInitiator`. Previously, the `GaussianMixtureInitiator` would only modify the last state of a `Track` so when a track with more than one state (from the `MultiMeasurementInitiator`) is presented, the result is a `Track` with a mixed set of states (`GaussianState` and `GaussianMixture`).

### DIFF
--- a/stonesoup/initiator/simple.py
+++ b/stonesoup/initiator/simple.py
@@ -378,7 +378,7 @@ class GaussianMixtureInitiator(GaussianInitiator):
                         timestamp=state.timestamp,
                         tag=[])]
                 track[n] = GaussianMixtureUpdate(
-                    hypothesis=track.hypothesis,
+                    hypothesis=getattr(state, 'hypothesis', None),
                     components=mixture)
 
         return tracks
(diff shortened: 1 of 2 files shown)


stonesoup/initiator/simple.py
```diff
@@ -369,16 +369,17 @@ def initiate(self, detections, timestamp, **kwargs):
...
+            for n, state in enumerate(track):
+                mixture = [
+                    TaggedWeightedGaussianState(
+                        state_vector=state.state_vector,
+                        covar=state.covar,
+                        weight=Probability(1),
+                        timestamp=state.timestamp,
+                        tag=[])]
+                track[n] = GaussianMixtureUpdate(
+                    hypothesis=track.hypothesis,
```
sdhiscocks, line 381: Maybe this should get per state as well?
```suggestion
                    hypothesis=getattr(state, 'hypothesis', None),
```

sdhiscocks (approved):

stonesoup/initiator/simple.py
jswright-dstl: Good point, changed.

akenyon (approved):

Issue #1032, opened 2024-05-29
Low efficiency in GOSPA metric computations


narykov:
I'm using [the GOSPA metric](https://stonesoup.readthedocs.io/en/latest/stonesoup.metricgenerator.ospametric.html#stonesoup.metricgenerator.ospametric.GOSPAMetric) to evaluate the tracking results in a scenario that has generated 40 tracks over 30,000 time steps, and I've been waiting for results for more than a couple of hours. I understand that this isn't a trivial scenario, but it made me think about how efficient is the metric's implementation. 

The description indicates that the metric relies on the auction algorithm, which may not be the fastest option here. Angel Garcia-Fernandez [provides](https://github.com/Agarciafernandez/MTT/blob/master/GOSPA%20code/GOSPA.m) an improved (MATLAB) implementation that uses the [`assign2D`](https://github.com/schwehr/Tracker-Component-Library/blob/master/Assignment%20Algorithms/2D%20Assignment/assign2D.m) function instead. It would be great to have it in Stone Soup, provided that `assign2D` has already been [implemented](https://stonesoup.readthedocs.io/en/latest/stonesoup.dataassociator.html#stonesoup.dataassociator.neighbour.GNNWith2DAssignment) to speed up the GNN algorithm.

sdhiscocks:
I've opened #1033 which switches out the algorithm for the SciPy implementation (which is based off of linked MATLAB version).

Just as another suggestion, you could reduced/interpolated truth and tracks to reduce the number of timestamps to speed up metrics, if you don't really need metric at every timestep.

PR #1059, opened 2024-07-15
GOSPAMetric: Avoid unnecessary copies and inefficient masking

marvonlar:
`GOSPAMetric.compute_over_time()` was taking a complete copy of _all_ measurements and ground truth states for every timestep in the sequence. Additionally, it used multiple full linear masks for every iteration on the sorted timestamps. These steps cause the metric to have $` \mathcal{O}(n^2) `$ running time in the number of timesteps, resulting in slow compute on long datasets.

This PR performs the copying once and uses `itertools.groupby()` to avoid the linear masking in each iteration. As a result, `GOSPAMetric` now uses slightly less memory (we perform the same copy, but save the masking) and, in practice, runs linearly in the number of timesteps. The PR also includes a test which demonstrates 30-40x speedup on a 1k timestep long sequence. Offline I'm seeing 150x speedup (going from 20 min to 8 sec) on a real-world dataset with 50k timesteps.

This patch is also likely to help in the case of #1032.

### DIFF
--- a/stonesoup/metricgenerator/ospametric.py
+++ b/stonesoup/metricgenerator/ospametric.py
@@ -189,17 +189,17 @@ class GOSPAMetric(MetricGenerator):
         gospa_metrics = []
 
         while next_truth is not None or next_meas is not None:
-            timestamp = min(next_truth[0], next_meas[0])
+            timestamp = min(group[0] for group in [next_truth, next_meas] if group is not None)
 
             truth_idxs = []
 
-            if timestamp == next_truth[0]:
+            if next_truth is not None and timestamp == next_truth[0]:
                 truth_idxs = np.fromiter(next_truth[1], dtype=int)
                 next_truth = next(truth_iter, None)
 
             meas_idxs = []
 
-            if timestamp == next_meas[0]:
+            if next_meas is not None and timestamp == next_meas[0]:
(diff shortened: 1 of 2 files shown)


stonesoup/metricgenerator/ospametric.py
```diff
@@ -319,6 +350,8 @@ def compute_gospa_metric(self, measured_states, truth_states):
         if num_truth_states == 0:
             # When truth states are empty all measured states are false
             opt_cost = -1.0 * num_measured_states * dummy_cost
+            if self.alpha == 2:
+                gospa_metric['false'] = opt_cost
```
marvonlar, line 354: This change was not part of my first patch, but came up as I added some tests to provide coverage of my original changes.

It seems reasonable that GOSPA should report tracks as false also in the case where there are 0 ground truth states, right?

I don't fully understand why we test for `self.alpha == 2`, but this is what is done everywhere else (see for instance [line 369](https://github.com/dstl/Stone-Soup/pull/1059/files#diff-0cd5f02bb81ea6e2a48b6ad697fdc78814cfbe9a041df8458670dbca164ab153R369)).

sdhiscocks (approved):
Thanks for the contribution: a significant speed up!

jswright-dstl (approved):

PR #1060, opened 2024-07-17
Update Kernels and documentation

jswright-dstl:
This PR adds some new Kernel classes and modifies the inheritance of existing Kernels. Additional documentation is added.


sdhiscocks (approved):

stonesoup/kernel.py
```diff
@@ -16,17 +17,17 @@ class Kernel(Base):
     @abstractmethod
     def __call__(self, state1, state2=None):
```
hpritchett-dstl, line 18: now that type of state1 and state2 is now StateVectors, would it make sense to change the name of these variables as to note be confused with the State type?

sdhiscocks: Thinking about this more, as you highlight it @hpritchett-dstl ...

I do wonder if it was better being states passed, as before. This mirrors the `Measure` class, and also `Model.function()` methods. Could there be other non state vector based information as part of the state that could be used in a kernel (metadata, Gaussian covariance, ...)?

stonesoup/kernel.py
```diff
@@ -31,6 +32,11 @@ def __call__(self, state1, state2=None):
         """
         raise NotImplementedError
 
+    def update_parameters(self, kwargs):
```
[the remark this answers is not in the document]
akenyon: having thought about it more, it's not intended to be a dictionary of _only_ parameters, so calling it parameters wouldn't reflect what it's doing either

stonesoup/kernel.py
```diff
@@ -1,4 +1,5 @@
 from abc import abstractmethod
+from typing import Sequence
```
sdhiscocks, line 2:
```suggestion
from collections.abc import Sequence
```

stonesoup/kernel.py
```diff
@@ -47,113 +57,228 @@ def _get_state_vectors(state1, state2):
...
+        state1 : :class:`~.List`
+        state2 : :class:`~.List`
+
+        Returns
+        -------
+        StateVectors
+            kernel state of a pair of input :class:`~.State` objects
+        """
+        state1, state2 = self._get_states(state1, state2)
+        return self.kernel.__call__(state1, state2, **kwargs)
```
sdhiscocks, line 284: These are both very similar. Maybe a `StateKernel` or `_StateKernel` that these inherit from?

sdhiscocks (approved):

akenyon (approved):

PR #1080, opened 2024-09-12
Particle filter proposal implementation

A-acuto:
This PR updates the particle filter implementation with proposals, in particular **prior as proposal** and **(local) kalman filter**. 
This is developed by @sglvladi and myself.

Now the particle filter uses **prior as proposal** as default, otherwise a Kalman filter proposal can be used. 
Since in the prediction step we need to perform an update, we have modified as well the distance hypothesiser to include the measurements (or detections) to perform correctly the predict-update step.

We have implemented the **local** Lalman Filter proposal which propagates each particle using the kalman filter predict and update and then samples once for each particle the new state (before doing the particle update).
There is another way, commonly known as **global** Kalman Filter, where the particle distribution is approximated by a gaussian, which is propagated using the Kalman Filter and then a new set of particles is drawn from the new state. However, we encountered a number of issues in the making it working accordingly, in particular with problems with weights normalisation during the data-association step.  I will open an issue to try to get to the bottom of this. 

[Summary paper that highlights the two implementations](https://onlinelibrary.wiley.com/doi/10.1155/2014/175820)

A couple of stone soup examples are :
[single target no data association case](https://gist.github.com/A-acuto/da62c375db90f26b456c31d7e705ee5d)
[data association case](https://gist.github.com/A-acuto/471c4be3221b55bc7e78647d8da38a00)

### DIFF
--- a/stonesoup/proposal/simple.py
+++ b/stonesoup/proposal/simple.py
@@ -1,7 +1,6 @@
 from typing import Union
 
 import numpy as np
-from enum import Enum
 from scipy.stats import multivariate_normal as mvn
 
 from stonesoup.base import Property
@@ -55,7 +54,6 @@ class PriorAsProposal(Proposal):
                                      prior=prior)
 
 
-
 class KFasProposal(Proposal):
     """This proposal uses the kalman filter prediction and update steps to
     generate new set of particles and weights
@@ -84,9 +82,6 @@ class KFasProposal(Proposal):
         : :class:`~.ParticlePrediction`
         """
 
-        # get the number of particles
-        number_particles = prior.state_vector.shape[1]
-
         if measurement is not None:
             timestamp = measurement.timestamp
             time_interval = measurement.timestamp - prior.timestamp
@@ -105,7 +100,7 @@ class KFasProposal(Proposal):
         if isinstance(self.predictor, SqrtKalmanPredictor):
             prior_cls = SqrtGaussianState
 
-        # Null covariance
+        # Null covariance for the particles
         null_covar = np.zeros_like(prior.covar)
 
         predictions = [
--- a/stonesoup/proposal/simple.py
+++ b/stonesoup/proposal/simple.py
@@ -1,0 +1,146 @@
+from typing import Union
+
+import numpy as np
+from enum import Enum
+from scipy.stats import multivariate_normal as mvn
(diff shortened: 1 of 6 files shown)


stonesoup/proposal/simple.py
```diff
@@ -0,0 +1,141 @@
...
+        # Draw the samples
+        samples = np.array([state.state_vector.reshape(-1) +
+                            mvn.rvs(cov=state.covar).T
+                            for state in updates])
+
+        # Compute the log of q(x_k|x_{k-1}, y_k)
+        post_log_weights = np.array([mvn.logpdf(sample,
+                                                np.array(update.mean).reshape(-1),
+                                                update.covar)
+                                     for sample, update in zip(samples, updates)])
```
sdhiscocks, line 127: This would be better to handle custom types, as mvn converts to float arrays before taking difference.

```suggestion
        post_log_weights = np.array([mvn.logpdf(sample - update.mean.reshape(-1), cov=update.covar)
                                     for sample, update in zip(samples, updates)])
```

Also, here you use `update.mean` but above `update.state_vector`. Should these not be both the same? (In theory they should be anyway I guess? but...)

stonesoup/proposal/tests/test_simple.py
```diff
@@ -0,0 +1,132 @@
...
+from stonesoup.predictor.particle import ParticlePredictor
+from stonesoup.types.detection import Detection
+from stonesoup.models.measurement.linear import LinearGaussian
+from stonesoup.types.hypothesis import SingleHypothesis
+
+
+def test_prior_proposal():
+    # test that the proposal as prior and basic PF implementation
+    # yield same results, since they are driven by the transition model
+    np.random.seed(16549)
```
sdhiscocks, line 23: See #1064 which aims to remove any setting of global random seeds in tests.

May need to add random state option to Proposal to support this, like models have.

sdhiscocks (approved):
Looks good, but would be nice to remove global seed setting from tests.

jswright-dstl (approved):

stonesoup/proposal/simple.py
A-acuto: I see your point, `mean` and `state_vector` are equal in this case. I changed it

PR #1085, opened 2024-10-01
Make passing measurements in distance hypothesiser optional

sdhiscocks:
Change in #1080 causes the hypothesiser not able to take advantage of caching, causing docs builds to fail.

Not sure this is the best fix, but I'll propose it for now, but may need more thought.

### DIFF
--- a/stonesoup/hypothesiser/distance.py
+++ b/stonesoup/hypothesiser/distance.py
@@ -26,6 +26,10 @@ class DistanceHypothesiser(Hypothesiser):
     include_all: bool = Property(
         default=False,
         doc="If `True`, hypotheses beyond missed distance will be returned. Default `False`")
+    predict_with_measurements: bool = Property(
+        default=False,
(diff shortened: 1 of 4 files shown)


apiszcz:
FYSA
When multiprocessing I had the cache issue, disabled with
self.Mahalanobi(state_covar_inv_cache_size=0)
May not be relevant, but caching and multiprocessing had some type of issue.

sdhiscocks:
> FYSA When multiprocessing I had the cache issue, disabled with self.Mahalanobi(state_covar_inv_cache_size=0) May not be relevant, but caching and multiprocessing had some type of issue.

Thanks @apiszcz. That was to do with pickling for multiprocessing, which should have been fixed in #750.

This issue is being caused by measurement being passed to every predict call by default, even if the predictor doesn't use it, so can't cache predictions across multiple measurements.

jswright-dstl (approved):

nperree-dstl (approved):

PR #1090, opened 2024-10-11
Platform orientation calculated from previous velocity

nperree-dstl:
Slight change to how orientation is calculated for moving platforms. Where the velocity of the platform is small, the orientation is calculated based on the direction it travelled from its previous position. Where the platform has only moved a small distance, and for a non-moving platform the orientation from the previous time step is used.

### DIFF
--- a/stonesoup/movable/movable.py
+++ b/stonesoup/movable/movable.py
@@ -295,8 +295,8 @@ class MovingMovable(Movable):
                 _, bearing = cart2pol(*velocity.flat)
                 self._property_orientation = StateVector([0, 0, bearing])
             else:
-                raise NotImplementedError('Orientation of a moving platform is only implemented for 2'
-                                          'and 3 dimensions')
+                raise NotImplementedError('Orientation of a moving platform is only' 
+                                          'implemented for 2 and 3 dimensions')
 
         return self._property_orientation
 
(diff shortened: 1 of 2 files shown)


nperree-dstl:
Struggling to run tests/docs on my machine so I think this should fix the bug/feature we discussed but I'm not 100% confident...

faultfactory:
Something that's not clear to me is what happens on first timestep and there is no motion? 
Do we default to ```StateVector([0.0,0.0,0.0])``` or do we instantiate the movable with an orientation to set any attached sensors in the right direction?

akenyon (changes requested):
I had the same thought as @faultfactory. For a platform with 0 velocity and only one state, `self._property_orientation` would not be assigned a value, which I think would result in an attribute error. We could add a default value or maybe just an error message to explain why orientation couldn't be calculated?

faultfactory:
> I had the same thought as @faultfactory. For a platform with 0 velocity and only one state, `self._property_orientation` would not be assigned a value, which I think would result in an attribute error. We could add a default value or maybe just an error message to explain why orientation couldn't be calculated?

I think you do both but ideally there wouldn't be an error.  Have a default of zero and then spit something out to the terminal on first use of that value. Then let the processing chain chew on that repeatedly until there's motion.

akenyon (approved):

sdhiscocks (approved):

PR #1092, opened 2024-10-17
Various package updates

sdhiscocks:
- Move to `pyproject.toml` and drop Python 3.8
- Fix git archival file format
- Fix issue with certain types being used with `Property`
- Allow some tests to be skipped (useful with Python 3.13 early testing waiting for other packages to update)
- Fix number of warnings being raised during tests
- Parallel docs building for CircleCI. Reduces build time by about ~30%
- Add Python 3.13 test (custom dependencies currently as not all packages support Python 3.13 for some optional features i.e. MFA).
- Remove dependencies from README, as they weren't up to date and documented better in `pyproject.toml`
- Update for PEP 585 supported from Python >=3.9, and needed by 2025 (assume Python 3.14).
- Add `plotly` to main dependencies.


pyproject.toml
```diff
@@ -3,3 +3,117 @@ requires = ["setuptools>=64", "setuptools_scm>=8"]
...
+dev = [
+    "confluent-kafka",
+    "coverage>=7.2",
+    "flake8",
+    "folium",
+    "h5py",
+    "pandas",
+    "ortools",
+    "pillow",
+    "plotly>=5.0",
```
jswright-dstl, line 55: Given we have most documentation using (Animated)Plotterly should plotly dependency move from dev optional dependencies to the required dependencies?

jswright-dstl (approved):

jswright-dstl (approved):

stonesoup/tests/test_plotter.py
```diff
@@ -106,6 +102,14 @@
 )
 
 
+@pytest.fixture(autouse=True)
```
gawebb-dstl, line 105: This isn't closing all the matplotlib plots. I've tested by putting in an additional test file which displays all of the unclosed plots and blocks (with ``plt.show()``). 
```
import matplotlib.pyplot as plt
import pytest

def test_nothing():
    plt.show()
```

gawebb-dstl (changes requested):
See comment about matplotlib plots being left open. Otherwise all good

stonesoup/tests/test_plotter.py
sdhiscocks: Thanks. Some figure from metric tests weren't closed properly. Fixed in c7baec5fb9ecf87fc9dce22d0b8e0713c94ec03d

gawebb-dstl (approved):

PR #1111, opened 2025-01-30
Fix StateMutableSequence to allow for subclasses with mandatory properties

yoshum:
Thank you for nice software.

In this PR, I fixed StateMutableSequence so that its subclasses can have mandatory properties. Indeed, if a subclass has a mandatory property, the subclass cannot be instantiated, or its mandatory property will be unexpectedly put in a list, depending on how arguments are passed to the constructor (see the changes in `stonesoup/types/tests/test_state.py`). A simple fix can resolve this issue.

There is one side effect that affects the error raised by `MovingMovable` when it is instantiated without specifying the mandatory argument "states". This change has made a test for MovingMovable fail, although I think for most users, the change itself is harmless. Therefore, I fix the test to make it pass.

### DIFF
--- a/stonesoup/movable/tests/test_movable.py
+++ b/stonesoup/movable/tests/test_movable.py
@@ -32,7 +32,7 @@ def test_empty_state_error():
                       velocity_mapping=[1, 2, 5],
                       transition_model=None,
                       )
-    with pytest.raises(ValueError, match='States must not be empty'):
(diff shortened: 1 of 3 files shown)


sdhiscocks (approved):
Thanks for the contribution @yoshum 👍

csherman-dstl (approved):

PR #1113, opened 2025-02-04
Make track uncertainty alpha an optional keyword argument

iwishiwasaneagle:
I often find that the track uncertainties overlap too much and result in a solid block of color obscuring the rest of the plot. This allows the user to decrease the alpha to prevent this.

### DIFF
--- a/stonesoup/plotter.py
+++ b/stonesoup/plotter.py
@@ -369,7 +369,7 @@ class Plotter(_Plotter):
            may be removed in the future.
         """
         label = kwargs.pop('track_label', None) or label
-        tracks_kwargs = dict(linestyle='-', marker="s", color=None)
+        tracks_kwargs = dict(linestyle='-', marker="s", color=None, alpha=0.2)
         tracks_kwargs.update(kwargs)
         if not isinstance(tracks, Collection) or isinstance(tracks, StateMutableSequence):
             tracks = {tracks}  # Make a set of length 1
@@ -437,13 +437,13 @@ class Plotter(_Plotter):
                         ellipse = Ellipse(xy=state.mean[mapping[:2], 0],
                                           width=2 * np.sqrt(w[max_ind]),
                                           height=2 * np.sqrt(w[min_ind]),
-                                          angle=np.rad2deg(orient), alpha=0.2,
+                                          angle=np.rad2deg(orient), alpha=track_kwargs['alpha'],
                                           color=track_colors[track])
                         self.ax.add_artist(ellipse)
                         artists.append(ellipse)
 
                 # Generate legend items for uncertainty ellipses
-                ellipse_handle = Ellipse((0.5, 0.5), 0.5, 0.5, alpha=0.2,
+                ellipse_handle = Ellipse((0.5, 0.5), 0.5, 0.5, alpha=track_kwargs['alpha'],
                                          color=tracks_kwargs['color'])
                 ellipse_label = "Uncertainty"
                 self.legend_dict[ellipse_label] = ellipse_handle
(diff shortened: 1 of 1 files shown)


jswright-dstl (approved):

stonesoup/plotter.py
```diff
@@ -437,13 +437,13 @@ def plot_tracks(self, tracks, mapping, uncertainty=False, particle=False, label=
                         ellipse = Ellipse(xy=state.mean[mapping[:2], 0],
                                           width=2 * np.sqrt(w[max_ind]),
                                           height=2 * np.sqrt(w[min_ind]),
-                                          angle=np.rad2deg(orient), alpha=0.2,
+                                          angle=np.rad2deg(orient), alpha=track_kwargs['alpha'],
```
csherman-dstl, line 440:
```suggestion
                                          angle=np.rad2deg(orient), alpha=tracks_kwargs['alpha'],
```

stonesoup/plotter.py
```diff
@@ -437,13 +437,13 @@ def plot_tracks(self, tracks, mapping, uncertainty=False, particle=False, label=
...
                                           color=track_colors[track])
                         self.ax.add_artist(ellipse)
                         artists.append(ellipse)
 
                 # Generate legend items for uncertainty ellipses
-                ellipse_handle = Ellipse((0.5, 0.5), 0.5, 0.5, alpha=0.2,
+                ellipse_handle = Ellipse((0.5, 0.5), 0.5, 0.5, alpha=track_kwargs['alpha'],
```
csherman-dstl, line 446:
```suggestion
                ellipse_handle = Ellipse((0.5, 0.5), 0.5, 0.5, alpha=tracks_kwargs['alpha'],
```

csherman-dstl (changes requested):
A slight typo that is causing the tests to fail

csherman-dstl (approved):

stonesoup/plotter.py
```diff
@@ -369,7 +369,7 @@ def plot_tracks(self, tracks, mapping, uncertainty=False, particle=False, label=
            may be removed in the future.
         """
         label = kwargs.pop('track_label', None) or label
-        tracks_kwargs = dict(linestyle='-', marker="s", color=None)
+        tracks_kwargs = dict(linestyle='-', marker="s", color=None, alpha=0.2)
```
sdhiscocks, line 372: This also has the consequence of making the track itself transparent and hard to see. Maybe a separate argument could be added for ellipse alpha value?

sdhiscocks: image

iwishiwasaneagle: I see... this is due to
https://github.com/iwishiwasaneagle/Stone-Soup/blob/4b1fb02dea83a1b8fc98add971cafad3a1f73513/stonesoup/plotter.py#L395-L398
Would changing it to `uncertainty_kwargs` be more apt?

sdhiscocks: With `uncertainty_kwargs` as at least that gives most flexibility for anything additional anyone might want. But that has complications with dealing with colour to keep sure that matches the track (unless specified). Whereas `uncertainty_alpha` is simpler and there aren't that many extra arguments other than alpha that seem that useful looking at matplotlib docs.

I think either approach would be fine.

iwishiwasaneagle: `uncertainty_alpha` is now a function kwarg. Running the bearings only example shows that the above bug has been fixed. I was further able to easily set the `uncertainty_alpha` by editing the example.
image

sdhiscocks:
Thanks for the contribution @iwishiwasaneagle

PR #1120, opened 2025-02-12
Sample Position Action Generator and Movable

timothy-glover:
This PR introduces additional `movable` components for actionable platforms. New sample based category of `movable` and action generator have been introduced: `_SampleActionableMovable` and `SamplePositionActionGenerator`, along with an implementation based on uniform sampling within a circle around the current platform position.

`CircleSamplePositionActionGenerator` implements the generator logic behind the contribution, yielding `MovePositionAction`s based on sampling from a circle around the current position. Radius of the circle is defined by the user with the property: `maximum_travel`. The user can also specify the number of samples to generate with `n_samples`. This generator is implemented with `CircleSampleActionableMovable` and inherits its base methods from `_SampleActionableMovable`.

Tests have been written for the new components and they have been incorporated into the documentation.

### DIFF
--- a/stonesoup/movable/action/move_position_action.py
+++ b/stonesoup/movable/action/move_position_action.py
@@ -187,22 +187,29 @@ class CircleSamplePositionActionGenerator(SamplePositionActionGenerator):
                                  end_time=self.end_time,
                                  target_value=self.current_value)
 
-        generated_samples = 0
+        radius_angle_samples = np.random.uniform([0, 0], [1, 2*np.pi], (self.n_samples, 2))
 
-        while generated_samples < self.n_samples:
+        values = self.maximum_travel*np.sqrt(radius_angle_samples[:, 0]) *\
+            np.array([np.sin(radius_angle_samples[:, 1]), np.cos(radius_angle_samples[:, 1])])
 
-            radius_angle_sample = np.random.uniform([0, 0], [1, 2*np.pi], 2)
+        target_values = self.current_value + values
 
-            value = self.maximum_travel*np.sqrt(radius_angle_sample[0]) *\
-                np.array([[np.sin(radius_angle_sample[1])], [np.cos(radius_angle_sample[1])]])
+        if self.action_space is not None:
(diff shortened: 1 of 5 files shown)


stonesoup/movable/action/move_position_action.py
```diff
@@ -104,3 +104,105 @@ def __iter__(self):
...
+
+    def __contains__(self, item):
+        return item in iter(self)
+
+    @abstractmethod
+    def __iter__(self):
+        raise NotImplementedError
+
+
+class CircleSamplePositionActionGenerator(SamplePositionActionGenerator):
```
hpritchett-dstl, line 163: suggest to import these classes in this file in the __init__.py to allow for neater imports elsewhere.

stonesoup/movable/action/move_position_action.py
```diff
@@ -104,3 +104,105 @@ def __iter__(self):
...
+
+        yield MovePositionAction(generator=self,
+                                 end_time=self.end_time,
+                                 target_value=self.current_value)
+
+        generated_samples = 0
+
+        while generated_samples < self.n_samples:
+
+            radius_angle_sample = np.random.uniform([0, 0], [1, 2*np.pi], 2)
```
hpritchett-dstl, line 194: This could be sampled outside of the while loop, something like: `np.random.uniform([0, 0], [1, 2*np.pi], (self.n_samples, 2)))`. Should be more efficient, especially when self.n_samples is large.

stonesoup/movable/action/move_position_action.py
```diff
@@ -104,3 +104,105 @@ def __iter__(self):
...
+
+            value = self.maximum_travel*np.sqrt(radius_angle_sample[0]) *\
+                np.array([[np.sin(radius_angle_sample[1])], [np.cos(radius_angle_sample[1])]])
```
hpritchett-dstl, line 197: similarly to the above, could make use of numpys vectorised math outside of the while loop

hpritchett-dstl (approved):

akenyon (approved):

PR #1130, opened 2025-02-27
Information Coverage Reward

timothy-glover:
This pull request introduces the `InformationCoverageReward` class which implements an information coverage control objective for the sensor management module. The reward works with an information state which spatially quantifies the level of information at discrete points throughout the environment. The class rewards minimisation of the difference between the information state and a user defined reference information state multiplied into, if triggered, a target density or a uniform density, depending on whether a target is tracked or whether it is being searched for. Information here is a quantification of the area covered by sensors, and is left to the user to define for generality across sensor types and applications. 

The class requires that the information state is stored in track metadata in this implementation as the state should be carried through time as it evolves according to the information model.

The reward has also only been tested with a single target scenario but the concept is applicable to multitarget tracking problems and the implementation should allow this.

The new reward class has been incorporated into existing sensor manager tests. There is a problem with it being used with the `OptimizeBasinHoppingSensorManager` so assertions for this sensor manager and `InformationCoverageReward` are skipped until the problem is resolved.

### DIFF
--- a/stonesoup/sensormanager/reward.py
+++ b/stonesoup/sensormanager/reward.py
@@ -2,7 +2,9 @@ from abc import ABC
 import copy
 import datetime
 from collections.abc import Mapping, Sequence
+from typing import Set, Callable
 
+from scipy.spatial import cKDTree
 import numpy as np
 
 from ..base import Base, Property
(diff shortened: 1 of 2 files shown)


timothy-glover:
Docs appear to be failing on an unrelated example

nperree-dstl (changes requested):
I think this is a really interesting reward function to add! It does take a bit of effort to get your head around though as it's quite complicated. I've left some comments on where the documentation could be improved but longer term an example showcasing it's use would be great. 

The only thing is I couldn't work out why the docs are failing, not sure if it's related to what you've done or not...

PR #1131, opened 2025-02-27
Rename PriorAsProposal to DynamicsProposal, and KFasProposal to KalmanProposal

sglvladi:
This PR renames the proposals introduced in #1080 to make them more intuitive and representative of their purpose. Specifically:
- `PriorAsProposal` has been renamed to `DynamicsProposal`
- `KFasProposal` has been renamed to `KalmanProposal`.

This is also an attempt to make the names of the classes consistent with the names that will be used in the paper we aim to publish to Fusion 2025.

**Note**: This PR introduces breaking changes for code that references the proposals directly. Any legacy code, or more generally code that does not explicitly specify a proposal, should run without errors.

### DIFF
--- a/stonesoup/predictor/particle.py
+++ b/stonesoup/predictor/particle.py
@@ -12,7 +12,7 @@ from .kalman import KalmanPredictor, ExtendedKalmanPredictor
 from ..base import Property
 from ..models.transition import TransitionModel
 from ..proposal.base import Proposal
-from ..proposal.simple import PriorAsProposal
+from ..proposal.simple import DynamicsProposal
 from ..sampler.particle import ParticleSampler
 from ..types.numeric import Probability
 from ..types.prediction import Prediction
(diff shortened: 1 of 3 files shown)


sglvladi:
Would it be worth to add backwards compatibility with a deprecation warning message for uses of `PriorAsProposal` and `KFasProposal` as per [PEP562](https://peps.python.org/pep-0562/)? For example by adding the code below to `proposal/simple.py`

```python
import warnings

__all__ = ["DynamicsProposal", "KalmanProposal"]

DEPRECATED_NAMES = [("PriorAsProposal", "DynamicsProposal"), ("KFasProposal", "KalmanProposal")]

def __getattr__(name):
    for old_name, new_name in DEPRECATED_NAMES:
        if name == old_name:
            warnings.warn(f"The '{old_name}' class or function has been renamed to '{new_name}' and will be removed in a future release.",
                          DeprecationWarning,
                          stacklevel=2)
            return globals()[new_name]
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")


def __dir__():
    return sorted(__all__ + [names[0] for names in DEPRECATED_NAMES])

```

Doing so would allow use of the deprecated class names, but will raise a `DeprecationWarning` message, e.g.:

```python

>>> from stonesoup.proposal.simple import PriorAsProposal

<input>:1: DeprecationWarning: The 'PriorAsProposal' class or function has been renamed to 'DynamicsProposal' and will be removed in a future release.

>>> print(a := PriorAsProposal(None))
DynamicsProposal(
    transition_model=None)
```

jswright-dstl (approved):

sdhiscocks (approved):

PR #1147, opened 2025-03-26
Add multi data feeders with queues

sdhiscocks:
These multi data feeders are designed to allow reading from real-time data sources, where there may be latencies or data coming out of sequence. This includes the standard FIFO, LIFO and Priorty Queue classes from Python standard library, as well as a custom prioity queue which drops old data, rather than blocking or raising error when full.

### DIFF
--- a/stonesoup/feeder/multi.py
+++ b/stonesoup/feeder/multi.py
@@ -70,7 +70,7 @@ class FIFOMultiDataFeeder(_QueueMultiDataFeeder):
     """FIFO Multi-data Feeder
 
     This returns data from multiple data readers as a single stream,
-    where each reader is consumed in separate thread and put into a
+    where each reader is consumed in a separate thread and put into a
     queue. The data is consumed first in, first out.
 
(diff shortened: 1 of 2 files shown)


stonesoup/feeder/multi.py
```diff
@@ -19,3 +23,133 @@ class MultiDataFeeder(DetectionFeeder, GroundTruthFeeder):
...
+                continue
+            yield time, data
+            self._queue.task_done()
+
+
+class FIFOMultiDataFeeder(_QueueMultiDataFeeder):
+    """FIFO Multi-data Feeder
+
+    This returns data from multiple data readers as a single stream,
+    where each reader is consumed in separate thread and put into a
```
timothy-glover, line 73:
```suggestion
    where each reader is consumed in a separate thread and put into a
```
Not sure if this should be `a separate thread` or `separate threads`. The same applies to all class docstrings.

csherman-dstl (approved):

jswright-dstl (approved):

PR #1148, opened 2025-03-26
Add SinglePointMeasurementInitiator

sdhiscocks:
This is like SinglePointInitiator, except the state vector of the measurement is used to replace the prior.

Also update the SingePointInitiator to allow use of updaters other than EKF.

### DIFF
--- a/stonesoup/initiator/simple.py
+++ b/stonesoup/initiator/simple.py
@@ -1,3 +1,5 @@
+import copy
+
 import numpy as np
 from scipy.stats import multivariate_normal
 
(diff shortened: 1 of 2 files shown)


stonesoup/initiator/simple.py
```diff
@@ -49,20 +59,87 @@ def initiate(self, detections, timestamp, **kwargs):
...
+                    except NotImplementedError:
+                        if not self.skip_non_reversible:
+                            raise
+                        else:
+                            continue
+                elif self.skip_non_reversible:
+                    continue
+                else:
+                    raise Exception("Invalid measurement model used.\
+                                    Must be instance of linear or reversible.")
```
hpritchett-dstl, line 126: in this current implementation, a user wouldn't be able to make use of a reversible measurement model they would pass through to the initiator if the detections had a non-reversible measurement model attached to them, e.g. `CartesianToElevationBearingRange `and `CartesianToElevationBearingRangeRate`. I suppose in that instance a mapping would also be needed to access the correct components in the detection for the reversible measurement model to map back to state space, so things will probably get overly complicated and messy for a "simple" initiator. Happy with the current implementation.

hpritchett-dstl (approved):

jswright-dstl (approved):

PR #1155, opened 2025-04-03
Computation performance with multi-sensor update wrappers

sdhiscocks:
This PR includes two updater wrappers that may result in computation performance increases when dealing with large number of measurements at the same time from multiple sensors. This may be slower when only using small amount of data, and also can result in non-optimal results in case of non-linear measurement models.

### DIFF
--- a/docs/source/stonesoup.updater.rst
+++ b/docs/source/stonesoup.updater.rst
@@ -95,3 +95,9 @@ Probabilistic
 
 .. automodule:: stonesoup.updater.probability
     :show-inheritance:
+
+Multi
+-----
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
Both multi-sensor wrappers only warn when the hypotheses contain different predictions, then `predictions.pop()` chooses an arbitrary one and fuses every measurement against it. That violates the documented same-prediction precondition and can make the result depend on set iteration rather than the hypotheses. Reject differing predictions (or handle them explicitly) and add a mismatch regression.

Issue #1145, opened 2025-03-21
Make Clutter From Measurement Plotter More Configurable


mkierzenka:
It's neat how `Plotter.plot_measurements()` automatically handles plotting `ClutterDetection`s as a different color/shape vs. `TrueDetection`s (e.g. MTT_3D_Platform example below).

![Image](https://github.com/user-attachments/assets/03b7b695-80dc-49c5-9f30-e63076658e8a)

But when visualizing simulations that have a very high clutter-to-true detection ratio, a user may sometimes want to plot only the true detections without the clutter. Is there an easy way to do this? Perhaps it would be helpful to extend the `Plotter.plot_measurements()` function with an optional argument "show_clutter"?

PR #1157, opened 2025-04-05
Make Clutter From Measurement Plotter More Configurable

mkierzenka:
Addresses #1145 

This PR updates the `plot_measurements(...)` method for all plotters/plotterlys. The method has a new optional parameter `show_clutter` (default `True`). The use case is to simplify plotting measurements in scenarios with a high clutter to truth ratio, where the user may want to sometimes make plots without the clutter points (which would obscure the rest of the data, for example).
I also added some unit tests for the new functionality, keeping all existing tests as they were.

Please let me know what you think!


sdhiscocks:
Duplicate of #1156

Issue #1146, opened 2025-03-21
Update Pandas DataFrame Reader Example To Use Built-In Reader


mkierzenka:
The [Use of Custom Readers that support Pandas DataFrames](https://stonesoup.readthedocs.io/en/latest/auto_examples/readers/Custom_Pandas_Dataloader.html) example is helpful to learn about how to implement a custom data reader in Stone Soup, but I think it may be confusing to new users. When I first read it, I thought it implied that in order to add support for reading truth/measurement data from DataFrames I needed to write my own `_DataFrameReader`, `DataFrameGroundTruthReader`, and `DataFrameDetectionReader` classes as shown in the first section of that example. However, these classes are now natively included in Stone Soup: `stonesoup/reader/pandas_reader.py`.

Perhaps this example should be updated to remove those sections showing reimplemented code and instead utilize the built-in classes? What do you think?

apiszcz:
FYSA: https://github.com/dstl/Stone-Soup/issues/971

mkierzenka:
> FYSA: [#971](https://github.com/dstl/Stone-Soup/issues/971)

Thanks for the heads up. Is your plan to add a new example or expand the existing one?

apiszcz:
No new plans, dataframe integration is useful.

mkierzenka:
I found #354 and #707 about this example code serving as a demonstration for how users can add custom Reader classes and figured it is worth keeping for that purpose. Instead I opted to add a note of clarification and ensure the DataFrame reader classes appear in the docs.

PR #1158, opened 2025-04-08
Updated the Pandas DataFrame Example and a few other small tweaks

mkierzenka:
Addresses #1146 

The main purpose of this PR is to clarify the Pandas DataFrame support in Stone Soup:
1. Mention the example is only for demonstration purposes (the support for reading pandas DFs is already included in Stone Soup). I was originally going to remove the `Reader` implementations from that example. But then I found #354 and #707 about it serving as a demonstration for how users can add custom Reader classes and figured it is worth keeping, along with a clarifying note.
2. Make the Pandas DataFrame Reader classes appear in the Stone Soup docs.

Along the way I found a few other tweaks to examples and documentation that seem like improvements (happy to discuss).
I also updated a few URLs in the sphinx document generation because the command output suggested to do so.

### DIFF
--- a/docs/examples/filters/Multi_Tracker_Example.py
+++ b/docs/examples/filters/Multi_Tracker_Example.py
@@ -241,7 +241,6 @@ init_transition_model = CombinedLinearGaussianTransitionModel(
     (ConstantVelocity(1), ConstantVelocity(1)))
 init_predictor_EKF = ExtendedKalmanPredictor(init_transition_model)
 init_predictor_UKF = UnscentedKalmanPredictor(init_transition_model)
-init_predictor_PF = ParticlePredictor(init_transition_model)
 
 # %%
 # The final step before running the trackers is to create the initiators:
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -88,9 +88,9 @@ sphinx_gallery_conf = {
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/{.major}'.format(sys.version_info), None),
-    'matplotlib': ('https://matplotlib.org/', None),
-    'numpy': ('https://docs.scipy.org/doc/numpy', None),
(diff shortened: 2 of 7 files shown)


docs/examples/filters/Multi_Tracker_Example.py
```diff
@@ -241,7 +241,6 @@
     (ConstantVelocity(1), ConstantVelocity(1)))
 init_predictor_EKF = ExtendedKalmanPredictor(init_transition_model)
 init_predictor_UKF = UnscentedKalmanPredictor(init_transition_model)
-init_predictor_PF = ParticlePredictor(init_transition_model)
```
mkierzenka, line 244: `init_predictor_PF` is unused here and in the Metrics example

docs/source/conf.py
mkierzenka, line 1: I noticed the following lines in the output when building the Sphinx docs:
```
intersphinx inventory has moved: https://matplotlib.org/objects.inv -> https://matplotlib.org/stable/objects.inv
intersphinx inventory has moved: https://docs.scipy.org/doc/scipy/reference/objects.inv -> https://docs.scipy.org/doc/scipy/objects.inv
intersphinx inventory has moved: https://docs.scipy.org/doc/numpy/objects.inv -> https://numpy.org/doc/stable/objects.inv
```

I manually checked those new URLs and they seem fine, so I updated the config file here. Now the build looks more regular:
```
loading intersphinx inventory 'matplotlib' from https://matplotlib.org/stable/objects.inv...
loading intersphinx inventory 'numpy' from https://numpy.org/doc/stable/objects.inv...
loading intersphinx inventory 'scipy' from https://docs.scipy.org/doc/scipy/objects.inv...
```

I confirmed in the docs I built locally that example code referencing each of those libraries still links properly.

jswright-dstl (approved):

sdhiscocks (approved):

PR #1100, opened 2024-11-14
Add Bayesian search examples

akenyon:
Add two examples showcasing the use of Bayesian search in Stone Soup. Note that much of length of the two files comes from plotting functions which are hidden by default in the rendered examples.

PR #1159, opened 2025-04-08
Add (updated) Bayesian search examples

akenyon:
This is an update to #1100, created as a fresh PR for cleanliness. All of the comments from the previous PR have been addressed, and some additional changes were made following reviews outside of GitHub.


jmbarr:
The final graph has random search outperforming sequential search in contrast to the accompanying text. Is the graph label wrong or do we need to let it run a bit longer?

docs/examples/sensormanagement/bayesian_search_example_2.py
```diff
@@ -0,0 +1,465 @@
...
+# this case, the initial distribution will be Gaussian, centred around (x=7, y=8).
+
+
+from stonesoup.types.state import ParticleState
+from stonesoup.types.array import StateVectors
+from scipy.stats import multivariate_normal
+
+# parameterise our prior probability distribution
+mu = np.array([7, 8])  # mean(x, y)
+sigma = np.array([[10, 0], [3, 6]])  # covariance matrix [[var(x), covar(xy)], [covar(xy), var(y)]]
```
jmbarr, line 134: covar(xy) should equal covar(yx). Doesn't the following line throw an error?

jmbarr: No because according to SciPy.stats the symmetry is not checked and only the lower triangle is used. No excuse not to correct it though.

docs/examples/sensormanagement/bayesian_search_example_2.py
```diff
@@ -0,0 +1,465 @@
...
+
+    return float(running_tot)
+
+
+# %%
+# Creating a Sensor Manager
+# ~~~~~~~~~~~~~~~~~~~~~~~~~
+# Finally, we create a :class:`~.BruteForceSensorManager`, which will evaluate every possible
+# movement of our platform using the custom reward function and return the best one.
+
```
jmbarr, line 197: 'Best' is subjective -- it will return the action with the highest numerical reward value.

docs/examples/sensormanagement/bayesian_search_example_2.py
```diff
@@ -0,0 +1,465 @@
...
+#     <br>  
+#     <video autoplay loop controls width=100% height="auto">
+#       <source src="../../_static/bayesian-search-ex2-plt1.mp4" type="video/mp4">
+#     </video>
+#     <br>
+#
+# | 
+#
+# As expected, at each timestep the platform and sensor move to the neighbouring group of cells
+# that has the highest total probability of containing the target.
```
jmbarr, line 455: Doesn't appear to do that for the first step though. 

I does, however, start from the wrong place in the video. Isn't the initial position (4,4)?

jmbarr (approved):
The minor stylist stuff not as important as ensuring the final graphs/videos in each example make sense.

docs/examples/sensormanagement/bayesian_search_example_2.py
```diff
@@ -0,0 +1,465 @@
...
+# platform. We will see how Bayesian search can be used to move the platform to the areas most
+# likely to contain the target.
+#
+# As in the first example, we will be using the class :class:`~.ParticleState` to build our search
+# space, but this time we will be creating a 2D search grid.
+
+# %%
+# Setup
+# -----
+# We begin with some generic imports, and then start the clock...
```
timothy-glover, line 28: Starting the clock gives the impression that the runtime of the notebook is going to influence the performance, when in reality it is just precomputing the timesteps.

docs/examples/sensormanagement/bayesian_search_example_2.py
```diff
@@ -0,0 +1,465 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~
+# We create the moving platform to which our sensor will be mounted. Rather than defining the
+# movement pattern of the platform in advance, we want the platform to automatically choose an
+# optimal movement at each timestep. We can use a platform with an
+# :class:`~.NStepDirectionalGridMovable` to achieve this.
+#
+# Platforms with this movement controller can generate a list of actions at each timestep which
+# can then be given to a :class:`~.SensorManager`. The sensor manager chooses the best action(s)
+# for the current timestep. We define the platform's initial position, and limit its movement to a
+# step of three grid cells in a single direction.
```
timothy-glover, line 50: I understand what you mean by single direction here but the platform can actually move in 4 directions. Might need some rewording.

timothy-glover (approved):
I haven't checked locally but the thumbnails don't seem to be rendering in the examples section on the online rendered docs.

timothy-glover:
> The final graph has random search outperforming sequential search in contrast to the accompanying text. Is the graph label wrong or do we need to let it run a bit longer?

Adding to this, I can't see any random number generator seed. This would be worth adding as it will ensure the graphs are the same after the example is regenerated with future updates. This may also be the cause of the incorrect accompanying text.

docs/examples/sensormanagement/bayesian_search_example_2.py
akenyon: Hopefully the current wording makes this clearer.

PR #1161, opened 2025-04-10
Add PathBasedPlatform

jswright-dstl:
This PR adds a `PathBasedPlatform` class. This class has the abiltiy to define a platform by a `GroundTruthPath` rather than a an individual `TransitionModel` or multiple.

### DIFF
--- a/stonesoup/platform/base.py
+++ b/stonesoup/platform/base.py
@@ -1,7 +1,10 @@
 import uuid
 from collections.abc import MutableSequence
+from datetime import datetime
+
 
 from ..base import Property, Base
+from ..functions.interpolate import interpolate_state_mutable_sequence
 from ..movable import Movable, FixedMovable, MovingMovable, MultiTransitionMovable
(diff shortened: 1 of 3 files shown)


stonesoup/platform/base.py
```diff
@@ -189,3 +192,17 @@ class MovingPlatform(Platform):
 
 class MultiTransitionMovingPlatform(Platform):
     _default_movable_class = MultiTransitionMovable
+
+
+class PathBasedPlatform(MovingPlatform):
+    path: GroundTruthPath = Property(doc="Ground truth path used for the platform. "
+                                     "This is for use in cases when platform paths are imported "
+                                     "from external data, e.g., from a CSV file")
+
```
gawebb-dstl, line 201: Is it worth giving the `transition_model` property a default value of `None` as it is no longer required required?

jswright-dstl: Added.

PR #1162, opened 2025-04-28
Add sub_tracks and root_tracks properties to track type

csherman-dstl:
The `sub_tracks` property contains a set of tracks that are important to the formation of the track, e.g. in the case of track fusion this would be the set of tracks that were fused together to form this track.
The `root_tracks` property is the total set of `sub_tracks` including the `sub_tracks` of tracks that were used to form the top level track.

### DIFF
--- a/stonesoup/types/track.py
+++ b/stonesoup/types/track.py
@@ -1,6 +1,7 @@
 import copy
 import uuid
 from collections.abc import MutableSequence, MutableMapping
+from typing import Collection
 
 from .multihypothesis import MultipleHypothesis
 from .state import State, StateMutableSequence
@@ -31,6 +32,10 @@ class Track(StateMutableSequence):
         default={}, doc="Initial dictionary of metadata items for track. Default `None` which "
                         "initialises track metadata as an empty dictionary.")
 
+    sub_tracks: set = Property(
+        default=None,
+        doc="The initial sub tracks of the track. Default ``None`` will initialise an empty set.")
+
     def __init__(self, *args, **kwargs):
 
         super().__init__(*args, **kwargs)
@@ -41,6 +46,13 @@ class Track(StateMutableSequence):
             self._update_metadata_from_state(state)
         if self.id is None:
             self.id = str(uuid.uuid4())
+
+        if self.sub_tracks is None:
+            self.sub_tracks = set()
+        elif isinstance(self.sub_tracks, Track):
+            self.sub_tracks = set([self.sub_tracks])
+        elif isinstance(self.sub_tracks, Collection):
+            self.sub_tracks = set(self.sub_tracks)
 
     def __setitem__(self, index, value):
         super().__setitem__(index, value)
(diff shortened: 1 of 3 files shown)


jmbarr:
Nomenclature: problem with `sub_tracks` is that as a name it suggest a subset of the `states` in the  `Track` (i.e. shorter bits, or part of a track). A better name might be `source_tracks`.

jmbarr:
Are `root_tracks` defined as those source tracks which themselves don't have any source tracks?

csherman-dstl:
> Are `root_tracks` defined as those source tracks which themselves don't have any source tracks?

Yes. In the case of there being source tracks that have source tracks of their own, `root_tracks` would be the collection of the lowest level source tracks

jmbarr:
Can you make the case for this inclusion via an example? I can see a case where you'd want to avoid fusing tracks which contain the same 'source tracks' and/or ensure that each 'root track' is only fused once (or at least fused with appropriate weighting). 

> > Are `root_tracks` defined as those source tracks which themselves don't have any source tracks?
> 
> Yes. In the case of there being source tracks that have source tracks of their own, `root_tracks` would be the collection of the lowest level source tracks

Is there a case for making `source_tracks` a tree? That would allow you to identify `root_tracks` (the leaves) but also gives you access to tools to augment the tree (i.e. when you do simple track fusion) or use the properties of the tree (i.e. when you want to do fancier track fusion to resolve shared inheritance/information reuse), or efficient methods to store/process the tree.

sglvladi:
Just a thought. Would it not make more sense to have this functionality under a new class, say `FusedTrack`, that is a subclass of `Track`?

jmbarr:
> Just a thought. Would it not make more sense to have this functionality under a new class, say `FusedTrack`, that is a subclass of `Track`?

We did think about that but I don't like it for two reasons. Firstly, your new thing isn't practically different from the old thing, save for a bit of metadata which might only be interrogated occasionally. We should apply a parsimonious approach to creation of new classes. Secondly, a track's provenance might be relatively complex. Suppose it was generated by track fusion and subsequently updated by plot fusion; initialised by plots, updated by track; was a combination of tracks but then split at some point in the past; etc. We'd need to head down the road of giving different names to all these types.

stonesoup/types/track.py
```diff
@@ -140,3 +155,13 @@ def _update_metadata_from_state(self, state):
                 hypothesis = state.hypothesis
                 if hypothesis and hypothesis.measurement.metadata is not None:
                     self.metadata.update(hypothesis.measurement.metadata)
+
+    @property
+    def root_tracks(self) -> set:
```
jmbarr, line 160: Consider not doing this. `source_tracks` vs `root_tracks` is unnecessary overspecification. If it isn't give an example of how you are using them and why the properties need to be distinct.

My view remains that a `source_tracks = tree()` is better with `root_tracks` if you need them identified as leaf nodes.

PR #1176, opened 2025-05-27
Use native Python datetime objects in GOSPA metric

sdhiscocks:
This allows support of datetimes with timezone, which currently raises a warning in NumPy but in future will error.


sdhiscocks:
See timezone deprecation warning in [NumPy Datetimes docs](https://numpy.org/doc/stable/reference/arrays.datetime.html#datetimes-and-timedeltas)

jswright-dstl (approved):

Issue #1181, opened 2025-06-08
Tests fail for RoadNetwork type for NetworkX 3.5+


sglvladi:
It looks like NetworkX have changed the return type of calls to ``shortest_path`` with no ``source`` and ``target`` from ``dict`` to a generator in version 3.5 (see networkx/networkx#7754). 

This breaks current tests for ``RoadNetwork.shortest_path()``.

PR #1187, opened 2025-06-13
Add MaxSpeedAction for platforms

csherman-dstl:
Add platform action that generates actions within a maximum range of a platform, based on the platforms maximum speed and the duration of the action.
Similar to `SamplePositionActionGenerator` but will produce actions according to a resolution and angle_resolution similar to the `AngleActionsGenerator`s instead of a random sample.
Includes `action_from_value` method to create actions from `StateVector`s.
`__contains__` method only looks for a position to be within the possible radius of the platform

### DIFF
--- a/stonesoup/movable/action/tests/test_move_position_action.py
+++ b/stonesoup/movable/action/tests/test_move_position_action.py
@@ -361,7 +361,7 @@ def test_max_speed_action_gen(gen_param_dict, state, position_mapping):
                 generator = platform.actions(start_timestamp).pop()
             return
 
-    generator = platform.actions(start_timestamp).pop()
+    generator = platform.actions(end_timestamp).pop()
 
     # Check that parameters have been set correctly
     assert generator.action_mapping == action_mapping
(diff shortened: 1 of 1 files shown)


stonesoup/movable/max_speed.py
nperree-dstl, line 1: I think you need to add this to the rst file so if appears on the docs

nperree-dstl (approved):
This works well and is a nice addition. It would be nice to replace the platform/moveable in the [Actionable Platform example](https://stonesoup.readthedocs.io/en/latest/auto_examples/sensormanagement/actionable_platform_example.html) with this one as it is more generic behaviour. 

I'm not sure what's causing the tests to fail though - I can't work out how it would be related to something you've added...

csherman-dstl:
I think the test fails come from #1181 which is now fixed, will rebase those changes in when I add the max_speed file to docs

stonesoup/movable/max_speed.py
```diff
@@ -0,0 +1,89 @@
...
+
+from stonesoup.base import Property
+from stonesoup.movable import FixedMovable
+from stonesoup.types.state import State
+
+from .action.move_position_action import MaxSpeedPositionActionGenerator
+
+
+class MaxSpeedActionableMovable(FixedMovable):
+    """Class for movables can move in any direction. To be used with
```
akenyon, line 12:
```suggestion
    """Class for movables that can move in any direction. To be used with
```

stonesoup/movable/max_speed.py
```diff
@@ -0,0 +1,89 @@
...
+        ----------
+        timestamp: datetime.datetime
+            Time of action finish.
+        start_timestamp: datetime.datetime, optional
+            Time of action start.
+
+        Returns
+        -------
+        : set of :class:`~.MoveToActionGenerator`
+            Set of grid action generators, that describe the bounds of each action space.
```
akenyon, line 60: Should the grid part be changed?

akenyon (approved):

PR #1203, opened 2025-08-01
Optimise Plotterly ellipse plotting

hpritchett-dstl:
Utilize Scipy's [ellipeinc](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.ellipeinc.html)  to calculate the points on an ellipse to plot.

Below are two plots using `Plotterly` to plot the track on the OpenSky demo with uncertainty = True.

Existing `_generate_ellipse_points method`
genereate_ellipse_points_old

Proposed `_generate_ellipse_points method`
generatre_ellipse_points_new


csherman-dstl (approved):

hpritchett-dstl:
note that this calculation is slightly different. I believe the existing implementation should have m as the eccentricity of the ellipse ,rather than its square, i.e `m = np.sqrt(1 - (b**2 / a**2))`

Scipy's `ellipeinc` parameterises the elliptic integral of the second kind differently, and so it makes sense to use the square of the eccentricity in this proposed method.

sdhiscocks (approved):

PR #1216, opened 2025-09-10
Add multi-hypothesiser based MFA with multi-model example

sdhiscocks:


hpritchett-dstl:
Only thing i can think of is to vectorize the multivariate normal logpdf calculation in each `PDAHypothesiser. hypothsise` call. We could potentially call each hypothesiser's predictor's predict and do one single vectorized `batch_multivariate_normal_logpdf` in `MFAHypothesiser.hypothesise` , and have these cached. I can see how this added bloat won't merit the speed increase though.

PR #1221, opened 2025-09-30
Introduce an example for tracking a target under known control

timothy-glover:
This PR introduces the first in a new category of examples: control models. This example shows the function of a control model and how a user of Stone Soup may create one and how it can be used to create a target trajectory. Assuming the control inputs and model are know to the tracking system, the example then demonstrates the use of these components to perform better target tracking. This has limited applications when tracking a target as we don't normally know what the target is doing but this is the first step into a planned series of examples of control models. There are no new components for this example.

### DIFF
--- a/docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
+++ b/docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
@@ -86,7 +86,7 @@ np.random.seed(1991)
 # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 #
 # To define the control model class, we will create a variant of :class:`~.LinearControlModel` and
-# ovewrite the :attr:`~.LinearControlModel.matrix` attribute according to the above expression.
+# overwrite the :attr:`~.LinearControlModel.matrix` attribute according to the above expression.
 
 
 from stonesoup.models.control.linear import LinearControlModel
--- a/docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
+++ b/docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
@@ -22,6 +22,13 @@ Control Models: Inference with Linear Gaussian Control Model
 #   5. Define and run baseline Kalman filter
 #   6. Define and run Kalman filter with control
 #   7. Compare performance
+#
+# Before proceeding with the example, the reader is reminded that this is illustrative
+# of how it is possible to create a control model, rather than the preferred way. There are
+# a number of possible techniques for implementing such models depending on how a problem
+# or system is defined. There are a number of choices made in this example which do not
+# define the only way to create the demonstrated behaviours. The reader may wish to consider
+# alternative approaches when implementing for their own system or problem at hand.
 
 # %%
 # Standard external module imports and Environment Setup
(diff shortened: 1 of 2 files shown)


docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
```diff
@@ -0,0 +1,330 @@
...
+# Initialise truth object
+truth = GroundTruthPath([GroundTruthState([0, 1, 0, 1], timestamp=start_time)])
+
+# Record these for later use when filtering.
+controlinputs = []
+for k in range(1, 181):
+
+    velocity = truth[-1].state_vector[[1, 3]]
+
+    if k > 40 and k < 90:
```
gawebb-dstl, line 190:
```suggestion
    if 40 < k < 90:
```

docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
```diff
@@ -0,0 +1,330 @@
...
+        # Rotate it to the right
+        newvelocity = right@velocity
+    elif k > 120 and k < 170:
```
gawebb-dstl, line 193:
```suggestion
    elif 120 < k < 170:
```

docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+# Transition model
+from stonesoup.models.transition.linear import CombinedLinearGaussianTransitionModel, \
+                                               ConstantVelocity
+transition_model = CombinedLinearGaussianTransitionModel([ConstantVelocity(0.0005),
+                                                          ConstantVelocity(0.0005)])
+
+# Control model
+control_model = ConstantAccelerationLinearControlModel(np.array([[1., 0],
```
jmbarr: Do we want to mention that the input control matrix here is a dummy just used to force the correct dimension? The custom class above overwrites the initial values with the ones we want.

jmbarr (approved):
One optional comment and I appear to have broken one of the test (which has nothing to do with my changes).

docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
```diff
@@ -0,0 +1,330 @@
...
+for k in range(1, 181):
+
+    velocity = truth[-1].state_vector[[1, 3]]
+
+    if k > 40 and k < 90:
+        # Rotate it to the right
+        newvelocity = right@velocity
+    elif k > 120 and k < 170:
+        # Rotate it to the left
+        newvelocity = left@velocity
```
sdhiscocks:
```suggestion
        newvelocity = left @ velocity
```

docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
```diff
@@ -0,0 +1,330 @@
...
+
+# Record these for later use when filtering.
+controlinputs = []
+for k in range(1, 181):
+
+    velocity = truth[-1].state_vector[[1, 3]]
+
+    if k > 40 and k < 90:
+        # Rotate it to the right
+        newvelocity = right@velocity
```
sdhiscocks:
```suggestion
        newvelocity = right @ velocity
```

sdhiscocks (approved):

PR #1235, opened 2025-11-20
Deprecate Orbital Functions

sgboakes:
Following the release of the [Astrodynamics Plugin](https://ieeexplore.ieee.org/abstract/document/11124098), and the lack of visibility of use for these functions in Stone Soup, this PR begins the deprecation of these functions for v1.9, to be removed in v1.10. This should give any remaining users enough time to seek out the astrodynamics plugin or other alternatives.


stonesoup/functions/orbital.py
```diff
@@ -441,6 +460,8 @@ def keplerian_to_rv(state_vector, grav_parameter=3.986004418e14):
     -------
     No checking undertaken. Assumes Keplerian elements rendered correctly as above
 
+    .. deprecated:: 1.9
```
csherman-dstl: is this in the astro plugin

jswright-dstl (approved):

nperree-dstl (approved):

Issue #1258, opened 2026-02-25
Deprecation warning when using `Angle` type


csherman-dstl:
#1235 added deprecation warnings for the orbital functions in Stone Soup.
In stonesoup.types.angle exists the classes `Inclination` and `EclipticLongitude` which use the deprecated orbital `mod_inclination` and `mod_elongitude` functions.
This dependency results in a deprecation warning being raised when trying to import the classes `Angle`, `Elevation` and `Bearing`.

I assume the simplest solution to this would be to move the import functions for `mod_inclination` and `mod_elongitude` within the `Inclination` and `EclipticLongitude` classes.

PR #1259, opened 2026-02-26
Fix Incorrect DeprecationWarning from Orbital Classes

sgboakes:
As raised in #1258, DeprecationWarnings were being raised when importing Angle types that were not being deprecated. This PR moves orbital-specific angle classes into the orbitalstate.py file, to remove the warning for other Angle classes.

### DIFF
--- a/stonesoup/types/angle.py
+++ b/stonesoup/types/angle.py
@@ -250,6 +250,7 @@ class Latitude(Elevation):
     Multiplication or division produces a float object rather than Latitude.
     """
 
+
 class Inclination(Angle):
     """(Orbital) Inclination angle class.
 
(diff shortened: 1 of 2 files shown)


stonesoup/types/angle.py
```diff
@@ -250,27 +249,3 @@ class Latitude(Elevation):
     The return type for addition and subtraction is Latitude.
     Multiplication or division produces a float object rather than Latitude.
     """
-
-
-class Inclination(Angle):
```
sdhiscocks: Concern with removing these, is this will be breaking change, although likely not commonly used.

How about moving the import into the `mod_angle` method? Not ideal as it will raise when used rather than when imported, but I think may be better to avoid breaking change.

```
    @staticmethod
    def mod_angle(value):
        from ..functions.orbital import mod_inclination
        return mod_inclination(value)
```

PR #1265, opened 2026-03-12
Change tutorial error values to be in the correct units

### LINKED ISSUE #1263: Small Mistake in Tutorial 2 (Units of Variance are Squared)
In [Tutorial 2](https://stonesoup.readthedocs.io/en/v1.8/auto_tutorials/02_ExtendedKalmanFilterTutorial.html) when setting up the CartesianToBearingRange model, the covariance is initialized by

`noise_covar=np.diag([np.radians(0.2), 1]),  # Covariance matrix. 0.2 degree variance in bearing and 1 metre in range `

The mistake here is that standard deviation has the same units as the mean / base variable, but variance (diagonal of the covariance matrix) should be in units squared of the base unit.
Therefore, the conversion from degrees to radians also has to be applied twice to convert from degree^2 to radians^2.
The correct initialization should therefore be 

 `noise_covar=np.diag([np.radians(np.radians(0.2)), 1]),  # Covariance matrix. 0.2 degree^2 variance in bearing and 1 metre^2 in range `

csherman-dstl:
Resolves #1263

### DIFF
--- a/docs/tutorials/02_ExtendedKalmanFilterTutorial.py
+++ b/docs/tutorials/02_ExtendedKalmanFilterTutorial.py
@@ -129,7 +129,7 @@ sensor_y = 0
 measurement_model = CartesianToBearingRange(
     ndim_state=4,
     mapping=(0, 2),
-    noise_covar=np.diag([np.radians(0.2) ** 2, 1]),  # Covariance matrix. 0.2 degree error in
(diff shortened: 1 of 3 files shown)


sdhiscocks (approved):

hpritchett-dstl:
should we inflate the standard deviation so that it matches the original values? I feel having increased measurement noise makes for a more effective tutorial for someone coming to the content for the first time.

hpritchett-dstl (approved):

PR #1270, opened 2026-03-23
Fix multiple sensor_locations bug in PCRBMetric, add support for multiple measurement models

lwlarsson:
Solves #1269


sylvesterkaczmarek (commented):
`PCRBMetric` renames the required public property from `measurement_model` to `measurement_models`, but this PR does not update `stonesoup/metricgenerator/tests/test_pcrbmetric.py`, which still constructs `PCRBMetric(measurement_model=...)` and passes a single model to these helpers. The current test suite therefore breaks before exercising the new behavior. Preserve backward-compatible singular input or update the API/tests together, with both single- and multi-model coverage.

PR #1271, opened 2026-03-25
Simple markovian GP

nfcc2:
Summary:
This PR introduces the Simple Markovian Gaussian Process (GP)-based linear Gaussian transition model into the Stone Soup library. This serves as the foundational model for more complex Dynamics-Informed models. These models combine the benefits of data-driven and mechanics-informed tracking, offering greater flexibility for tracking applications where standard kinematic models may be insufficient.

This was submitted as a draft PR #1140 to facilitate submitting the paper "Dynamics-Informed Gaussian Process models in Stone Soup" for Fusion 2025. 

This PR only introduces the Simple Markovian GP. Future key additions will include:

`Simple Markovian GP`: A basic zero-mean GP models position using a windowed subset of past states.
Integrated GP (iGP): As proposed in [1], where an iGP models target position, and the target velocity is a stochastic process purely driven by a GP disturbance.
`Dynamics-Informed Integrated GP` (iDGP): A novel GP formulation where the target velocity evolves under a first-order differential equation with a GP disturbance.
`Twice-Integrated GP` (iiGP) and Dynamics-Informed Twice-Integrated GP (iiDGP): Integrated versions of the iGP and iDGP respectively. A GP models target position while the driving GP acts as a disturbance on acceleration.
Currently, these models are implemented for the case where the driving GP has the squared exponential (SE) covariance function, as the analytical expressions for their covariance functions have been derived in [1] and the new Fusion paper. Alternative kernels can be implemented by overriding class methods. These models integrate directly into the existing Kalman filtering framework.

Changes Introduced:

Added class: `MarkovianGP`
Added wrapper class `GPPredictorWrapper` to support GP-based state propagation. The GP models have attribute `requires_track` set to `True`. The wrapper passes track history to the GP model to extract timestamps for computing the covariance matrices. See discussion in PR https://github.com/dstl/Stone-Soup/pull/564on GPs. Alternative approaches may be preferred?
See this [GitHub repo](https://github.com/nfcc2/dynamics-informed-gps.git) for example usage.

References
[1] F. Lydeard, B. I. Ahmad, and S. Godsill, “Integrated Gaussian Processes for Tracking,” IEEE Open Journal of Signal Processing, pp. 1–9, 2025.


sylvesterkaczmarek (commented):
`_gp_pred` is cached only by the time vector, but its result also depends on mutable `kernel_params` and `jitter`. Reusing the same timestamps after changing or tuning either returns stale GP weights/variance, and in-place edits to the parameter dict cannot invalidate the cache at all. Include the hyperparameters in the cache key or rework cache invalidation, with a same-time-vector parameter-change regression.

PR #1278, opened 2026-05-29
FOV Interaction Sensor Management

jswright-dstl:
This PR adds the `FOVInteractionRewardFunction` which when combined with the composite reward function `UncertaintyRewardFunction` using the `MultiplicativeRewardFunction` is able to demonstarate the reward function keeping the target inside the sensor platform's field of view, but penalises if the sensor platform enters the target's field of view (which we assume is known or estimated).

### DIFF
--- a/docs/examples/sensormanagement/fov_interaction.py
+++ b/docs/examples/sensormanagement/fov_interaction.py
@@ -1,0 +1,245 @@
+#!/usr/bin/env python
+
+"""
+===========================================
+FOV based Reward Function Sensor Management
+===========================================
+"""
+
+# %%
+# This notebook is designed to showcase the use of Stone Soup for simulating "interesting"
+# behaviour between a sensor platform and a target.
+# The target is travelling through the scenario with a known FOV, the sensor platform has to
+# maintain sensing of the target (i.e., keep the target in the sensor platform's FOV but not
+# enter the target's FOV otherwise it would be detected).
+#
+# Background and notation
+# -----------------------
+#
+# *[Some more info on something useful here]*.
+
+# %%
+# Setting Up the Scenario
+# -----------------------
+# We generate a ground truth of a target following a constant velocity with an amount of
+# noise that makes the truth interesting.
+
+import numpy as np
+from datetime import datetime, timedelta
+
+from stonesoup.models.transition.linear import (CombinedLinearGaussianTransitionModel,
+                                                ConstantVelocity)
+from stonesoup.types.groundtruth import GroundTruthPath, GroundTruthState
+
+np.random.seed(1990)
+
+start_time = datetime.now().replace(second=0, microsecond=0)
+
+transition_model = CombinedLinearGaussianTransitionModel(
+    [ConstantVelocity(10), ConstantVelocity(10)])
+
+truth = GroundTruthPath([GroundTruthState([-450, 5, 450, -5], timestamp=start_time)])
+duration = 120
+timesteps = [start_time]
+
+for k in range(1, duration):
+    timesteps.append(start_time+timedelta(seconds=k))
+    truth.append(GroundTruthState(
+        transition_model.function(truth[k-1], noise=True, time_interval=timedelta(seconds=1)),
+        timestamp=timesteps[k]))
+
+# %%
+# Visualising the ground truth
+# ----------------------------
+
+from stonesoup.plotter import AnimatedPlotterly
+plotter = AnimatedPlotterly(timesteps, tail_length=0.3)
+plotter.plot_ground_truths(truth, [0, 2])
+plotter.fig
+
+# %%
+# See, interesting.
+
+# %%
+# Creating the Actionable Platform
+# --------------------------------
+#
+# *Next we create the actionable platform itself*. We use a :class:`~.MovingPlatform` (because it
+# would be silly to use a :class:`~.FixedPlatform` for a moving platform) with a
+# :class:`~.MaxSpeedActionableMovable` movement controller.
+# The platform starts alongside the target (as this is not a search problem (yet)).
+# We define the platform's movement such that it is capable of moving up to a maximum speed in a
+# number of directions (8 cardinal directions (cardinal and ordinal)).
+#
+# We add a :class:`~.RadarRotatingBearingRange` radar to this platform, which has a field of
+# view of 360 degrees, a range of 200, and is capable of rotating its dwell centre by
+# 90 degrees each timestep. Similarly a :class:`~.RadarRotatingBearingRange` radar is added to the
+# target platform (primarily for use in plotting the target FOV) with a field of view of 360
+# degrees and a range of 100.
+
+from stonesoup.platform import MovingPlatform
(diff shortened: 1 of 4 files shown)


stonesoup/sensormanager/reward.py
```diff
@@ -443,3 +445,72 @@ def _generate_detections(self, predicted_tracks, sensors, timestamp=None):
...
+    This function rewards the sensor for keeping the target in its FOV while
+    penalising it for entering the target's FOV.
+    """
+    predictor: KalmanPredictor = Property(doc="")
+    updater: ExtendedKalmanUpdater = Property(doc="")
+    sensor_fov_radius: float = Property(default=20.0, doc="")
+    target_fov_radius: float = Property(default=10.0, doc="")
+    sensor_mapping: list[int] = Property(default=(0, 1), doc="")
+    target_mapping: list[int] = Property(default=(0, 2), doc="")
+    fov_scale: float = Property(default=1.0, doc="")
```
timothy-glover: I think the doc strings are missing from these properties

stonesoup/sensormanager/reward.py
```diff
@@ -443,3 +445,72 @@ def _generate_detections(self, predicted_tracks, sensors, timestamp=None):
...
+            measure = Euclidean(self.sensor_mapping, self.target_mapping)
+            distance = measure(sensor_pos, target_pos)
+            # Reward for keeping the target in the sensor's FOV
+            tracking_reward = (2 * no_tracks
+                               if distance <= self.sensor_fov_radius * self.fov_scale
+                               else -no_tracks)
+            # Penalty for entering the target's FOV
+            lack_of_stealth_penalty = (-no_tracks if distance <= self.target_fov_radius else 0.0)
+            # Combine the reward and penalty
+            total_reward += tracking_reward + lack_of_stealth_penalty
```
timothy-glover, line 531: Some comments could be added to indicate that these loops are not currently used in this single target/sensor reward function for clarity.

Euclidean could be moved outside the loop but you may want it to stay in if there are are plans for having multiple targets and potentially multiple target mappings.

stonesoup/plotter.py
```diff
@@ -3606,3 +3606,47 @@ def plot_state_sequence(self, state_sequences, angle_mapping: int, range_mapping
 
                 frame.data = data_
                 frame.traces = traces_
+
+
+def plot_sensor_fov(fig_, sensor_set, sensor_history, color='blue', alpha=0.2, label="Sensor FOV"):
```
timothy-glover: It looks like `sensor_history` is the usual dictionary that we create in SM examples but it would be good to add some documentation for it here

stonesoup/sensormanager/tests/test_rewardfunction.py
```diff
@@ -98,3 +135,54 @@ def test_unequal_additive():
...
+
+
+def test_target_in_sensor_fov_but_not_in_target_fov(fov_reward):
+    sensor = DummySensor(State([0, 0, 0, 0]))
+    config = {sensor: [DummyAction()]}
+    tracks = {Track(State([15, 0, 0, 0]))}
+    metric_time = datetime.now()
+    reward = fov_reward(config, tracks, metric_time)
+    # distance ~14.14 < 20, so tracking_reward=2, lack_of_stealth_penalty=0
+    assert reward == 2.0
```
timothy-glover: Not required but these tests could be combined with a `pytest.mark.parametrize`

stonesoup/sensormanager/reward.py
```diff
@@ -443,3 +445,72 @@ def _generate_detections(self, predicted_tracks, sensors, timestamp=None):
...
+            target_pos = target
+
+            measure = Euclidean(self.sensor_mapping, self.target_mapping)
+            distance = measure(sensor_pos, target_pos)
+            # Reward for keeping the target in the sensor's FOV
+            tracking_reward = (2 * no_tracks
+                               if distance <= self.sensor_fov_radius * self.fov_scale
+                               else -no_tracks)
+            # Penalty for entering the target's FOV
+            lack_of_stealth_penalty = (-no_tracks if distance <= self.target_fov_radius else 0.0)
```
csherman-dstl: Should weighting of having target in FOV vs being in target FOV be configurable?

stonesoup/plotter.py
```diff
@@ -3606,3 +3606,47 @@ def plot_state_sequence(self, state_sequences, angle_mapping: int, range_mapping
...
+        fig_.add_trace(go.Scatter(
+            mode='markers', marker=dict(color=color, size=6),
+            showlegend=False, hoverinfo='skip'))
+
+    for frame in fig_.frames:
+        traces_ = list(frame.traces)
+        data_ = list(frame.data)
+
+        timestring = frame.name
+        timestamp = datetime.strptime(timestring, "%Y-%m-%d %H:%M:%S")
```
csherman-dstl: This line fails if there are microseconds in the timestamp, could either use
```suggestion
        timestamp = datetime.fromisoformat(timestring)
```
or
```suggestion
        timestamp = datetime.strptime(timestring, "%Y-%m-%d %H:%M:%S.%f")
```

stonesoup/sensormanager/tests/test_rewardfunction.py
```diff
@@ -98,3 +135,54 @@ def test_unequal_additive():
...
+    return FOVInteractionRewardFunction(
+        predictor=DummyPredictor(),
+        updater=DummyUpdater(),
+        sensor_fov_radius=20.0,
+        target_fov_radius=10.0,
+        sensor_mapping=[0, 2],
+        target_mapping=[0, 2])
+
+
+def test_target_in_sensor_fov_and_not_in_target_fov(fov_reward):
```
csherman-dstl: Is this test named right?
The target is in the sensors fov and the sensor is in the targets fov

stonesoup/plotter.py
csherman-dstl: Probably not the second one because it will break if there is a timezone

docs/examples/sensormanagement/fov_interaction.py
```diff
@@ -0,0 +1,242 @@
...
+
+for timestep in timesteps[1:]:
+    target_platform.move(timestep)
+    target_sensor_history[timestep][target_sensor] = copy.deepcopy(target_sensor)
+target_sensor_set = {target_sensor}
+sensor_set = {sensor}
+
+plot_sensor_fov(plotter.fig, sensor_set, sensor_history)
+plot_sensor_fov(plotter.fig, target_sensor_set, target_sensor_history, label="Target FOV",
+                color="red")
```
csherman-dstl:
```suggestion
                colour="red")
```
Fallen victim to the Americanisation of spelling.

Either this or you can Americanise the argument in the function

jswright-dstl: I made the change in the function but forgot to update the usage.

csherman-dstl (approved):

sdhiscocks (approved):

PR #1279, opened 2026-05-29
Add Sensor Management Areas of interest example

jswright-dstl:
This PR adds the `AOIAccess2DRewardFunction` and `AreaOfInterest` shape, which are able to demonstarate the reward function changing as the target changes areas of interest.


stonesoup/types/shape.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
```
sdhiscocks, line 1: Wonder if this would be better in `types` module?

stonesoup/sensormanager/reward.py
```diff
@@ -443,3 +446,122 @@ def _generate_detections(self, predicted_tracks, sensors, timestamp=None):
...
+class AOIAccess2DRewardFunction(RewardFunction):
+    interest_thresholds: Mapping[int, RewardFunction] = Property(default=None,
+                                                                 doc="Mapping of interest "
+                                                                 "thresholds to reward functions")
+    access_thresholds: Mapping[int, RewardFunction] = Property(default=None,
+                                                               doc="Mapping of access "
+                                                               "thresholds to reward functions")
+    default_reward: RewardFunction = Property(doc="Default reward function")
+    areas: Sequence[AreaOfInterest] = Property(doc="List of areas")
+    target_mapping: Sequence = Property(doc="Position mapping for the target")
```
sdhiscocks:
```suggestion
    target_mapping: tuple[int, int] = Property(doc="Position mapping for the target")
```

stonesoup/sensormanager/reward.py
```diff
@@ -443,3 +446,122 @@ def _generate_detections(self, predicted_tracks, sensors, timestamp=None):
...
+    This function rewards the sensor for keeping the target in its FOV while
+    penalising it for entering the target's FOV.
+    """
+    predictor: KalmanPredictor = Property(doc="")
+    updater: ExtendedKalmanUpdater = Property(doc="")
+    sensor_fov_radius: float = Property(default=20.0, doc="")
+    target_fov_radius: float = Property(default=10.0, doc="")
+    sensor_mapping: list[int] = Property(default=(0, 1), doc="")
+    target_mapping: list[int] = Property(default=(0, 2), doc="")
+    fov_scale: float = Property(default=1.0, doc="")
```
sdhiscocks, line 463: Missing doc strings

sdhiscocks (changes requested):

docs/examples/sensormanagement/area_of_interest_sensor_management.py
```diff
@@ -0,0 +1,363 @@
...
+# --------------------------
+# Different areas of interest can be defined by setting minimum/maximum values for x
+# and y using the :class:`~.AreaOfInterest`. Here we have 3 areas of different
+# levels of interest.
+
+from stonesoup.sensormanager.shape import AreaOfInterest
+
+area1 = AreaOfInterest(xmax=0, interest=4)
+area2 = AreaOfInterest(xmin=0, interest=7)
+area3 = AreaOfInterest(xmin=1500, interest=10)
```
timothy-glover, line 136: It might be worth adding some comments on how the areas of interest are being defined (intervals in x direction and infinite in the y direction. Areas 2 and 3 are also overlapping with the both extending to infinity. There aren't any problems with defining it this way as the higher priority area takes precedence in the reward function but its not very transparent on first inspection.

stonesoup/types/shape.py
```diff
@@ -0,0 +1,24 @@
...
+            raise ValueError("Area must have xmin < xmax and ymin < ymax")
+
+
+class AreaOfInterest(Area):
+    interest: int = Property(default=1,
+                             doc="Interest level of the area."
+                                 "0: Low interest, 10: high interest.")
+    access: int = Property(default=1,
+                           doc="Access level of the area."
+                               "0: full access, 10: high threat/no go area.")
```
timothy-glover, line 30: Missing docstrings for these classes

stonesoup/sensormanager/reward.py
```diff
@@ -443,3 +446,122 @@ def _generate_detections(self, predicted_tracks, sensors, timestamp=None):
...
+            # Penalty for entering the target's FOV
+            lack_of_stealth_penalty = (-no_tracks if distance <= self.target_fov_radius else 0.0)
+            # print("lack_of_stealth_penalty: ", lack_of_stealth_penalty)
+            # Combine the reward and penalty
+            total_reward += tracking_reward + lack_of_stealth_penalty
+
+        return total_reward
+
+
+class AOIAccess2DRewardFunction(RewardFunction):
```
timothy-glover: Missing docstring for the reward function

stonesoup/sensormanager/reward.py
```diff
@@ -443,3 +446,122 @@ def _generate_detections(self, predicted_tracks, sensors, timestamp=None):
...
+                                interest_reward = reward
+
+                    access_reward = None
+                    if self.access_thresholds is not None:
+                        for threshold, reward in self.access_thresholds.items():
+                            if threshold <= area.access:
+                                access_reward = reward
+
+                    if interest_reward and access_reward:
+                        reward_func = AdditiveRewardFunction([interest_reward, access_reward])
```
timothy-glover, line 574: I like the consideration of having both areas of interest and areas with restricted access. I had to check the AOI class to find out what the purpose of the `access_reward` is though. This is how I have interpreted the functionality here: The area of interest is an area where we may want sensors to act with extra freedom to improve tracking. The access areas are areas where we don't want sensors to go or be penalised for going into. Given these contrasting objectives, does the additive reward function need to take the negative of the `access_reward`?

Might be good to add the access part to the example so its clear.

stonesoup/plotter.py
```diff
@@ -3606,3 +3606,45 @@ def plot_state_sequence(self, state_sequences, angle_mapping: int, range_mapping
 
                 frame.data = data_
                 frame.traces = traces_
+
+
+def plot_sensor_fov(fig_, sensor_set, sensor_history, color='blue', alpha=0.2, label="Sensor FOV"):
```
timothy-glover, line 3612: This function seems like it should be a method on the plotterly class rather than a standalone that only works with plotterly. Looks like it is also missing tests.

timothy-glover (changes requested):

stonesoup/sensormanager/reward.py
jswright-dstl: Added in #1278

stonesoup/sensormanager/reward.py
nperree-dstl: Added information in docstring of class which should resolve this. Also slight renaming of class to put less emphasis on Access as that was unneccessary.

stonesoup/plotter.py
nperree-dstl: Resolved in #1278

sdhiscocks (approved):

timothy-glover (approved):

PR #1281, opened 2026-05-29
Add TransitionBasedLinearControlModel to control model example

jswright-dstl:


docs/examples/controlmodels/control_models_inference_with_linear_gaussian_control_model.py
```diff
@@ -148,34 +170,11 @@ def matrix(self, time_interval, **kwargs) -> np.ndarray:
...
-control_model = ConstantAccelerationLinearControlModel(np.array([[1., 0],
-                                                                 [1., 0],
-                                                                 [0, 1.],
-                                                                 [0, 1.]]),
-                                                       control_noise=np.diag([0.005,
-                                                                              0.005]))
+# target will traverse according to nearly constant velocity when not manoeuvring.
 
 # %%
 # We can then calculate the control inputs and subsequent ground truth states.
```
timothy-glover:
```suggestion
```

timothy-glover: I don't this line is needed with your changes

stonesoup/models/control/linear.py
```diff
@@ -69,3 +71,35 @@ def function(self, control_input, prior=None, noise=False, **kwargs) -> StateVec
                 noise = 0
 
         return self.matrix(**kwargs) @ (control_vector + noise)
+
+
+class TransitionBasedLinearControlModel(LinearControlModel):
+    r"""A model that applies a Nth Derivative Transition Model over a specified time period.
```
timothy-glover:
```suggestion
    r"""A model that applies an Nth Derivative Transition Model over a specified time period.
```

stonesoup/models/control/linear.py
```diff
@@ -69,3 +71,35 @@ def function(self, control_input, prior=None, noise=False, **kwargs) -> StateVec
...
+    This is just a :class:`~.LinearControlModel` which accepts a time_interval input
+    to matrix to compute the control matrix.
```
timothy-glover:
```suggestion
   to compute the control matrix.
```

timothy-glover (commented):
All valuable changes for the control models. One thing I thought of when looking at the matrix calculation is that there is no way to have the control action included in the output of the control model function (if acceleration was also included in the `state_vector` for example. This is true of the constant acceleration class created in the previous version of the control model example and doesn't need to be incorporated here but maybe we could include a flag property in the control model class which indicates whether the control control action is also part of the state and should be returned from the function.

timothy-glover (approved):

timothy-glover (approved):

sdhiscocks (approved):

PR #1290, opened 2026-06-23
Fix support for NumPy 2.5

sdhiscocks:
`linalg.eig{,vals}` now only returns complex numbers, even if result is real. Change to use `eig{,vals}h` which returns real numbers

### DIFF
--- a/docs/examples/dataassociation/MFA_example.py
+++ b/docs/examples/dataassociation/MFA_example.py
@@ -32,7 +32,7 @@ slide_window = 3  # Slide window; used by MFA data associator
 # present within the slide window.
 
 def plot_covar(state, ax, color=None):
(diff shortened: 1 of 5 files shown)


jmbarr:
Any idea why Kalman updater tests fail on circleci? (They pass locally.)

jmbarr:
So.. revert on eigvals?

sdhiscocks:
> Any idea why Kalman updater tests fail on circleci? (They pass locally.)

Not sure as fine local here as well, so reverted for now. They seem happy with complex numbers so no need to change anyway.

Changes are now for plotting and find nearest positive definite matrix function.

jswright-dstl (approved):

jmbarr (approved):

PR #1291, opened 2026-06-23
DetectorContext: Add support for dynamic prob_detection and clutter_spatial_density

martinvl:
This pull request implements the suggested `DetectorContext` abstraction proposed in [Context-Aware Sensor Modeling for Asynchronous Multi-Sensor Tracking in Stone Soup](https://arxiv.org/abs/2603.15137) (accepted to FUSION 2026).

The tl;dr is that we want to support cases where probability of detection is a function of the track/predicted state, and clutter intensity is a function of the detection/measurement state. This comes up in partially covered, asynchronous, or otherwise non-uniform sensing scenarios where single global `prob_detection` and `clutter_spatial_density` values are too restrictive.

The existing Stone Soup data-source flow is kept as:

```python
time, detections
```

The change is that `detections` may now be a `DetectionSet`, which is a small extension of the usual `set` of detections with an optional `detector_context` attribute. Plain `set` inputs are still supported.

The attached `DetectorContext` provides two query points:

- `prob_detection(hypothesis)`, for evaluating detection probability from the predicted track state and current detector context.
- `clutter_spatial_density(detection)`, for evaluating clutter intensity at a particular detection/measurement.

Trackers that can exploit this read `getattr(detections, "detector_context", None)` and pass it through to the relevant tracking components using the `detector_context` keyword. Components that use detection probability or clutter intensity can then query the context instead of treating those values as fixed scalars.

This initial PR wires the interface into the GM point-process path, including the existing `PointProcessMultiTargetTracker` and point-process updater classes. Existing scalar `prob_detection` and `clutter_spatial_density` configuration is preserved by falling back to a constant `SimpleDetectorContext` when no explicit context is supplied.

Readers, trackers, and the relevant tracking components need to be updated to make use of `DetectorContext`. The compatibility goal is that mixed old/new pipelines still run: plain `set` detections continue to work, `DetectionSet` remains set-like for existing trackers, and components that are not given an explicit context continue to use their existing scalar configuration. Such mixed pipelines may not exploit detector context end-to-end, but they should preserve existing behaviour.

The PR also includes a small filter tutorial showing the scalar path, a `DetectionSet` carrying detector context, and the same update using context-dependent detection probability and clutter intensity.

Some obvious follow-up work is deliberately left out to keep the first PR small: preserving detector context through feeders that rebuild detection sets, adding PDA/JPDA support, and extending the same pattern to SMC-PHD/particle tracker paths.

### DIFF
--- a/docs/source/stonesoup.types.rst
+++ b/docs/source/stonesoup.types.rst
@@ -34,6 +34,12 @@ Detection Types
 .. automodule:: stonesoup.types.detection
     :show-inheritance:
 
+Detector Context Types
+----------------------
+
+.. automodule:: stonesoup.types.detector_context
+    :show-inheritance:
+
 Graph Types
 -----------
 .. automodule:: stonesoup.types.graph
@@ -126,5 +132,4 @@ Update Types
 
 .. automodule:: stonesoup.types.update
     :show-inheritance:
-
 
--- a/docs/tutorials/filters/DetectorContextTutorial.py
+++ b/docs/tutorials/filters/DetectorContextTutorial.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python
+
+"""
+=========================
+DetectorContext tutorial
+=========================
+"""
+
+# %%
+# Some probabilistic trackers use detection probability and clutter density in their hypothesis
+# weights. DetectorContext allows these quantities to depend on the current hypothesis or
+# detection while preserving the existing scalar configuration path.
+
+import datetime
+
+import numpy as np
+
+from stonesoup.hypothesiser.distance import DistanceHypothesiser
+from stonesoup.hypothesiser.gaussianmixture import GaussianMixtureHypothesiser
+from stonesoup.measures import Mahalanobis
+from stonesoup.mixturereducer.gaussianmixture import GaussianMixtureReducer
+from stonesoup.models.measurement.linear import LinearGaussian
+from stonesoup.models.transition.linear import ConstantVelocity
+from stonesoup.predictor.kalman import KalmanPredictor
+from stonesoup.tracker.pointprocess import PointProcessMultiTargetTracker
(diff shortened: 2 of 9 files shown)


sylvesterkaczmarek:
There is an inconsistency between the scalar and callable `prob_detection` paths that can change the numerical semantics of the point-process update.

For a scalar, `SimpleDetectorContext.prob_detection()` always returns `Probability(self._prob_detection)`. For a callable, however, it returns the callable result unchanged. The tutorial/tests use callables returning ordinary floats such as `0.5`, so the same detection probability is a `Probability` in one configuration and a `float` in the other.

That matters here because component weights are `Probability` objects and normally retain Stone Soup's log-domain arithmetic. A float returned by the callable causes the multiplication to fall back to ordinary floating-point weights instead. It also means a callable returning an invalid negative probability can flow through as a negative component weight, whereas the equivalent scalar value would be rejected when constructing `Probability`.

Could the callable result also be normalised through `Probability(...)`, e.g. `return Probability(self._prob_detection(hypothesis))`? That would make scalar and callable contexts equivalent and preserve the existing weight representation. A regression comparing scalar `0.5` with `lambda _: 0.5`, plus an invalid negative callable, would pin this behaviour.

sylvesterkaczmarek (commented):
The scalar and callable prob_detection paths still have different semantics in the current implementation.

A scalar value is normalised through Probability(...), but a callable result is returned directly. That means lambda _: 0.5 produces a plain float while 0.5 produces a Probability, and an invalid callable result such as a negative value bypasses the validation the scalar path receives.

Since these values feed directly into point-process component weights, could the callable result also be wrapped with Probability(...)? A regression comparing scalar 0.5 with lambda _: 0.5, plus an invalid negative callable, would make the two configuration forms genuinely equivalent.

PR #1293, opened 2026-06-26
Iterated Kalman predictor

jmbarr:
Most straightforward version of an iterated predictor. Inherits from `ExtendedKalmanPredictor` and defaults to that (number of  iterations is 1 by default). Uses standard tests but may need an additional test to demonstrate its difference from `ExtendedKalmanPredictor`. TODO: should work out how not to use `copy.copy`.

### DIFF
--- a/stonesoup/predictor/kalman.py
+++ b/stonesoup/predictor/kalman.py
@@ -1,3 +1,4 @@
+from copy import copy
 from functools import partial
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


sdhiscocks (approved):

sdhiscocks:
> TODO: should work out how not to use `copy.copy`.

I think the shallow copy is fine.

PR #1295, opened 2026-07-03
Create generic sensor models

csherman-dstl:
Currently all the sensor models in Stone Soup are "Radar" Models with a lot of common logic to produce the measurement models and check if targets are visible.
This PR introduces "generic" sensor models for the 2D and 3D sensor which take in a Measurement Model Type to produce detections from.
Each of the Current Radar... Models have been Refactored to use these base classes in a way such that the functionality should remain identical.

Things to add to move this from draft:
- `is_clutter_detectable`: As clutter measurements are produced in the measurement space and then checked for if they are detectable it is harder to generalise this check across different sensor models. This will currently break if a non default measurement model is used due to the different measurement space
- type `class` is not serialisable by default and so storing a `GenericSensor` in a YAML file does not work

### DIFF
--- a/stonesoup/sensor/detection_probability.py
+++ b/stonesoup/sensor/detection_probability.py
@@ -1,0 +1,56 @@
+from abc import abstractmethod
+import numpy as np
+
+from stonesoup.base import Base, Property
+from stonesoup.types.array import StateVector
+
+
+class DetectionProbability(Base):
+    """Base class to define sensor detection probability"""
+    @abstractmethod
+    def __call__(self, state_vector: StateVector) -> float:
+        """
+        Method to determine the probability of detection based on the location of the target in
+        bearing, range space. Note for a 2D sensor the state vector will be [bearing, range] and
+        for a 3D sensor the state vector will be [elevation, bearing, range]
+
+        Parameters
+        ----------
+        state_vector : StateVector
+            location of the target in bearing range space
+
+        Returns
+        -------
+        float
+            The resulting probability of detection
+        """
+
+
+class ConstantDetectionProbability(DetectionProbability):
+    """:class:`.DetectionProbability` class to implement a constant probability of detection"""
+    p_d: float = Property(default=1., doc="The probaiblity of detection")
+
+    def __call__(self, *args, **kwargs) -> float:
+        return self.p_d
+
+
+class ExponentialDecayDetectionProbability(DetectionProbability):
+    """:class:`.DetectionProbability` class to implement a probability of detection of decays as
+    the range to the target increases"""
+    decay_rate: float = Property(default=np.inf, doc="The characteristic range of the "
+                                                     "exponential decay")
+
+    def __call__(self, state_vector, *args, **kwargs) -> float:
+        range = state_vector[-1, :].astype(float)
+        return np.exp(-range / self.decay_rate)
+
+
+class SigmoidDetectionProbability(DetectionProbability):
+    """:class:`.DetectionProbability` class to implement a probability of detection which observes
+    a sigmoid function"""
+    decay_rate: float = Property(doc="The rate at which the Sigmoid function decays")
+    midpoint: float = Property(doc="The range at which the probability of detection is 50%")
+
+    def __call__(self, state_vector, *args, **kwargs):
+        range = state_vector[-1, :].astype(float)
+        return 1 - 1/(1 + np.exp(-(range-self.midpoint)/self.decay_rate))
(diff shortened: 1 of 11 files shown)


gawebb-dstl:
Nice refactor overall and I’m all for removing a lot of the duplication between the radar sensors. This refactor moves the design further towards a couple of large base classes rather than smaller, more focused components. As a personal preference, I'd lean towards smaller classes and composition/mixins over monolithic sensor classes. For example, Generic2DSensor and Generic3DSensor seem to accumulate a lot of functionality and configuration in a single place. 

For example, RadarElevationBearingRange has 22 properties, while only a handful may be relevant for basic use. 
Something else that was introduced prior to this PR but could be fixed in this PR is that all of the radar sensors inherit from VisibilityInformed2DSensor whether they’re 2D or 3D. The obstacle properties/methods may be a little confusing for a 3D sensor and may imply that the obstacle code works in 3D.

I wonder whether this might benefit from being broken down into smaller capabilities that can be composed together. For example:
* Detectability (including FoV behaviour, visibility/obstacle modelling and max range)
* Clutter generation 
* Measurement generation 

Another thought is that many of the concrete radar sensor classes are becoming quite small. I wonder if factory/builder pattern could work where you specify if you want to include obstacles or a max value. 
These are mostly personal preferences rather than concerns with the implementation itself, so I don't see them as blocking, just observations.

stonesoup/sensor/detection_probability.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class ExponentialDecayDetectionProbability(DetectionProbability):
+    """:class:`.DetectionProbability` class to implement a probability of detection of decays as
+    the range to the target increases"""
+    decay_rate: float = Property(default=np.inf, doc="The characteristic range of the "
+                                                     "exponential decay")
+
+    def __call__(self, state_vector, *args, **kwargs) -> float:
+        range = state_vector[-1, :].astype(float)
```
gawebb-dstl, line 44: I think we should avoid hardcoding the index (-1) that is passed to the DetectionProbability model. I think a mapping should be added.

Also you should avoid naming a variable `range` as it masks the built-in range function

gawebb-dstl:
I've got some ideas about a refactor, let me know what you think

```

class GenericSensor(SimpleSensor):
    
    det_models: list[DetectionProbability] = Property()
    # Other properties
    
    def prob_of_detect(self, state):
        p_d = 1.0
        for det_mod in det_models:
            measurement_model = self.create_measurement_model(measurement_model_type=det_mod.measurement_model_type)
            state_vector_in_measurement_space =  measurement_model.function(state, noise=False)
            p_d *= det_mod.prod_of_detect(state_vector_in_measurement_space)
        return p_d
    
    def is_detectable(self, state, min_pd:float = 0 ) -> bool: # or _is_detectable
        p_d = prob_of_detect(state)
        return p_d > min_pd
        
    def is_detected(self, state) -> bool: # put in measure method
        p_d = prob_of_detect(state)
        return random_state.random() > p_d
        
    # Other methods
        
        
class HardLimitDetectionProbability(DetectionProbability):
    measurement_model_type: type[MeasurementModel]
    mapping: int
    min_value: float = Property(default=-np.inf)
    max_value: float = Property(default=np.inf)

    def __call__(self, state_vector, *args, **kwargs):
        value = state_vector[mapping, :].astype(float)
        return np.logical_and(min_value <= value, value <= max_value)

        
def MinMaxRange2D(min_range: float = 0, max_range: float = np.inf) -> DetectionProbability:
    return HardLimitDetectionProbability(
        measurement_model_type=CartesianToBearingRange,
        mapping=1,
        min_value=min_range,
        max_value=max_range
    )

    
def LimitFoV2D(fov_angle) -> DetectionProbability:
    fov_min = -fov_angle / 2
    fov_max = fov_angle / 2
    
    return HardLimitDetectionProbability(
        measurement_model_type=CartesianToBearingRange,
        mapping=0,
        min_value=fov_min,
        max_value=fov_max
    )

    
class RadarBearingRange(GenericSensor):
    
        det_models = kwargs.pop('det_models', None)
        
        max_range = kwargs.pop('max_range', None)
        if max_range is not None:
            deprecation_warning()
            det_models.append(MinMaxRange2D(max_range))
        ...
        
class VisibilityInformed2DSensor(DetectionProbability):
    # This almost works as a DetectionProbability
    # The only issue is that it needs a position.
```

sylvesterkaczmarek (commented):
I think the linear-model branch is currently unreachable: model_class is a class object, but the check uses isinstance(model_class, LinearGaussian). Should this be issubclass(...) instead?

PR #1301, opened 2026-07-20
Angle based measures

jswright-dstl:
This PR adds measures that focus on differences in angles by comparing the heading of one state with the heading of another. The options are `AngularDifference` that computes the angle difference between the two states, and `AngularDifferenceWithLeftRightAmbiguity` that calculates the absolute angle to the heading of `state2` from the heading of `state1` resulting in left-right ambiguity.

### DIFF
--- a/stonesoup/measures/state.py
+++ b/stonesoup/measures/state.py
@@ -548,12 +548,9 @@ class AngularDifference(Measure):
         state_vector1 = getattr(state1, "mean", state1.state_vector)
         state_vector2 = getattr(state2, "mean", state2.state_vector)
 
-        if self.mapping is not None:
-            angle1 = Angle(np.arctan2(*state_vector1[self.mapping, 0][::-1]))
-            angle2 = Angle(np.arctan2(*state_vector2[self.mapping, 0][::-1]))
-            return angle1 - angle2
-        else:
-            raise NotImplementedError('AngularDifference requires a mapping to be defined')
+        angle1 = Angle(np.arctan2(*state_vector1[self.mapping, 0][::-1]))
+        angle2 = Angle(np.arctan2(*state_vector2[self.mapping2, 0][::-1]))
+        return angle1 - angle2
 
 
 class AngularDifferenceWithLeftRightAmbiguity(AngularDifference):
--- a/stonesoup/measures/state.py
+++ b/stonesoup/measures/state.py
@@ -7,6 +7,7 @@ from scipy.spatial import distance
 
 from .base import BaseMeasure
 from ..base import Property
+from ..types.angle import Angle
 from ..types.state import State, ParticleState, GaussianState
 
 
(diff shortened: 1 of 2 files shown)


stonesoup/measures/state.py
```diff
@@ -506,3 +507,62 @@ def __call__(self, state1, state2):
...
+
+    The angular difference, :math:`\Delta\theta` between a pair of angles :math:`\theta_{1}` and
+    :math:`\theta_{2}` calculated from a mapping such as velocity is defined as:
+
+    .. math::
+        \theta_{1} = \arctan(\dot{y}_{1}, \dot{x}_{1}
+        \theta_{2} = \arctan(\dot{y}_{2}, \dot{x}_{2}
+        \Delta\theta = \theta_{1} - \theta_{2}
+    """
+
```
sdhiscocks, line 533: Should redefine mapping here if it's not optional.

```suggestion
    mapping: np.ndarray = Property(
        doc="Mapping array which specifies which elements within the"
            " state vectors are to be assessed as part of the measure"
    )

```

stonesoup/measures/state.py
```diff
@@ -506,3 +507,62 @@ def __call__(self, state1, state2):
...
+        Angle
+            Angular difference between two input :class:`~.State`
+
+        """
+        state_vector1 = getattr(state1, "mean", state1.state_vector)
+        state_vector2 = getattr(state2, "mean", state2.state_vector)
+
+        if self.mapping is not None:
+            angle1 = Angle(np.arctan2(*state_vector1[self.mapping, 0][::-1]))
+            angle2 = Angle(np.arctan2(*state_vector2[self.mapping, 0][::-1]))
```
sdhiscocks: Existing inherit `__init__` method should set `mapping2` to `mapping` if undefined, and allow setting separately, so just this line change should enable ability to have different mappings like other measures.

```suggestion
            angle2 = Angle(np.arctan2(*state_vector2[self.mapping2, 0][::-1]))
```

sdhiscocks (changes requested):
Some changes needed around mappings (I've not tested the suggested changes). Otherwise looks good 👍

stonesoup/measures/state.py
csherman-dstl: Does there also need to be a check that mappings are of length 2?

csherman-dstl (approved):

stonesoup/measures/tests/test_state.py
```diff
@@ -512,3 +513,38 @@ def test_gaussian_kld_raise_errors():
...
+    state1 = State(StateVector([1., 1.]))
+    state2 = State(StateVector([-1., 1.]))
+    measure = measures.AngularDifferenceWithLeftRightAmbiguity(mapping=[0, 1])
+
+    result1 = measure(state, state1)
+    result2 = measure(state, state2)
+
+    assert isinstance(result1, Angle)
+    assert isinstance(result2, Angle)
+    assert float(result2) == float(result2)
```
csherman-dstl:
```suggestion
    assert float(result1) == float(result2)
```
I assume its meant to be this?

csherman-dstl:
Only other consideration I have would be about whether it is worth using an implementation involving the dot product so that the same measure can be used for 2D and 3D velocities

sdhiscocks (approved):

PR #1303, opened 2026-07-21
Fixes for mappings and object arrays in measures

sdhiscocks:
Fixes issues with object arrays and handling of object arrays that will be deprecated in SciPy 1.20.

### DIFF
--- a/stonesoup/measures/state.py
+++ b/stonesoup/measures/state.py
@@ -3,7 +3,6 @@ from abc import abstractmethod
 from functools import lru_cache
 
 import numpy as np
-from scipy.spatial import distance
 
 from .base import BaseMeasure
 from ..base import Property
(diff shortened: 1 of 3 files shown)


jswright-dstl (approved):

stonesoup/measures/state.py
```diff
@@ -88,10 +87,24 @@ def __call__(self, state1, state2):
         state_vector2 = getattr(state2, 'mean', state2.state_vector)
 
         if self.mapping is not None:
-            return distance.euclidean(state_vector1[self.mapping, 0],
-                                      state_vector2[self.mapping2, 0])
+            if len(self.mapping) != len(self.mapping2):
+                raise ValueError(
+                    f"Shape mismatch between mapping and mapping2, "
+                    f"{len(self.mapping)} != {len(self.mapping2)}."
```
csherman-dstl: Does it make sense for this check to be in the base classes __init__ instead of constantly repeated? I know the other check has to be here.

csherman-dstl (approved):

stonesoup/measures/state.py
sdhiscocks: Yes, that would be better to catch it when creating an instance as well, and also means code can be in common `__init__` method.

PR #1307, opened 2026-07-29
Decaystate

jmbarr:
A non-linear transition model and accompanying state allowing one to model an active substance prone to decay (e.g. as in a radioactive substance).

### DIFF
--- a/stonesoup/models/transition/nonlinear.py
+++ b/stonesoup/models/transition/nonlinear.py
@@ -1,5 +1,6 @@
 import copy
 from collections.abc import Sequence
+from turtle import st
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


stonesoup/models/transition/nonlinear.py
```diff
@@ -210,3 +210,155 @@ def covar(self, time_interval, **kwargs):
...
+
+        Returns
+        -------
+        : :class:`~.StateVector`
+            The state vector after decay
+        """
+        out_state_vector = copy.copy(state.state_vector)
+        for i, element in enumerate(state.state_vector):
+            if element == 1 and np.random.uniform() < self.prob_decay(state, time_interval):
+                out_state_vector[i] *= 0
```
csherman-dstl:
```suggestion
        decay_postions = np.random.uniform(size=out_state_vector.shape) < self.prob_decay(state, time_interval)
        out_state_vector *= decay_positions
```
Does this achieve the same outcome?

stonesoup/types/state.py
```diff
@@ -1308,3 +1308,25 @@ def __len__(self):
 
 
 State.register(CompositeState)  # noqa: E305
+
+
+class DecayState(State):
+    """Embodies a state with elements (atoms) that can transition from their undecayed state (1) to
+    their decayed state (0). That is handled by the transition function. The state vector is of 1xN
+    StateVectors type and can be initialised either by specifying the number of atoms (N) or by
```
csherman-dstl: I assume this is meant to be a StateVectors state where the n `atoms` sit in the same state dimension but currently the state vector is n dimensional.

Is this correct?

stonesoup/types/state.py
```diff
@@ -1308,3 +1308,25 @@ def __len__(self):
...
+    """
+
+    halflife: float = Property(doc="The half life (timedelta object)")
+
+    def __init__(self, *args, **kwargs):
+
+        if type(args[0]) is int:
+            state_vector = StateVector(np.ones(args[0]), dtype=np.int8)
+        elif type(args[0]) is StateVector:
+            state_vector = args[0]
```
csherman-dstl:
```suggestion
    def __init__(self, state_vector, *args, **kwargs):

        if isinstance(state_vector, int):
            state_vector = StateVector(np.ones(state_vector), dtype=np.int8)
```
There is some issue where the first argument is neither an int or StateVector and so the `state_vector` variable is never set

stonesoup/models/transition/nonlinear.py
```diff
@@ -210,3 +210,155 @@ def covar(self, time_interval, **kwargs):
...
+
+    .. math::
+
+        p(1 \rightarrow 0) = 1 - \exp(- \lambda \Delta t)
+
+    where :math:`\lambda` is the decay constant, related to the half life, :math:`T` via,
+
+    .. math::
+
+        \lambda = \frac{\log_e 2}{T}
```
csherman-dstl: I know its not wrong but log_e just looks wrong

stonesoup/models/transition/nonlinear.py
```diff
@@ -210,3 +210,155 @@ def covar(self, time_interval, **kwargs):
...
+                else:
+                    prob *= self.prob_decay(state_ini, time_interval)
+            else:
+                if state_fin.state_vector[i] == 1:
+                    # This should never happen, but if it does, we'd want to kill the loop
+                    prob *= 0.
+                else:
+                    prob *= 1
+
+        return prob
```
csherman-dstl:
```suggestion
    prob = np.ones(shape=state.state_vector.shape)
    ini_ones = state_ini.state_vector == 1
    ini_zeros = state_ini.state_vector == 0
    fin_ones = state_fin.state_vector == 1
    fin_zeros = state_fin.state_vector == 0
    
    if np.any(fin_ones & ini_zeros):
        return 0.
    
    p_d = self.prob_decay(state_ini, time_interval)
    prob[fin_ones & ini_ones] = 1 - p_d
    prob[fin_zeros & ini_ones] = p_d
    return np.prod(prob)
```
Wrote a less iterative version, I think this gives the same answer

stonesoup/models/transition/nonlinear.py
jmbarr: Depends on how much of a mathematician vs engineer you are I suppose. Or how old. What do the kids use to unambiguously indicate a logarithm to the base e? 

\ln ?

stonesoup/models/transition/nonlinear.py
jmbarr: Apparently not. I now get an error that `decay_positions` is undefined.

jmbarr: Ah -- if you could spell 'positions'.

jmbarr: And indeed the `<` should  be '>' because you want the `True`== decayed positions to be the ones *0 in the subsequent operation.

stonesoup/models/transition/nonlinear.py
jmbarr: \ln it is

stonesoup/types/state.py
jmbarr: Should say Nx1 `StateVector` rather than 1xN `StateVectors` (now corrected). This could be augmented into a NxM `ParticleDecayState` with M particles, but that's for another day/application.

sylvesterkaczmarek (commented):
The covariance construction looks dimensionally wrong for more than one atom. state_vector is N×1, so np.diag(state.state_vector * covar) extracts a one-element diagonal rather than creating an N×N matrix. Could this flatten first before np.diag()?

PR #1309, opened 2026-08-04
Modernise Video Processing demo with Ultralytics YOLO detector

sdhiscocks:
Replace the demo's aging building blocks with actively-maintained libraries, and add a reusable YOLO detector component.

- Add UltralyticsBoxObjectDetector (stonesoup/detector/ultralytics.py), mirroring TensorFlowBoxObjectDetector: same [x, y, w, h] state vector and raw_box/class/score metadata, so downstream feeders are unchanged. The TensorFlow detector is retained as an alternative.
- Update VideoClipReader for the MoviePy 2.x API (moviepy.editor removed, subclip -> subclipped, fl_image -> image_transform), which the previous code broke against.
- Rewrite the Video Processing demo to use the YOLO detector (weights auto-download, replacing the ~500 MB TensorFlow model and pbtxt label map) and yt-dlp in place of the unmaintained pytube.
- pyproject: add ultralytics extra, add yt-dlp and pin moviepy>=2.0 in the video extra; add ultralytics to autodoc_mock_imports.


csherman-dstl (approved):

jswright-dstl (approved):

sglvladi:
Looks good and approve!👍My only comment would be: Should we have a group of Video Processing demos, and showcase both capabilities? Conscious that this would potentially add to the maintenance overhead, so happy to be brushed off!

sdhiscocks:
> Looks good and approve!👍My only comment would be: Should we have a group of Video Processing demos, and showcase both capabilities? Conscious that this would potentially add to the maintenance overhead, so happy to be brushed off!

I think keeping just the one example for reasons you mention, and obviously our focus is on the tracking, rather than the detection part.

PR #1311, opened 2026-08-07
Fix GNN associators with infinite missed distance

### LINKED ISSUE #824: Data association with Global Nearest Neighbor Filter not working as expected
I've found that using a global nearest neighbor filter as a data associator in my own code creates an unexpected effect. 

Basically, the Global Nearest Neighbor associates the measurement to the first track, rather than the track with the smallest distance. 

In reality, this might just be user error, but I wanted to make an issue demonstrating what I'm seeing so that I can be sure this is not a problem with Stone Soup. 

See the attached files and script as a demonstration of what I mean (you may have to change the path in the script, so that it points at the pickle files correctly, depending on where you put them). 
[issueDemonstrationFiles.zip](https://github.com/dstl/Stone-Soup/files/11931646/issueDemonstrationFiles.zip)

sdhiscocks:
The default missed distance of DistanceHypothesiser is infinite. Where a missed detection is unavoidable, such as when there are more tracks than gated detections, every joint hypothesis then has an infinite total distance. GlobalNearestNeighbour could not distinguish between these, so returned an arbitrary (first enumerated) association, and GNNWith2DAssignment raised "Assignment was not feasible".

Infinite missed detection distances are now replaced with a distance marginally greater than the greatest finite distance, keeping a missed detection less favourable than any detection, whilst allowing joint hypotheses to be compared.

Fixes #824

### DIFF
--- a/stonesoup/dataassociator/neighbour.py
+++ b/stonesoup/dataassociator/neighbour.py
@@ -1,4 +1,5 @@
 import itertools
+from math import isfinite
 
 import numpy as np
 from scipy.optimize import linear_sum_assignment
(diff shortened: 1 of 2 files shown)


jswright-dstl (approved):

stonesoup/dataassociator/neighbour.py
```diff
@@ -1,13 +1,58 @@
...
+    for track_hypotheses in hypotheses.values():
+        for hypothesis in track_hypotheses:
+            if not isinstance(hypothesis, SingleDistanceHypothesis):
+                continue
+            distance = hypothesis.distance
+            if not hypothesis:
+                if not isfinite(distance):
+                    missed_hypotheses.append(hypothesis)
+            elif distance > max_distance and isfinite(distance):
+                max_distance = distance
```
csherman-dstl, line 48:
```suggestion
            if not hypothesis:
                if not isfinite(distance):
                    missed_hypotheses.append(hypothesis)
            elif distance > max_distance:
                max_distance = distance
```
or
```suggestion
            if not isfinite(distance):
                missed_hypotheses.append(hypothesis)
            elif distance > max_distance:
                max_distance = distance
```

I feel like it should be one of these options. Either a true hypothesis can have infinite distance (and therefore we need to handle it) or it can't (and so we dont need to check).

csherman-dstl (approved):

Issue #1160, opened 2025-04-10
`MultiMeasurementInitiator` does not handle `MultipleHypothesis`


sgboakes:
If you want to use a data associator that returns a `MultipleHypothesis` (e.g. `PDAHypothesiser`) with the `MultiMeasurementInitiator`, the code will break regardless of whether the initiator chosen handles the `MultipleHypothesis` type. The following code from `MultiMeasurementInitiator` contains the issue, where it is assumed that `associations` contains `SingleHypothesis`  types:

```
if self.holding_tracks:
    associations = self.data_associator.associate(
        self.holding_tracks, detections, timestamp, **kwargs)

    for track, hypothesis in associations.items():
        if hypothesis:
            state_post = self.updater.update(hypothesis)
            track.append(state_post)
            associated_detections.add(hypothesis.measurement)
        else:
            track.append(hypothesis.prediction)

        if sum(1 for state in track if not self.updates_only or isinstance(state, Update))\
                >= self.min_points:
            sure_tracks.add(track)
            self.holding_tracks.remove(track)

    self.holding_tracks -= self.deleter.delete_tracks(self.holding_tracks)

self.holding_tracks |= self.initiator.initiate(
    detections - associated_detections, timestamp)

return sure_tracks
```

When generating holding tracks in the `MultiMeasurementInitiator`, we should determine the type of associations generated by the data associator, and handle both `SingleHypothesis` and `MultipleHypothesis` types by looping through the hypotheses in `MultipleHypothesis` if required.

PR #1313, opened 2026-08-13
Support for MultipleHypothesis in MultiMeasurementInitiator

AaronSeel:
This PR includes an addition to the MultiMeasurementInitiator to handle MultipleHypothesis instead of just SingleHypothesis as a fix to issue #1160 . 
Otherwise it is not possible to work with initiators when using the MultiTargetMixtureTracker with a DataAssociator returning a MultipleHypothesis like the PDAHypothesiser. This can be done by typechecking for a MultipleHypothesis in the assiciations and iterating over it if necessary.

### DIFF
--- a/stonesoup/initiator/simple.py
+++ b/stonesoup/initiator/simple.py
@@ -18,7 +18,7 @@ from ..types.state import State, GaussianState, ParticleState, TaggedWeightedGau
     ASDGaussianState, EnsembleState
 from ..types.track import Track
 from ..types.update import ParticleStateUpdate, Update, \
-    GaussianMixtureUpdate, ASDGaussianStateUpdate, EnsembleStateUpdate
+    GaussianMixtureUpdate, GaussianStateUpdate, ASDGaussianStateUpdate, EnsembleStateUpdate
 from ..updater import Updater
 from ..updater.kalman import ExtendedKalmanUpdater
 
@@ -309,7 +309,7 @@ class MultiMeasurementInitiator(GaussianInitiator):
                         weights = np.asarray(posterior_state_weights)
 
                         post_mean, post_covar = gm_reduce_single(means, covars, weights)
-                        track.append(GaussianState(
+                        track.append(GaussianStateUpdate(
                             post_mean, post_covar,
                             multihypothesis,
                             multihypothesis[0].measurement.timestamp))
--- a/stonesoup/initiator/simple.py
+++ b/stonesoup/initiator/simple.py
@@ -14,6 +14,7 @@ from ..types.hypothesis import SingleHypothesis
 from ..types.mixture import GaussianMixture
 from ..types.multihypothesis import MultipleHypothesis
 from ..types.numeric import Probability
+from ..types.prediction import GaussianStatePrediction
 from ..types.state import State, GaussianState, ParticleState, TaggedWeightedGaussianState, \
     ASDGaussianState, EnsembleState
 from ..types.track import Track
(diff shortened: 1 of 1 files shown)


stonesoup/initiator/simple.py
```diff
@@ -288,13 +289,20 @@ def initiate(self, detections, timestamp, **kwargs):
...
                 self.holding_tracks, detections, timestamp)
 
-            for track, hypothesis in associations.items():
-                if hypothesis:
-                    state_post = self.updater.update(hypothesis)
-                    track.append(state_post)
-                    associated_detections.add(hypothesis.measurement)
+            for track, multihypothesis in associations.items():
+                if multihypothesis:
+                    if multihypothesis is MultipleHypothesis:
```
sdhiscocks:
```suggestion
                    if isinstance(multihypothesis, MultipleHypothesis):
```

stonesoup/initiator/simple.py
```diff
@@ -288,13 +289,20 @@ def initiate(self, detections, timestamp, **kwargs):
...
+                        for hypothesis in multihypothesis:
+                            if not hypothesis:
+                                continue
+                            track.append(self.updater.update(hypothesis))
+                            associated_detections.add(hypothesis.measurement)
```
sdhiscocks: This isn't quite right, as this isn't factoring in weights, etc. and something more like this is required: https://github.com/dstl/Stone-Soup/blob/50e965aa15ec63d1d3eb13bb275639d136d8557b/stonesoup/tracker/simple.py#L262-L282

Or possibly just using the [PDAUpdater](https://stonesoup.readthedocs.io/en/v1.9.1/stonesoup.updater.html#stonesoup.updater.probability.PDAUpdater) would work as that accepts multi-hypothesis.

sdhiscocks (changes requested):
Thanks for contribution @AaronSeel.

stonesoup/initiator/simple.py
AaronSeel: Thank you for your time and reply. 
In the newest commit I tried to apply your propositions. Is this the way you were imagining it?
After reducing the Gaussian Mixture to a Gaussian State it gets appended to the current track.

sylvesterkaczmarek:
There looks to be a type/construction regression in the new `MultipleHypothesis` path.

The mixture-reduction logic is based on the existing PDA tracker, but that implementation appends a `GaussianStateUpdate(post_mean, post_covar, multihypothesis, timestamp)`. Here the same four arguments are passed to `GaussianState(...)`.

`GaussianState` has no `hypothesis` property; that belongs to `Update` / `GaussianStateUpdate`. So this path either cannot construct the state correctly or loses the update/hypothesis semantics that `updates_only` and downstream track logic rely on.

Could this use `GaussianStateUpdate` here, as the existing PDA tracker does? It would also be useful to add a regression that actually drives a `MultipleHypothesis` through `MultiMeasurementInitiator`, since this PR currently changes no tests and that would catch the constructor/type mismatch directly.

sylvesterkaczmarek (commented):
The GaussianStateUpdate construction issue from my earlier comment is fixed in the current head. There still seems to be one behavioural mismatch with the existing PDA tracker logic, though.

MultiTargetMixtureTracker only treats a detection as associated when its hypothesis weight exceeds the missed-detection weight, and it appends a GaussianStatePrediction when no measurement hypothesis beats the missed detection. Here every non-null hypothesis is added to associated_detections, and the reduced mixture is always appended as a GaussianStateUpdate.

With PDA, that means even a very weak measurement hypothesis can suppress initiation from that detection and can count as an update toward min_points / updates_only. Could this mirror the existing tracker semantics by comparing against the missed-detection weight before marking detections associated and choosing update versus prediction? A regression where the missed-detection hypothesis dominates would be useful to pin this behaviour.

AaronSeel:
Thank you for your input, that is something I did miss before. 

Now the behavior should be the same as in the MultiTargetMixtureTracker for either appending a GaussianStateUpdate or a GaussianStatePrediction.

PR #1315, opened 2026-08-16
Fix PCRB multiple sensor location handling

sylvesterkaczmarek:
## Summary

Fixes PCRB information-matrix calculation when more than one sensor location is supplied.

`StateVectors[:, i]` does not preserve the `StateVector` shape for a single column in this case. The resulting one-dimensional value can be assigned to nonlinear measurement models such as `CartesianToBearingRange`, where it broadcasts against the state vector and produces an incorrect Jacobian/information contribution.

Each sensor location is now explicitly normalized to `StateVector` before assigning it to a copied measurement model.

A regression test uses two nonlinear bearing/range sensor locations and verifies the combined information matrix numerically.

Addresses #1269

## Validation

- Regression covers two sensor locations.
- Uses `CartesianToBearingRange`, exercising `translation_offset`.
- Expected combined information matrix is asserted numerically.
- Existing single-sensor behaviour is unchanged.
- Patch is isolated to the PCRB implementation and its unit tests.

## Scope

This PR intentionally fixes the existing multiple-location `StateVector` shape bug only. Support for heterogeneous measurement models and mobile sensor histories is broader design work and is not introduced here.

### DIFF
--- a/stonesoup/metricgenerator/pcrbmetric.py
+++ b/stonesoup/metricgenerator/pcrbmetric.py
@@ -8,7 +8,7 @@ from .base import MetricGenerator
 from ..base import Property
 from ..types.state import GaussianState
 from ..types.groundtruth import GroundTruthState, GroundTruthPath
-from ..types.array import StateVectors
+from ..types.array import StateVector, StateVectors
 from ..models.transition import TransitionModel
 from ..models.measurement import MeasurementModel
 from ..types.metric import TimeRangeMetric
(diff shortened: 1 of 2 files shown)


sdhiscocks:
There is the issue of handling different measurement models and mobile sensors that needs addressing, so think this needs more thought.

sylvesterkaczmarek:
Thanks. Agreed that fixing only the `StateVector` shape leaves the broader sensor model ambiguous. I think the cleanest direction is to model each sensor together with its measurement model and location history rather than extending parallel `sensor_locations` and `measurement_models` inputs independently. 

Would you prefer `PCRBMetric` to accept actual `Sensor` instances, so each sensor provides its own measurement model and platform/movement history, or should the existing arguments be extended to support sequences?

sylvesterkaczmarek:
I have narrowed the PR scope to the existing multiple-location shape bug and changed the issue reference to `Addresses #1269`. The broader heterogeneous-measurement/mobile-sensor design is now explicitly left out of this patch. If that scope works for you, this should be reviewable as the isolated bug fix.

PR #1318, opened 2026-08-18
Fix threshold-aware one-to-one assignment

### LINKED ISSUE #895: Non-optimal results from the `OneToOneAssociator` when an `association_threshold` is used
You can get non-optimal results from the `OneToOneAssociator` when an `association_threshold` is used. See the code below

```
from stonesoup.dataassociator.general import OneToOneAssociator


def print_results(assoc_dict_print, unassocs_a_print, unassocs_b_print, measure):
    print(assoc_dict_print, "\n", unassocs_a_print, "\n", unassocs_b_print)
    total = sum(measure(obj_1, obj_2) for obj_1, obj_2 in assoc_dict_print.items())
    print(f"Total Measure={total}")


def multiply(item1: float, item2: float) -> float:
    return item1 * item2


numbers_a = (1, 2, 3, 5)
numbers_b = (1, 2, 4, 7)

associator = OneToOneAssociator(measure=multiply,
                                maximise_measure=True,
                                association_threshold=5)

all_assoc_dict = associator.association_dict(numbers_a, numbers_b)
assoc_dict = {a: all_assoc_dict[a]
              for a in numbers_a
              if all_assoc_dict[a] is not None}

unassocs_a = {x for x in numbers_a if all_assoc_dict[x] is None}
unassocs_b = {x for x in numbers_b if all_assoc_dict[x] is None}

print("1-2-1 Output")
print_results(assoc_dict, unassocs_a, unassocs_b, multiply)

# {3: 4, 5: 7} 
#  {1, 2} 
#  {1, 2}
# Total Measure=47

print("Better output")
better_output = {2: 4, 3: 2, 5: 7}
print_results(better_output, {1}, {1}, multiply)

# Total Measure=49
```

sylvesterkaczmarek:
## Summary

Fixes the non-optimal association behaviour reported in #895 when `OneToOneAssociator` is used with an `association_threshold`.

The current implementation solves a complete assignment over real object pairs first, then removes pairs that fail the threshold. A threshold-failing pair can therefore consume a row or column during optimisation and prevent a better feasible partial assignment from being selected.

## Change

- Add an explicit unassociated choice for each object in the first collection at the existing `fail_value`.
- Run `linear_sum_assignment` on the augmented matrix so threshold gating is part of the optimisation itself.
- Ignore dummy assignments when building the returned `AssociationSet`.
- Preserve the existing threshold semantics and support both maximise and minimise modes.

## Regression coverage

Adds a focused regression test based on the exact example in #895:

- the previous result has total measure 47;
- the feasible partial assignment has total measure 49;
- the expected pairs are selected and the remaining objects are reported as unassociated.

The same regression is also exercised in minimise mode using the equivalent negated measure.
- Added coverage for a fully gated assignment and for unequal collection sizes.

Closes #895

### DIFF
--- a/stonesoup/dataassociator/general.py
+++ b/stonesoup/dataassociator/general.py
@@ -76,32 +76,48 @@ class OneToOneAssociator(Associator):
             for j, b in enumerate(list_of_bs):
                 distance_matrix[i, j] = self.individual_weighting(a, b)
 
-        # Use "shortest path" assignment algorithm on distance matrix
-        # to assign tracks to nearest detection
-        # Maximise flag = true for probability instance
-        # (converts minimisation problem to maximisation problem)
-        row_ind, col_ind = linear_sum_assignment(
-            distance_matrix, self.maximise_measure)
+        if self.maximise_measure:
+            valid_matrix = distance_matrix > self.association_threshold
+            preference_matrix = distance_matrix
+        else:
+            valid_matrix = distance_matrix < self.association_threshold
+            preference_matrix = -distance_matrix
+
+        # Make leaving an object unassociated part of the optimisation rather than assigning all
+        # possible real pairs first and applying the threshold afterwards. The transformed score
+        # gives every valid pair a cardinality bonus larger than the maximum possible difference
+        # between equal-cardinality matchings. This therefore maximises the number of valid pairs
+        # first, then preserves the original maximise/minimise objective within that cardinality.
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
Added two focused edge-case regressions around the threshold-aware assignment: all candidate pairs gated out, and rectangular collections where only the maximum feasible number can be associated. The focused test file now passes 4/4.

PR #1321, opened 2026-08-18
Fix OpenSky zero-valued position filtering

sylvesterkaczmarek:
## Summary

Fixes OpenSky state filtering so valid zero-valued longitude, latitude, and geometric altitude values are retained.

`_OpenSkyNetworkReader.data_gen()` currently checks position availability with `all(...)`. That treats numeric zero as missing, so an otherwise valid state is dropped whenever longitude, latitude, or geometric altitude is exactly `0.0`.

This is particularly visible around the Greenwich meridian and equator, and can also affect zero-altitude values.

## Change

- Treat only `None` as missing for longitude, latitude, and geometric altitude.
- Preserve existing rejection of incomplete position records.
- Leave on-ground filtering, timestamp handling, metadata, and state-vector layout unchanged.

## Regression coverage

Adds an offline unit test with a mocked OpenSky response that verifies:

- a state at longitude `0.0`, latitude `0.0`, and geometric altitude `0.0` is accepted;
- a state with a genuinely missing (`None`) longitude is still rejected;
- the accepted state retains the expected `[longitude, latitude, altitude]` state vector and metadata.

No network access is required for the regression test.

### DIFF
--- a/stonesoup/reader/opensky.py
+++ b/stonesoup/reader/opensky.py
@@ -79,8 +79,8 @@ class _OpenSkyNetworkReader(Reader):
                 for state in data['states']:
                     if state[8]:  # On ground
                         continue
-                    # Must have position (lon, lat, geo-alt)
-                    if not all(state[index] for index in (5, 6, 13)):
+                    # Must have position (lon, lat, geo-alt). Zero is a valid coordinate/value.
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
This remains current and mergeable with no failing checks. Could a maintainer review this when convenient?

PR #1323, opened 2026-08-18
Handle SIAP metrics across track state gaps

### LINKED ISSUE #1103: SIAP metrics do not support gaps in detections
# 🐛 Bug Report

## Description
When no object is detected at a given timestamp, there is no track for this timestamp even if there is a groundtruth. This should be supported by the Accuracy metrics, but it produces an error.

### Current Behavior
When there is a gap in the detections, the function accuracy_at_time tries to get a non existing timestamp key from the tracks and returns an error.

```
{
	"name": "IndexError",
	"message": "timestamp not found in states",
	"stack": "---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[13], line 1
----> 1 metrics = metric_manager.generate_metrics()

File ~/Library/Caches/pypoetry/virtualenvs/cosita-K1LlpWRX-py3.12/lib/python3.12/site-packages/stonesoup/metricgenerator/manager.py:91, in MultiManager.generate_metrics(self)
     89     print(\"appel self.associate_tracks(generator)\")
     90     self.associate_tracks(generator)
---> 91 metric_list = generator.compute_metric(self)
     92 if not isinstance(metric_list, list):  # If not already a list, force it to be one
     93     metric_list = [metric_list]

File ~/Library/Caches/pypoetry/virtualenvs/cosita-K1LlpWRX-py3.12/lib/python3.12/site-packages/stonesoup/metricgenerator/tracktotruthmetrics.py:114, in SIAPMetrics.compute_metric(self, manager, **kwargs)
    112 Nt = self.num_tracks_at_time(tracks, timestamp)
    113 N_sum += Nt
--> 114 PAt = self.accuracy_at_time(manager, timestamp, self.position_measure)
    115 PA_sum += PAt
    116 VAt = self.accuracy_at_time(manager, timestamp, self.velocity_measure)

File ~/Library/Caches/pypoetry/virtualenvs/cosita-K1LlpWRX-py3.12/lib/python3.12/site-packages/stonesoup/metricgenerator/tracktotruthmetrics.py:327, in SIAPMetrics.accuracy_at_time(self, manager, timestamp, measure)
    325 for association in associations:
    326     truth, track = self.truth_track_from_association(association)
--> 327     error_sum += measure(track[timestamp], truth[timestamp])
    328 return error_sum

File ~/Library/Caches/pypoetry/virtualenvs/cosita-K1LlpWRX-py3.12/lib/python3.12/site-packages/stonesoup/types/state.py:307, in StateMutableSequence.__getitem__(self, index)
    305             return state
    306     else:
--> 307         raise IndexError('timestamp not found in states')
    308 elif isinstance(index, slice):
    309     return StateMutableSequence(self.states.__getitem__(index))

IndexError: timestamp not found in states"
}
```

### Expected Behavior
The accuracy metrics should ignore timestamps where the track is not defined. Additionnaly the function num_tracks_at_time should take into account timestamps when the track is associated to a truth in terms of time interval, but not defined at the given timestamp.

---

## Steps to Reproduce
Provide the steps to reproduce the problem:
1. Download the notebook https://stonesoup.readthedocs.io/en/v1.3/auto_examples/metrics/Metrics.html
2. Replace the detector, suppressing a few detections:

```python
class PartialPlatformDetectionSimulator(PlatformDetectionSimulator):
    @BufferedGenerator.generator_method
    def detections_gen(self):
        for time, truths in self.groundtruth:
            print(time, time.second % 10)
            if time.second % 10 == 0:
                continue
            # Move platforms and carry out sensor actions.
            for platform in self.platforms:
                platform.move(time)
                for sensor in platform.sensors:
                    sensor.act(time)

            # Make measurements from sensors
            for platform in self.platforms:
                for sensor in platform.sensors:
                    truths_to_be_measured = truths.union(self.platforms) - {platform}
                    detections = sensor.measure(truths_to_be_measured)
                    yield time, detections

detector = PartialPlatformDetectionSimulator(ground_truth_gen, platforms=[platform])
```

---

### Environment
- **Operating System**: macOS Sonoma 14
- **Stonesoup Version**: 1.4

sylvesterkaczmarek:
## Summary

Fixes SIAP metric generation when an association time range spans a timestamp for which the associated `Track` has no state, addressing #1103.

`AssociationSet.associations_at_timestamp()` can legitimately return an association based on its time range even when the track itself has a gap at that timestamp. The SIAP implementation previously treated that association as an available track and then indexed `track[timestamp]`, which raises `IndexError: timestamp not found in states`.

## Change

- Count an associated track in `NA(t)` only when the track actually contains a state at the requested timestamp.
- Skip kinematic accuracy contributions when either side of an association has no state at that timestamp.
- Keep the association time-range model unchanged. The fix is local to SIAP metric calculation.

This keeps the denominator used for positional and velocity accuracy aligned with the states that can actually contribute an error measurement.

## Regression coverage

Adds a focused test with:

- a truth state at `t0`, `t1`, and `t2`;
- a track state at `t0` and `t2`, with an intentional gap at `t1`;
- one `TimeRangeAssociation` spanning `t0` through `t2`.

The regression verifies that at `t1`:

- the track is not counted as present;
- it is not counted as an associated track for SIAP accuracy;
- positional accuracy contributes zero rather than raising an exception;
- full `compute_metric()` completes and reports zero positional accuracy for the gap timestamp.

Existing behaviour is unchanged at timestamps where track and truth states are present.

Closes #1103

### DIFF
--- a/stonesoup/metricgenerator/tests/test_siap_track_gaps.py
+++ b/stonesoup/metricgenerator/tests/test_siap_track_gaps.py
@@ -1,0 +1,56 @@
+import datetime
+
+from ...measures import Euclidean
+from ...types.association import AssociationSet, TimeRangeAssociation
+from ...types.groundtruth import GroundTruthPath, GroundTruthState
+from ...types.state import State
+from ...types.time import TimeRange
+from ...types.track import Track
+from ..manager import MultiManager
+from ..tracktotruthmetrics import SIAPMetrics
+
+
+def test_siap_ignores_association_timestamp_without_track_state():
+    t0 = datetime.datetime(2024, 1, 1, 0, 0, 0)
+    t1 = t0 + datetime.timedelta(seconds=1)
+    t2 = t0 + datetime.timedelta(seconds=2)
+
+    truth = GroundTruthPath([
+        GroundTruthState([0, 0, 0, 0], timestamp=t0),
+        GroundTruthState([1, 0, 1, 0], timestamp=t1),
+        GroundTruthState([2, 0, 2, 0], timestamp=t2),
+    ])
+    track = Track([
+        State([0.1, 0, 0.1, 0], timestamp=t0),
+        State([2.1, 0, 2.1, 0], timestamp=t2),
+    ])
+
+    manager = MultiManager()
+    manager.add_data({
+        'groundtruth_paths': [truth],
+        'tracks': [track],
+    })
+    manager.association_set = AssociationSet({
+        TimeRangeAssociation(
+            {truth, track},
+            time_range=TimeRange(t0, t2),
+        )
+    })
+
+    metric = SIAPMetrics(
+        position_measure=Euclidean((0, 2)),
+        velocity_measure=Euclidean((1, 3)),
+    )
+    manager.generators = [metric]
+
+    assert metric.num_tracks_at_time([track], t1) == 0
+    assert metric.num_associated_tracks_at_time(manager, [track], t1) == 0
+    assert metric.accuracy_at_time(manager, t1, metric.position_measure) == 0
+
+    metrics = metric.compute_metric(manager)
+    position_at_times = next(
+        item for item in metrics if item.title == 'SIAP Position Accuracy at times'
+    )
+    gap_metric = next(item for item in position_at_times.value if item.timestamp == t1)
+
+    assert gap_metric.value == 0
--- a/stonesoup/metricgenerator/tests/test_siap_track_gaps.py
+++ b/stonesoup/metricgenerator/tests/test_siap_track_gaps.py
@@ -1,5 +1,7 @@
 import datetime
 
+import pytest
+
 from ...measures import Euclidean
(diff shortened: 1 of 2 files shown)


sdhiscocks:
A common approach used to resolve this which is preferable, is to [interpolate tracks or truth](https://github.com/dstl/Stone-Soup/blob/ec032a9f7ba73c9cc34d1c4f7824af8cab8fc0f1/stonesoup/functions/interpolate.py) (depending on situation).

sylvesterkaczmarek:
Thanks. I've reworked this to use `interpolate_state_mutable_sequence` for missing track/truth states, and updated the counts so they remain consistent with the interpolated accuracy calculation. I've also added regression coverage for both track and truth gaps.

sylvesterkaczmarek:
I updated this to use the existing interpolation approach you suggested and added regression coverage for both track and truth gaps. Could you take another look when convenient?

PR #1324, opened 2026-08-18
Support streamed DataFrames in pandas readers

### LINKED ISSUE #971: Pandas DataFrame Reader, additional example.
Is it possible to document an example for the Pandas Dataframe reader that reads files from folder that has continue new data files added?

https://stonesoup.readthedocs.io/en/v1.2/auto_examples/Custom_Pandas_Dataloader.html#dataframe-detection-reader

sylvesterkaczmarek:
## Summary

Addresses #971 by allowing Stone Soup's pandas readers to consume either a single `pandas.DataFrame` or an iterable that yields DataFrames over time.

The existing `_DataFrameReader` requires the complete DataFrame to be available up front. That works well for in-memory datasets, but it prevents the same reader classes from being used naturally with batch-oriented sources such as a directory where new files are added, database result batches, or another generator that produces DataFrames incrementally.

## Change

- Extend `_DataFrameReader.dataframe` to accept either one DataFrame or an iterable of DataFrames.
- Consume iterable inputs lazily, one DataFrame at a time, and yield their records into the existing `_DictReader` pipeline.
- Validate every yielded chunk and raise a clear `TypeError` if the iterable produces something other than a DataFrame.
- Preserve the existing single-DataFrame path unchanged.
- Document that rows across yielded DataFrames must remain in time order, matching the ordering requirement of the existing dictionary-reader implementation.
- Apply the capability consistently to `DataFrameDetectionReader`, `DataFrameGroundTruthReader`, and `DataFrameTrackReader` through their shared base class.

## Regression coverage

Adds focused tests showing that:

- a generator yielding two DataFrames is accepted;
- a timestamp split across the DataFrame boundary is still grouped into a single reader timestep;
- all records are retained with the expected state vectors;
- a non-DataFrame chunk fails with an explicit error.

## Documentation

Adds a Sphinx-Gallery example using two CSV files in a temporary directory. A DataFrame generator reads the files sequentially and feeds them directly into `DataFrameDetectionReader`, demonstrating the pattern requested in #971 without introducing file-watching behaviour into the reader itself.

The example also makes the ordering contract explicit: the producer is responsible for yielding chronological batches; the reader does not buffer or reorder arbitrary out-of-order input.

Closes #971

### DIFF
--- a/docs/examples/readers/Streaming_Pandas_DataFrames.py
+++ b/docs/examples/readers/Streaming_Pandas_DataFrames.py
@@ -1,0 +1,73 @@
+"""
+Streaming pandas DataFrames
+===========================
+
+Stone Soup's pandas readers can consume an iterable of DataFrames as well as one DataFrame held
+entirely in memory. This is useful when data arrives in batches, for example as CSV files added to
+a directory over time.
+
+The reader consumes the iterable lazily. Each yielded DataFrame must use the same schema expected
+by the reader and the sequence of rows across all DataFrames must be in time order.
+
+This example uses a finite temporary directory so it is deterministic when built by Sphinx-Gallery.
+In a live application, ``dataframe_batches`` can be replaced by a directory watcher, message-queue
+consumer, database cursor, or another generator that yields new DataFrames as data arrives.
+"""
+
+# %%
+# Create two small CSV batches. The second batch deliberately starts at the same timestamp at which
+# the first batch ends, demonstrating that detections are still grouped by timestamp across a
+# DataFrame boundary.
+import datetime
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek:
I traced the docs failure to the documentation extension expecting property types that expose either the `typing` API or a concrete `__name__`; the PEP 604 union (`DataFrame | Iterable[...]`) is a `types.UnionType` and caused the build exception. I changed the annotation to `typing.Union[...]`, preserving the same API intent while making it compatible with the existing docs extension. Fresh CI should now exercise the fix.

PR #1327, opened 2026-08-18
Document plotting when running examples as scripts

### LINKED ISSUE #739: Plotter & Plotterly figures do not display automatically when scripts/functions are ran from the command line/IDE
Figures from the `Plotter` and `Plotterly` class don't automatically appear when the scripts are ran in Pycharm or the command line. I expect they won't display when ran from other IDEs.

For each the figures can made to display:
For the `Plotterly` class using: `plotter.fig.show()` instead of `plotter.fig` makes the figures appear. However using `plotter.fig.show()` in the tutorials breaks the Sphinx docs. 
For the `Plotter` class, `plt.show()` is needed at the end of the script to force the figures to appear. 

I feel this behaviour could trip up people who are new to Stone-Soup when they try to run the tutorials or examples

sylvesterkaczmarek:
## Summary

Adds explicit guidance for displaying Stone Soup plots when tutorials, examples or demonstrations are run as normal Python scripts or from an IDE, addressing #739.

Notebook frontends and Sphinx-Gallery render plotting objects as part of their own display pipeline. A standalone Python process does not necessarily do that automatically, so users can run an example successfully and still see no figure.

## Change

Adds a **Running examples outside notebooks** documentation page covering both plotting backends:

- `Plotter` / Matplotlib: call `plt.show()` after constructing the figure;
- `Plotterly` / Plotly: call `plotter.fig.show()`;
- IDEs where Plotly's default renderer is unsuitable: explicitly select the `browser` renderer;
- environments where interactive rendering is unreliable: write a standalone HTML file with `plotter.fig.write_html(...)`.

The page also explains why these explicit display calls should not simply be inserted into all documentation examples: Sphinx-Gallery and notebooks already manage rendering themselves.

The page is linked from the main documentation toctree so users who download or copy an example can find the guidance without needing to search issue history.

## Scope

Documentation only. No plotting API or backend behaviour is changed.

Closes #739

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -126,6 +126,7 @@ Contents
     auto_tutorials/index
     auto_examples/index
     auto_demos/index
+    running_examples
     contributing
     copyright
 
--- a/docs/source/running_examples.rst
+++ b/docs/source/running_examples.rst
@@ -1,0 +1,64 @@
+Running examples outside notebooks
+==================================
+
+Stone Soup's tutorials, examples and demonstrations are built primarily for
+Sphinx-Gallery and notebook-style use. When the same code is run as a normal
+Python script or from an IDE, the plotting backend may not display the final
+figure automatically. The plotting object is still created; it just needs an
+explicit display step appropriate to the backend.
+
+Matplotlib ``Plotter``
+----------------------
+
+For :class:`~.Plotter`, display the completed figure with Matplotlib's
+``show`` function at the end of the script::
+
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek:
I aligned the renderer guidance more closely with the issue discussion by explicitly naming Spyder as an example IDE where selecting Plotly's browser renderer may be needed. The existing show() and write_html() guidance is unchanged. This push also starts the missing CI run for the PR.

Issue #1050, opened 2024-06-20
Demo zips from Documents are Unusable as is.


GadgetSteve:
The zip files linked in the documents do not include the data needed to be able to run them as follows:
Several of the demos also depend upon `folium` which is not a part of the default installation nor mentioned on the demos page.
1. AIS_Solent_Tracker.py & .ipynb - both require SolentAIS_20160112_130211.csv which is present in the repository but not from the documents demo page.
2. OpenSky_Demo.py & .ipynb - both require OpenSky_Plane_States.csv which is present in this repository but is not downloadable from _nor mentioned in_ the documents. The plot display on the notebook relies upon a set of png files that are only available in the checked out repository '../_static/sphinx_gallery/Plane_Headings/Plane_{angle}.png' -  Both also depend upon `folium` which is not a part of the default installation nor mentioned on the demos page.
3. UAV_Tutorial.py & .ipynb - both require UAV_Rot.csv which is present in the repository but not from the documents demo page.
4. Video_Processing.py & .ipynb - there are a number of additional dependencies that are not mentioned on the demos page but are in the notebook. These both rely on using pytube to download http://www.youtube.com/watch?v=MNn9qKG2UFI and save it to `sample1.mp4` - this is not possible if you are on an isolated or firewalled system _(many of the people who might need to run these demos are)_.

_Several of the demos also depend upon `folium` which is not a part of the default installation nor mentioned on the demos page._

Additional note - the .py versions of these demos do not seem to produce any output nor any indication of what they have done.

GadgetSteve:
Suggestion - since the individual demo pages do include the links to the required downloads and mentions of the additional libraries a note to see those pages on the page with the zips might be a way to resolve this at least partially.

sdhiscocks:
Thanks @GadgetSteve

The ZIP file generation is done by `sphinx-gallery` so may have to see whether there is way to include other files.

As you suggest, a note with ZIPs may be an option to at least highlight that additional files/libraries are needed to run the demos.

We've also designed the demos (and other examples/tutorials) to run as Notebooks, so may be better removing download of the `.py` files.

PR #1328, opened 2026-08-18
Document requirements for downloaded demonstrations

sylvesterkaczmarek:
## Summary

Improves the documentation around downloaded Stone Soup demonstrations, addressing the remaining usability problem described in #1050.

Sphinx-Gallery makes the generated notebook/source downloads convenient, but some demonstrations depend on repository data files, optional packages, generated assets, or external downloads that are not present in every generated archive. This can leave a locally downloaded demo looking self-contained when it is not.

The issue discussion already suggested adding guidance alongside the demonstration downloads so users know to check the full demo page and obtain the additional resources. This change adds that guidance to the demonstrations gallery introduction.

## Change

The demonstrations landing page now:

- explains that generated downloads may require supporting data, assets, packages, or network downloads;
- links directly to the repository data used by `AIS_Solent_Tracker`, `OpenSky_Demo`, and `UAV_tutorial`;
- notes that the AIS and OpenSky demos use `folium`, which is available through the development extra;
- points `Video_Processing` users to the `video` and `ultralytics` extras;
- makes the network requirements of the video download and first-use model-weight download explicit;
- recommends a repository checkout, plus pre-fetching external assets, for restricted or isolated environments.

This complements #1051, which added the OpenSky CSV/icon links, by providing the gallery-level guidance suggested in the discussion on #1050.

## Scope

Documentation only. The Sphinx-Gallery archive generation itself is unchanged.

Addresses #1050

### DIFF
--- a/docs/demos/README.rst
+++ b/docs/demos/README.rst
@@ -3,3 +3,32 @@ Demonstrations
 
 Here are a selection of demonstrations of Stone Soup against real data sets.
 
+Running downloaded demonstrations
+---------------------------------
+
+The Sphinx-Gallery download links are useful for taking a demonstration into a
+local notebook, but some demonstrations also depend on data files, generated
+assets, optional Python packages, or external downloads that are not contained
+in every generated archive. Before running a downloaded demonstration, check
+its documentation page for the complete setup instructions.
+
+The repository data files used by the demonstrations can be downloaded directly:
+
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek:
I tightened this to the approach suggested in the issue discussion: a concise gallery-level warning that downloaded demos may need supporting files, optional packages or network-fetched resources, with users directed back to the full demo page for current setup instructions. I also changed the issue linkage to `Addresses #1050`, since Sphinx-Gallery archive generation and the separate `.py` download question remain outside this documentation-only change.

PR #1329, opened 2026-08-18
Prioritise API objects in documentation search

### LINKED ISSUE #1198: Documentation search results
Using the search in the webpage the results in the latest versions are not in the most intuitive order. 
Like if searching for a datatype or component the first results are other objects or components calling the intended search (like shown below). 
Maybe some parameters in the package dealing with the building the docs can be tuned to improve the results of the search function.  

A simple example can be seen below (comparing 1.4 vs latest) using TrueDetection
1.4 example
Image
Latest (1.8) example 

Image

sylvesterkaczmarek:
## Summary

Improves the ordering of Stone Soup documentation search results so exact API object matches rank ahead of pages that merely mention the same term, addressing #1198.

The reported example is a search for `TrueDetection`, where the most useful result is the API entry for the type itself. Sphinx's default search weights give an exact object-name match a lower score than a title match, which can allow surrounding documentation pages to appear before the requested component.

## Change

- Configure Sphinx's supported `html_search_scorer` hook.
- Add a small custom scorer that raises the weight for exact API object-name matches from the default `11` to `50`.
- Raise partial object-name matches from `6` to `20`, so searches using the final component/type name still favour API entries.
- Preserve Sphinx's default object-priority, title, partial-title, term and partial-term weights.

This is deliberately a ranking-only change. It does not alter the generated search index, stemming, documented objects, or page content.

## Rationale

Stone Soup's documentation is strongly API-oriented. When a query exactly matches a documented class or component name, that object's own API entry is normally the most useful first result. General page-title and text matches remain unchanged and continue to rank normally when there is no strong object match.

The implementation uses Sphinx's documented search-scorer interface rather than patching generated search JavaScript or theme files.

## Scope

Documentation build/search configuration only. Runtime Stone Soup behaviour is unchanged.

Closes #1198

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -260,9 +260,8 @@ html_css_files = ['css/custom.css']
 # Now only 'ja' uses this config value
 # html_search_options = {'type': 'default'}
 
-# The name of a javascript file (relative to the configuration directory) that
-# implements a search results scorer. If empty, the default will be used.
-# html_search_scorer = 'scorer.js'
+# Prefer exact API object matches over pages that only mention the query.
+html_search_scorer = 'search_scorer.js'
 
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'StoneSoupdoc'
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
I traced the docs failure to Sphinx opening `html_search_scorer` relative to the build working directory, which produced `FileNotFoundError: search_scorer.js`. I have updated the config to resolve the scorer path relative to `conf.py`, and pushed the fix so CI can rerun.

PR #1330, opened 2026-08-20
Improve YAMLWriter documentation

### LINKED ISSUE #324: Improve YAMLWriter documentation
`YAMLWriter` documentation is missing, making use unclear without looking at the code.

sylvesterkaczmarek:
## Summary

Expands the `YAMLWriter` documentation so its usage and output format are clear without reading the implementation.

## Change

- explain that at least one data source is required
- document which source drives iteration when several sources are supplied
- note the synchronization expectation for multiple sources
- describe the explicit multi-document YAML output and the keys it can contain
- add a minimal context-manager example so the output file is closed reliably

## Scope

Documentation only. Runtime behaviour is unchanged.

Closes #324

### DIFF
--- a/stonesoup/writer/yaml.py
+++ b/stonesoup/writer/yaml.py
@@ -8,7 +8,29 @@ from .base import Writer
 
 
 class YAMLWriter(Writer):
-    """YAML Writer"""
+    """Write Stone Soup data sources to a YAML document stream.
+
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek:
This remains current and mergeable with no failing checks. Could a maintainer review this when convenient?

PR #1333, opened 2026-08-20
Make seeded sensor measurements deterministic

### LINKED ISSUE #1237: Simulation is not deterministic
The root of all evil is the std-lib "set" type, which is ordered at random (or depending on the memory address of the set items). 

This leads to that a sensor with a RandomState that iterates randomly ordered GroundTruthState objects introducing the same sequence of noise, but not to the same sequence of platforms. 

The bug is treacherous since the set order may not change between multiple program executions. One solution to fix this is to use ordered_set, but be aware that sorting a list created from a set using hash() will yield the same problem.

Simplest way to recreate this is to use one radar sensor on platform 1 that measure platform 2 and platform 3.

sylvesterkaczmarek:
## Summary

Fixes nondeterministic noise-to-target assignment in `SimpleSensor` when ground truths come from an unordered collection.

`SimpleSensor` generates a repeatable sequence of noise samples for a fixed seed, but previously paired those samples with targets in collection iteration order. Set iteration can vary between executions, so the same noise sequence could be attached to different targets.

## Change

- order detectable ground truths deterministically by state vector, timestamp, and stable metadata before assigning generated noise;
- avoid object hashes, memory addresses, and randomly generated path IDs as ordering keys;
- add regression coverage for reversed input order and for co-located truths distinguished by stable metadata.

## Scope

The returned detections remain a set and measurement/noise generation itself is unchanged.

Closes #1237

### DIFF
--- a/stonesoup/sensor/sensor.py
+++ b/stonesoup/sensor/sensor.py
@@ -105,8 +105,10 @@ class SimpleSensor(Sensor, ABC):
 
         measurement_model = self.measurement_model
 
-        detectable_ground_truths = [truth for truth in ground_truths
-                                    if self.is_detectable(truth, measurement_model)]
+        detectable_ground_truths = sorted(
+            (truth for truth in ground_truths
+             if self.is_detectable(truth, measurement_model)),
+            key=lambda truth: tuple(np.asarray(truth.state_vector, dtype=np.float64).ravel()))
 
         if noise is True:
             random_state = random_state if random_state is not None else self.random_state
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Sorting only by `state_vector` still leaves equal-key truths in their incoming order. Because `measure()` accepts a set and Python's sort is stable, two distinct co-located truths with identical vectors can retain nondeterministic set order, so the same seeded noise draws can still be assigned to different truths across runs. Please add a stable identity tiebreaker (or key noise by stable truth identity) and a regression with two distinct truths sharing the same vector.

sylvesterkaczmarek:
Addressed the equal-state tie case raised in the review. The deterministic ordering key now includes timestamp and sorted metadata after the numeric state vector, without relying on hashes, memory addresses, or generated path UUIDs. Added a regression with two co-located truths distinguished by stable target metadata. Focused sensor tests pass: 12 passed.

PR #1334, opened 2026-08-20
Handle Mahalanobis roundoff before square root

### LINKED ISSUE #847: invalid value encountered in sqrt mitigation
Hello, this error is not new, probably due to my data, is there any recommended approaches to reduce or eliminate this?

```
Lib\site-packages\stonesoup\measures.py:259: RuntimeWarning: invalid value encountered in sqrt
  return np.sqrt(super().__call__(state1, state2))
```

sylvesterkaczmarek:
## Summary

Avoids `RuntimeWarning`/`NaN` when a squared Mahalanobis distance is a tiny negative value caused by floating-point roundoff.

## Change

- clamp tiny negative squared distances within numerical tolerance to zero before `sqrt`;
- raise a clear `ValueError` for materially negative squared distances;
- add regression coverage for both the roundoff case and a materially invalid negative result.

This follows the numerical guard already used by the Gaussian Hellinger measure.

Closes #847

### DIFF
--- a/stonesoup/measures/state.py
+++ b/stonesoup/measures/state.py
@@ -284,7 +284,10 @@ class Mahalanobis(SquaredMahalanobis):
             objects
 
         """
-        return np.sqrt(super().__call__(state1, state2))
+        squared_distance = super().__call__(state1, state2)
+        if np.any(squared_distance < -1e-10):
+            raise ValueError("Squared Mahalanobis distance shouldn't be less than 0")
+        return np.sqrt(np.maximum(squared_distance, 0))
 
 
 class SquaredGaussianHellinger(Measure):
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
The Python 3.14 CI failure exposed that the added `-1e-10` cutoff was too strict for normal numerical behaviour: the architecture test can produce small negative squared Mahalanobis values from finite-precision covariance inversion, which caused the new `ValueError` to terminate the tracking thread. I have removed that extra failure mode and now clamp negative squared values to zero immediately before `sqrt`, which is consistent with the squared distance being mathematically non-negative and directly addresses the reported warning/NaN. The focused measure tests pass (45 passed, 1 expected failure), and I repeated the previously failing architecture test 30 times successfully.

PR #1335, opened 2026-08-20
Clarify covariance values in Kalman tutorial

### LINKED ISSUE #347: Tutorial covariances changed to squares
Not an essential change, but it may provide some clarity.
Perhaps when defining covariance matrices / noise in the tutorials, we could use squares (eg. noise_covar = np.diag([2**2]) instead of np.diag([2])) so that it is clear that the user defines variances rather than standard deviations in model instantiations.

sylvesterkaczmarek:
## Summary

Clarifies the first prior covariance definition in the Kalman filter tutorial, following the discussion in #347.

## Change

Adds a note explaining that covariance diagonal entries are variances rather than standard deviations, including the relationship that a standard deviation `σ` corresponds to variance `σ²`.

## Scope

Documentation only. No tutorial calculations or runtime behaviour change.

Closes #347


csherman-dstl (approved):

sylvesterkaczmarek:
The documentation change already has maintainer approval, but this branch had never received any CI statuses. I pushed an empty commit solely to trigger the project checks; the approved documentation content is unchanged.

PR #1336, opened 2026-08-20
Add commit message guidance for contributors

### LINKED ISSUE #508: Encourage good commit message etiquette
While easy enough to understand in context of its branch, a standalone commit message can be difficult to understand the meaning of when merged to the main branch.
This isn't an essential, but it would be nice if we could encourage good Git commit message etiquette (see: https://blog.oddbit.com/post/2019-06-14-git-etiquette-commit-messages/) such that the reviewing of PRs, and tracing of commit history, becomes easier.

(Unless this is already encouraged somewhere, and I've yet to come across it... which is perhaps still reason to raise this issue?)

sylvesterkaczmarek:
## Summary

Adds concise commit-message guidance to the contributor documentation so standalone commits remain understandable after merge.

## Change

- recommend concise, descriptive imperative summaries;
- discourage vague subjects such as `fix`, `updates` and `changes`;
- explain when a commit body is useful for recording intent;
- recommend referencing relevant issues where they provide context.

## Scope

Documentation only.

Closes #508

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -79,6 +79,19 @@ Use standard library and existing well maintained external libraries where
 possible. New external libraries should be licensed permissive (e.g MIT_) or
 weak copyleft (e.g. LGPL_)
 
+Commit Messages
+---------------
+Commit messages become part of the permanent project history and should make
+sense when read outside the context of their development branch.
+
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek:

PR #1337, opened 2026-08-20
Allow singular covariance in Gaussian likelihoods

### LINKED ISSUE #518: Singular Covariance in Motion Model causes issue with Model.pdf()
If the Discrete Time (Bar-Shalom terminology) covariance models are used (as apposed to the Discretized / Continuous Time version) then if the Time increment is set to 1 sec this leads to a singular covariance matrix for the transition model. Calling Model.pdf() then raises a Singular Matrix exception.

Here's an explicit example - for the Constant Velocity model the covariance matrix has the form:

$$ Q = \begin{bmatrix}\frac{1}{4}T^4 & \frac{1}{2}T^3 \\
\frac{1}{2}T^3 & T^2
\end{bmatrix} \sigma^2 $$

Setting $T=1$ gives:

$$
Q = \begin{bmatrix}\frac{1}{4} & \frac{1}{2} \\
\frac{1}{2} & 1
\end{bmatrix} \sigma^2
$$

Since the 2nd row is twice the first row the matrix is singular and thus the exception gets raised. Changing to a different Time Increment alleviates the issue. 

Notes:
- The Discrete Time Constant Acceleration model also has the same issue
- The Discretized Continuous Time models do not have this issue due to their different scaling
- I'm not sure where else this issue may cause problems - Particle Filters, Initiators, Data Associators?

sylvesterkaczmarek:
## Summary

Allows Gaussian model likelihood evaluation for positive-semidefinite singular covariance matrices instead of raising from SciPy.

## Change

- pass `allow_singular=True` to `multivariate_normal.logpdf`;
- preserve existing behavior for ordinary positive-definite covariance matrices;
- add a regression test using a singular covariance matrix and compare against SciPy's singular-Gaussian likelihood.

Closes #518

### DIFF
--- a/stonesoup/models/base.py
+++ b/stonesoup/models/base.py
@@ -337,7 +337,7 @@ class GaussianModel(Model):
         # This is required as log pdf converts arrays to floats
         likelihood = np.atleast_1d(
             multivariate_normal.logpdf((state1.state_vector - self.function(state2, **kwargs)).T,
-                                       cov=covar))
+                                       cov=covar, allow_singular=True))
 
         if len(likelihood) == 1:
             likelihood = likelihood[0]
--- a/stonesoup/models/transition/tests/test_cv.py
+++ b/stonesoup/models/transition/tests/test_cv.py
@@ -5,7 +5,8 @@ import pytest
 import numpy as np
 from scipy.stats import multivariate_normal
 
-from ..linear import ConstantVelocity
+from ..linear import ConstantVelocity, LinearGaussianTimeInvariantTransitionModel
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek:
Fixed the CI regression test failure. `Model.pdf()` returns Stone Soup `Probability`, whose equality operator cannot compare directly with `pytest.approx`; the test now compares `float(probability)` instead. The transition-model test suite passes locally.

sdhiscocks (approved):

sylvesterkaczmarek:
Thanks for reviewing and approving this. Much appreciated.

Issue #316, opened 2020-10-13
Consistency with angular measurements & rotations


DaveKirkland:
PR #310 brought up an inconsistency between the velocity, rotations and the angular measurements. The convention for this should be standardized and documented.

DaveKirkland:
So looking at the rotation in the non-linear measurement and rotation, I see a couple of issues:

1. Rotations are only currently only occurring in the non-linear model. I don't see why they couldn't occur in the linear model as well. If that's the case, then there isn't really anything that ties the bearing/elevation orientation to direction of the rotation. Intuitively it makes sense to have a +ve rotation increase an angle in bearing/elevation models, but that could be very different to the intuitive rotations in linear/Cartesian models.

PhilipOakley:
I'd like to highlight that there is no "right answer" to the choice of reference frames and rotation angles. Rather, there are many different standards for the reference frames. 

The key issues, from a human error viewpoint is to _highlight_ that there are many different reference systems (with examples) and that users will need to be rather careful of the differences between Stone-Soup's options and their own particular system. It's no good documenting [just one choice](https://xkcd.com/927/).

Many major platform systems will have a rather thick tome (e.g. ~80 pages for the early Eurofighter IIRC) detailing all the different reference systems, distortions and offsets (lever-arm) that apply between the different sensors, and that is just for a single platform's attempt at consistency. The rotation discussion also needs to include the [alias and alibi](https://en.wikipedia.org/wiki/Active_and_passive_transformation) distinctions. 

My key point is that the documentation should point at the issues/problems, not just the 'solution'.

PR #1339, opened 2026-08-20
Document sensor reference frame conventions

sylvesterkaczmarek:
## Summary

Adds explicit reference-frame guidance for platforms and mounted sensors, addressing the documentation concerns raised in #316.

## Change

- explain that moving-platform orientation is derived from velocity direction
- distinguish platform-local `mounting_offset` from sensor `rotation_offset`
- document that orientation/rotation vectors use Cartesian x/y/z axes
- warn that external conventions can differ in axis order, handedness, sign, and active/passive interpretation
- link directly to the API properties defining Stone Soup's exact convention

## Scope

Documentation only. This does not change or attempt to standardise model rotation behaviour.

Addresses #316.

### DIFF
--- a/docs/source/stonesoup.sensor.rst
+++ b/docs/source/stonesoup.sensor.rst
@@ -13,6 +13,26 @@ Sensors
 .. automodule:: stonesoup.sensor.base
     :show-inheritance:
 
+Reference frames and mounted sensors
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek:

PR #1341, opened 2026-08-20
Register combined linear Gaussian transition model

### LINKED ISSUE #538: CombinedLinear Gaussian Transition Model
One issue that we noticed is regarding the CombinedLinearGaussianTransitionModel. 

For example looking at the metric example we have a KalmanPredictor. This has the following property:

` transition_model: LinearGaussianTransitionModel = Property(
        doc="The transition model to be used.")`

In the Example the transition model chosen is CombinedLinearGaussianTransitionModel. 
Looking at the code the class is the following:

`class CombinedLinearGaussianTransitionModel(LinearModel, _CombinedGaussianTransitionModel):`

So the type is not LinearGaussianModel and neither are `_CombinedGaussianTransitionModel(TransitionModel, GaussianModel):` and `class LinearModel(Model):`

Because the UInterface uses the parents and children classes to give the opportunity to choose the class, in our case is not possible to select CombinedLinearGaussianTransitionModel as property of the KalmanPredictor. 

Maybe `class CombinedLinearGaussianTransitionModel(LinearModel, _CombinedGaussianTransitionModel)` should have LinearGaussianTransitionModel  as parent class? 

We can't see the link between KalmanPredictor and CombinedLinearGaussianTransitionModel

sylvesterkaczmarek:
## Summary

Registers `CombinedLinearGaussianTransitionModel` as a virtual subclass of `LinearGaussianTransitionModel`, following the maintainer-proposed approach in #538.

## Change

- register the combined linear Gaussian model with `LinearGaussianTransitionModel`
- preserve the existing implementation and inheritance structure
- add regression coverage confirming `issubclass(CombinedLinearGaussianTransitionModel, LinearGaussianTransitionModel)`

This makes the model visible through the expected linear-Gaussian type relationship without inheriting additional implementation or properties.

Closes #538

### DIFF
--- a/stonesoup/models/transition/__init__.py
+++ b/stonesoup/models/transition/__init__.py
@@ -1,3 +1,7 @@
 from .base import TransitionModel
+from .linear import LinearGaussianTransitionModel, CombinedLinearGaussianTransitionModel
+
+
+LinearGaussianTransitionModel.register(CombinedLinearGaussianTransitionModel)
 
 __all__ = ['TransitionModel']
--- a/stonesoup/models/transition/tests/test_registration.py
+++ b/stonesoup/models/transition/tests/test_registration.py
@@ -1,0 +1,5 @@
+from ..linear import CombinedLinearGaussianTransitionModel, LinearGaussianTransitionModel
+
+
+def test_combined_linear_gaussian_registered_as_linear_gaussian():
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek:

sdhiscocks (approved):

sylvesterkaczmarek:
Thanks for reviewing and approving this. Much appreciated.

Issue #942, opened 2024-01-30
`GaussianStateUpdate` is not an isinstance of `StateUpdate`


gawebb-dstl:
```
>>> isinstance(GaussianStateUpdate([1],[[1]], None), StateUpdate)
False
```
This isn't intuitive (to me at least). In the above code `StateUpdate` can quite easily be replaced by `Update`.

This isn't a big issue but I wonder if `StateUpdate` is required. It's only usage within the ‘main’ branch is for tests. However it may be used in other branches or repositories.

PR #1342, opened 2026-08-20
Clarify update type hierarchy

sylvesterkaczmarek:
## Summary

Clarifies the update-type hierarchy that prompted #942 without changing the public inheritance structure.

## Change

- explain that `Update` is the common base class for update types
- document that each specialised update combines `Update` with its corresponding state type
- clarify that `GaussianStateUpdate` and other specialised updates are siblings of `StateUpdate`, rather than subclasses of it
- describe `StateUpdate` as the update counterpart to the plain `State` type

## Scope

Documentation only. Runtime behaviour and class inheritance are unchanged.

Addresses #942

### DIFF
--- a/stonesoup/types/update.py
+++ b/stonesoup/types/update.py
@@ -13,20 +13,30 @@ from ..base import Property
 
 
 class Update(Type, CreatableFromState):
-    """ Update type
+    """Update type.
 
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek:

Issue #848, opened 2023-08-29
Make more measurements in gap between observed points?


ColeBromfield01:
[kalman_3d.txt](https://github.com/dstl/Stone-Soup/files/12467021/kalman_3d.txt)
[points2.csv](https://github.com/dstl/Stone-Soup/files/12467023/points2.csv)
(Code and points, updated)

Hi again!

Thanks for the help in solving my previous issue--now trying to figure out something more minor.  My latest recorded points have an (intentional) gap in measurements.
![track3](https://github.com/dstl/Stone-Soup/assets/143104236/b05bd2b4-253a-4e30-a09b-20f4e4a8c513)

As you can see, only a couple of Kalman points are predicted in this gap.  Is there a way to code it so that more predictions are made in that gap?

PR #1343, opened 2026-08-20
Document intermediate prediction timing

sylvesterkaczmarek:
## Summary

Clarifies how to obtain denser predictions across measurement gaps, addressing the question in #848.

## Change

- explain that a predictor can be called for any requested timestamp
- document that tracker prediction times are normally driven by detector/input timestamps
- make explicit that trackers do not automatically insert intermediate prediction states
- describe two ways to obtain denser predictions:
  - call the predictor directly at intermediate timestamps
  - yield intermediate tracker timestamps with empty detection sets so existing tracks advance using predictions

## Scope

Documentation only. Tracker and predictor behaviour is unchanged.

Addresses #848

### DIFF
--- a/docs/source/stonesoup.predictor.rst
+++ b/docs/source/stonesoup.predictor.rst
@@ -4,6 +4,19 @@ Predictors
 .. automodule:: stonesoup.predictor
     :no-members:
 
+Prediction timing in trackers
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek:

Issue #1072, opened 2024-08-05
PyTorch Support for Detectors


karacanil:
Hey,

I was wondering if there will be a PyTorch support under the **stonesoup.detector** module in the future?
What is the recommended way of integrating a PyTorch detector into our tracker pipeline.
For example, this pipeline here: [pipline](https://stonesoup.readthedocs.io/en/v1.3/auto_demos/Video_Processing.html#sphx-glr-auto-demos-video-processing-py)

Thanks.

sdhiscocks:
@karacanil I think that would be a good addition. Not something I'm aware anyone else has done yet, but would be welcome contribution.

Best bet would be to look at the tensor flow detector (see below for link), but replace the tensor flow elements with your PyTorch model, or generic component like the tensor flow one where you provide your own model (or file path to model).

The key thing is the `_get_detections_from_frame(self, frame)` method which has a frame as an input with pixels array, which you can then pass to your model and generate one or more `Detection` objects from each frame.

https://github.com/dstl/Stone-Soup/blob/4b6bd37813d4c96e3efa63847ab3f1ea7153a508/stonesoup/detector/tensorflow.py#L20-L112

PR #1344, opened 2026-08-20
Document PyTorch detector integration pattern

sylvesterkaczmarek:
## Summary

Documents the integration pattern for PyTorch and other external object-detection frameworks, following the maintainer guidance in #1072.

## Change

- point users to `TensorFlowBoxObjectDetector` as the existing reference implementation
- explain the frame-to-`Detection` contract used by video object detectors
- identify `_get_detections_from_frame(frame)` as the framework-specific inference/conversion step
- describe replacing TensorFlow inference with PyTorch model calls while retaining Stone Soup `Detection` outputs
- recommend the public `Detector` interface for reusable external integrations, noting that the video helper is private

## Scope

Documentation only. No PyTorch dependency is added and no detector runtime behaviour changes.

Addresses #1072

### DIFF
--- a/docs/source/stonesoup.detector.rst
+++ b/docs/source/stonesoup.detector.rst
@@ -7,6 +7,25 @@ Detectors
 .. automodule:: stonesoup.detector.base
     :show-inheritance:
 
+Integrating other object-detection frameworks
+---------------------------------------------
+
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek:

PR #1345, opened 2026-08-25
Fix CircleSamplePositionActionGenerator membership checks

### LINKED ISSUE #1179: Contains method issue on
The inherited `__contains__` method on `CircleSamplePositionActionGenerator` isn't suitable, as the action choices aren't discrete, and also values are generated randomly.

https://github.com/dstl/Stone-Soup/blob/458045c91819930bee2e8afe5b92e1cc34267a80/stonesoup/movable/action/move_position_action.py#L155-L156

`__contains__` method should probably just check that the move is within the max travel radius.

sylvesterkaczmarek:
Fixes #1179.

`CircleSamplePositionActionGenerator` inherited a membership check that iterated the generator. Because the generator samples random positions on each iteration, checking whether an existing action was contained in it was effectively a new random draw rather than a validation of the action.

This change gives the circle sampler a deterministic `__contains__` implementation based on `maximum_travel`, matching the constraint used when actions are generated. It accepts both `MovePositionAction` and `StateVector` inputs and uses the same small floating-point tolerance as `MaxSpeedPositionActionGenerator`.

A focused regression test covers an in-range state vector, an in-range action, and an out-of-range state vector.

### DIFF
--- a/stonesoup/movable/action/move_position_action.py
+++ b/stonesoup/movable/action/move_position_action.py
@@ -173,6 +173,8 @@ class CircleSamplePositionActionGenerator(SamplePositionActionGenerator):
         "sampling area."
     )
 
+    epsilon = 1e-6
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
`__contains__` checks the circle radius but ignores `action_space`. A target can return `True` even though `__iter__` would never yield it because it is outside the configured bounds. Apply the same action-space check here and add a bounded-space regression.

sylvesterkaczmarek:

sdhiscocks:
Appears to be copied from #1251

PR #1346, opened 2026-08-25
Fix OpenSky ground truths without ICAO24

sylvesterkaczmarek:
## Summary

Fix `OpenSkyNetworkGroundTruthReader` when OpenSky returns a state without an ICAO24 identifier.

For missing `icao24`, the reader correctly creates a standalone `GroundTruthPath`, but after appending the state it tries to add `groundtruth_dict[path_id]` to the current results. With `path_id is None`, no such dictionary entry exists and the reader raises `KeyError`.

## Change

- add the local `path` object to `updated_paths`
- preserve the existing dictionary-backed path behaviour for states that do have ICAO24 identifiers
- add a regression test for an unidentified OpenSky state

This only affects the existing missing-ICAO24 fallback path.

### DIFF
--- a/stonesoup/reader/opensky.py
+++ b/stonesoup/reader/opensky.py
@@ -167,6 +167,6 @@ class OpenSkyNetworkGroundTruthReader(_OpenSkyNetworkReader, GroundTruthReader):
                     path = groundtruth_dict[path_id]
 
                 path.append(GroundTruthState(state.state_vector, state.timestamp, metadata))
-                updated_paths.add(groundtruth_dict[path_id])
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
The only failing CI job was Python 3.11, and its failure was in `stonesoup/tests/test_stitcher.py::test_correct_no_stitched_tracks1`, which is unrelated to this OpenSky reader change. All other Python versions, flake8, and docs passed. I also re-ran the new missing-ICAO24 regression locally and it passes. I have triggered a fresh CI run without changing the substantive patch.

PR #1347, opened 2026-08-25
Fix zero-valued OpenSky positions

sylvesterkaczmarek:
## Summary

Preserve valid OpenSky states when longitude, latitude or geometric altitude is `0.0`.

The reader currently validates position data with a truthiness check. That treats valid numeric zero values as missing, so aircraft can be dropped when they are on the prime meridian, equator, or at zero geometric altitude.

## Change

- reject only `None` position values rather than all falsy values
- add a regression test covering zero longitude, latitude and altitude
- keep genuinely missing position values filtered out

No API shape changes are introduced.

### DIFF
--- a/stonesoup/reader/opensky.py
+++ b/stonesoup/reader/opensky.py
@@ -80,7 +80,7 @@ class _OpenSkyNetworkReader(Reader):
                     if state[8]:  # On ground
                         continue
                     # Must have position (lon, lat, geo-alt)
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
Closing this in favour of #1321, which contains the same OpenSky zero-valued position fix and regression coverage. Keeping the earlier PR avoids duplicate review effort.

Issue #1314, opened 2026-08-14
`is_clutter_detectable` method works differently to `is_detectable` method


csherman-dstl:
When a sensor makes a measurement of a true state, the `is_detectable` method is used to check if the Cartesian location of the target is visible to the sensor. 

When using a clutter model to produce clutter detections however, a number of Cartesian points are generated which are immediately converted into detections and then only the detection space is used to determine whether the clutter was detectable.

1. This means that a bearing only sensor with max range 10 m would produce a clutter detection on clutter which is 100 m away as there is no range value to compare.
2. Similarly, a clutter detection produced from a point behind an object would not be ruled out.
3. This means that the `is_clutter_detectable` method is unique to each measurement model implementation.

It is reasonable to consider the clutter detections as being related to the generated Cartesian position or is this method just a proxy to produce random locations in measurement space?
If the clutter is related to the Cartesian position then should the clutter models instead produce a set of States which can be measured to make clutter detections with?

PR #1348, opened 2026-08-26
Modify `ClutterModel` Class to return clutter in state space

csherman-dstl:
As mentioned in #1314 clutter models currently produce Clutter objects in the measurement space.
This PR changes this functionality so that the clutter is produced in the state space.
The sensor's `measure` function now converts the clutter from state space to measurement space just as it does for the ground truths.
Also it seems weird to me that the clutter model uses a `measurement_model` property which is given to it in the sensor model but the clutter model doesn't actually have a measurement_model attribute.

Potential breaking changes:
 * To remove the `measurement_model` which is attached to the clutter model, the clutter model now has `ndim_state` and `mapping` properties in line with MeasurementModel and TransitionModel classes
 * I now can't see a reason for the `is_clutter_detectable` method to exist and so have removed it but am happy to re-add just pointing to `is_detectable` it for backwards compatibility.


sylvesterkaczmarek:
I think this changes an existing multi-sensor behaviour in a way that is easy to miss.

Previously `Sensor.measure()` assigned its current `measurement_model` to the shared `ClutterModel`, and `ClutterModel.function()` used that model's `mapping` when placing the generated Cartesian values into the state. This meant one clutter model could be shared by sensors with different measurement mappings, provided they operated over the same state space.

With this change, `mapping` is stored permanently on `ClutterModel` and `function()` uses `self.mapping`, while the measurement model assigned by the sensor is no longer involved in constructing the clutter state. So if two sensors share a clutter model but map different state dimensions, the second sensor will generate clutter in the first model's configured dimensions before applying its own measurement model.

That also seems to conflict with the class documentation, which still says the same `ClutterModel` can be used with multiple measurement models as long as they operate in the same state space.

Could `function()` take the active mapping/measurement model from the calling sensor instead, or should sharing across different mappings now be explicitly unsupported? A regression with two sensors sharing one clutter model but using different mappings would make the intended contract clear.

sylvesterkaczmarek (commented):
The move to generating clutter in state space is cleaner, but I still think the sharing contract is inconsistent here.

ClutterModel now owns a fixed mapping, while its class documentation still says the same clutter model can be shared by multiple measurement models as long as they operate in the same state space. Two sensors sharing one clutter model but using different mappings would therefore generate clutter into whichever mapping is configured on the clutter model, rather than the mapping of the active sensor.

There is also a leftover self.clutter_model.measurement_model = measurement_model assignment in SimpleSensor.measure(), even though ClutterModel.function() no longer reads that attribute.

Could the active mapping be supplied by the sensor when clutter is generated, or should sharing across different mappings now be explicitly unsupported? A regression with two sensors sharing one clutter model but using different mappings would pin the intended behaviour.

PR #1349, opened 2026-08-31
Add optional data storage to dictionary readers

### LINKED ISSUE #470: Have data readers store data
Some data readers build a dictionary when being iterated through. For example, the `CSVGroundTruthReader` builds a `groundtruth_dict` so that it can keep track of path ids that it has come across, and avoid yielding them in separate `GroundTruthPath`.

In some situations, it might be useful to keep hold of these dictionaries (eg. for visualisation, evaluation of the entire dataset), perhaps just stored as an attribute of the reader.
For large datasets this probably wouldn't be good, but could be an option that the user can (de-)select?

sylvesterkaczmarek:
## Summary

Adds opt-in retention of the internal path and track dictionaries built by dictionary-backed readers, addressing #470.

## Change

- Add `store_data=False` to ground-truth and track dictionary readers.
- Expose retained data through `groundtruth_dict` and `track_dict` when enabled.
- Keep the default behaviour unchanged so readers do not retain these dictionaries after iteration unless requested.
- Add focused tests for both enabled and default-disabled behaviour.

## Scope

No change to yielded reader data or existing defaults. The additional memory cost only applies when `store_data=True`.

Closes #470

### DIFF
--- a/stonesoup/reader/generic.py
+++ b/stonesoup/reader/generic.py
@@ -109,6 +109,14 @@ class _DictGroundTruthReader(GroundTruthReader, _DictReader):
     """An abstract reader for dictionaries containing truth data."""
 
     path_id_field: str = Property(doc='Name of column to be used as path ID')
+    store_data: bool = Property(
+        default=False,
+        doc='Keep the dictionary of ground truth paths after iteration')
+
+    @property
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:

Issue #729, opened 2022-09-27
Extend 2d density plotter to support more state types


sdhiscocks:
Reference #458, it would be good for the density plotter to also support plotting Gaussian mixtures (like used in PHD filter) and other types of state.

PR #1350, opened 2026-08-31
Support Gaussian mixtures in 2D density plots

sylvesterkaczmarek:
## Summary

Extends `Plotter.plot_density()` to support `GaussianMixture`, addressing #729.

## Change

- Detect `GaussianMixture` inputs in the existing 2D density plotter.
- Evaluate the weighted Gaussian component densities directly using each component's mapped mean, covariance, and weight.
- Build the density grid from the component extents using three standard deviations.
- Preserve the existing KDE behaviour for collections of state sequences.
- Continue forwarding plotting keyword arguments to `pcolormesh`.

## Regression coverage

Adds a focused test with two weighted Gaussian components, including correlated mapped covariance, and verifies the plotted density against the direct weighted multivariate-normal calculation.

Addresses #729

### DIFF
--- a/stonesoup/plotter.py
+++ b/stonesoup/plotter.py
@@ -13,7 +13,7 @@ from matplotlib.legend_handler import HandlerPatch
 from matplotlib.lines import Line2D
 from matplotlib.patches import Ellipse, Patch, Polygon
 from scipy.special import ellipeinc
-from scipy.stats import kde
+from scipy.stats import kde, multivariate_normal
 try:
     from plotly import colors
 except ImportError:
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The mixture branch uses raw component weights, but `GaussianMixture` does not require them to sum to one and its own `mean` normalises them. Equivalent mixtures with weights `[0.4, 0.6]` and `[4, 6]` therefore plot densities 10× apart. Normalise the weights before summing and add an unnormalised-weight test.

sylvesterkaczmarek:
I checked the failed docs job. The failure was outside this change, in `trackfusion/track_fusion_example.py`, where SciPy rejected a non-positive-definite covariance during a gallery run. The Python test matrix for this PR was otherwise green. I have also removed unrelated wording/error-message edits from `plotter.py` so the PR stays focused on Gaussian-mixture density support; the push triggers a fresh docs run.

PR #1351, opened 2026-08-31
Add optional Sobol sampling for circle actions

### LINKED ISSUE #1180: Use (optionally?) quasi random number generator for actions with CircleSamplePositionActionGenerator
Using quasi random number generator, like Sobol, would result is better distribution of samples.

https://docs.scipy.org/doc/scipy/reference/stats.qmc.html

sylvesterkaczmarek:
## Summary

Adds opt-in Sobol quasi-random sampling to `CircleSamplePositionActionGenerator`, addressing #1180.

## Change

- Add `use_qmc=False` to circle position sampling and expose it through `CircleSampleActionableMovable`.
- Use a deterministic 2D Sobol sequence when enabled while preserving the existing pseudorandom default.
- Skip the Sobol origin because the generator already yields a stay-put default action.
- Continue the Sobol sequence when samples outside `action_space` need replacement.
- Add regression tests for exact Sobol samples, bounded action spaces, and default behaviour.

## Scope

Existing behaviour is unchanged unless `use_qmc=True`.

Closes #1180

### DIFF
--- a/stonesoup/movable/action/move_position_action.py
+++ b/stonesoup/movable/action/move_position_action.py
@@ -3,6 +3,7 @@ from collections.abc import Iterator, Sequence
 from itertools import product
 
 import numpy as np
+from scipy.stats import qmc
 
 from ...base import Property
 from ...sensormanager.action import ActionGenerator, Action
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek:

sdhiscocks (approved):

sylvesterkaczmarek:
Thanks for the review and approval. Much appreciated.

